PR #64, opened 2012-08-19
Use a hash table for anchors

peterkelly:
This patch greatly improves performance on documents containing a large number of id/name attributes by using a hash table instead of a linked list while checking for duplicate ids.


balthisar:
@peterkelly, I will close this PR, as it's rolled into the develop-500 branch currently.

Unfortunately due to the complexity of all of the neglected PR's over the years, the PR history is a bit fractured. This means that your name won't be shown in the commit history, but the links for the commits will still link back to this PR so that you will still be acknowledged for your contributions.

PR #76, opened 2012-12-13
Update src/tags.c to allow meta tags to be put between span tags.

hdrlab:
Patch to allow meta tags to be put between span tags. Tidy was terminating the span section at the meta tag. The suggested change corrects this.

AFAIK, putting meta tags inside a span block is valid HTML5 (e.g., for embedding microdata).


balthisar:
@hdrlab, I'm going to close this because of the behavior described in the previous comment. Please feel free to resubmit it if you can prevent the missing line break issue. Your pr _will not be ignored_ for 2.5 years this time.

PR #84, opened 2013-03-22
Added support for building with CMake

thekvs:


balthisar:
@thekvs, I will close this PR. As you may or may not know all current development is now in the develop-500 branch, and you'll be happy to know that it's completely cmake based. We hope to bring the development branch into master very soon.

PR #109, opened 2013-12-03
Improve backwards compatibility

ermshiperete:


balthisar:
@ermshiperete, I'm going to close this PR. At least one of these PR's is pulled into the current development branch develop-500. Because of accumulation of so many unfulfilled PR's, can I ask that you evaluate these against the develop-500 branch?

As you can see there is a _lot_ of activity going on right now, and I can promise that if you submit a PR to the develop-500 branch, you're not going to be ignored for a year again.

Issue #111, opened 2014-01-09
Tidy omits closing tag for self closing div when doing HTML5 output


ermshiperete:
```
cat test.html
<html><body><div a='b'/></body></html>
```

Running `tidy test.html` produces the following incorrect HTML5 code where the closing div tag is missing:

```
<!DOCTYPE html>
<html>
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 (experimental) for Linux https://github.com/w3c/tidy-html5/tree/c63cc39">
<title></title>
</head>
<body>
<div a='b'>
</body>
</html>
```

geoffmcl:
The latest development branch, develop-500 is now on the repo...
$ git checkout develop-500
Does this branch correctly handle this closing tag?
If not could you provide a simple sample, if different to the above, together with the configuration used...
If yes, can we close this issue?

balthisar:
It works, and no response from @ermshiperete. I will close this, but @ermshiperete please feel free to open it again if the problem isn't solved.

PR #113, opened 2014-01-09
Print end tag for self-closing tag

ermshiperete:
When we output HTML5 and we encounter a self-closing tag we have to
output an end tag to be standard conform.

This fixes issue #111.


balthisar:
@ermshiperete, I will close this PR, as it's rolled into the develop-500 branch currently.

Unfortunately due to the complexity of all of the neglected PR's over the years, the PR history is a bit fractured. This means that your name won't be shown in the commit history, but the links for the commits will still link back to this PR so that you will still be acknowledged for your contributions.

PR #115, opened 2014-02-27
Add support for 'role' attribute.

saadatm:
This commit adds support for 'role' attribute for all HTML elements except:
 - those elements that are obsolete in HTML5. (See a list of all HTML5 elements here: http://dev.w3.org/html5/markup/elements.html)
 - those elements that can have "no role" AND have strong native semantics. See: http://www.w3.org/html/wg/drafts/html/master/dom.html#sec-strong-native-semantics (The only exception to this exception is the "input" element.)

The commit does not check for valid 'role' values.


balthisar:
@saadatm, this has been pulled into develop-500 branch, and so I will close this PR.

Unfortunately due to the complexity of all of the neglected PR's over the years, the PR history is a bit fractured. This means that your name won't be shown in the commit history, but the links for the commits will still link back to this PR so that you will still have acknowledgements for your contributions.

PR #118, opened 2014-04-27
Make messages passed into the callback filter localizable 

balthisar:
Implements a TidyReportFilter2 that passes all of the separate strings to the filter callback function instead of an assembled string, making it possible to localize each of the individual strings in the application layer that is using TidyLib.

This avoids having to localize `localize.c` and makes it possible to use the host OS' built-in localization methods in the host OS environment.


balthisar:
Looks like this was pulled into develop-500, so I will close this.

PR #121, opened 2014-07-31
Change Unicode to UTF-16

sebgod:
Unicode is already supported via UTF-8,
what is supported with this flag is the transfer format UTF-16,
which commonly misnamed Unicode, e.g. in the Microsoft Developer Documentation.


balthisar:
@sebgod, I feel bad asking you to close another PR. As Tidy has moved to `cmake` exclusively in the develop-500 branch this PR will no longer be relevant. Would you mind closing the PR if that explanation is satisfactory?

sebgod:
Ok! Good to reduce the number of build systems.

PR #135, opened 2014-12-07
Add support for emscripten as a compilation target

lovasoa:
This allows tidy to be compiled as a .js file, and run in a browser.

Demonstration here: http://lovasoa.github.io/tidy-html5/demo.html

I'm not used to C projects infrastructures, so sorry if I didn't organize things the right way.


balthisar:
@lovasoa, this is an excellent piece of work. As you may or may not know, we've abandoned all of the old build systems in favor of `cmake` and all of the current work is taking place in the develop-500 branch.

Could I ask you to close this PR with no further action? If you can work this into `cmake` and post another PR, you won't be ignored for so long again (as they say, we're under new management).

geoffmcl:
@lovasoa, as @balthisar points out this is quite interesting. I recently ran across a simple example that uses cmake as the build mechanism - http://forums.4fips.com/viewtopic.php?f=3&t=1201 

There he uses a python script to run cmake with a number of `special` parameters to use the emscripten SDK, but this could be manual, or done in any scripting... The important point is that cmake CMakeLists.txt is used as the `generator`...

I was able to successfully build the js for that sample, using cmake to generate a Makefile, and mingw-make to do the compile... and it ran fine... I was in Windows 7.

Now that we are using a cmake system you need to rebase your clone to bring it up to date, like -

 1. git remote add upstream git@github.com:htacg/tidy-html5.git
 2. git fetch upstream
 3. git rebase upstream/master

Or just delete it, and start again. A PR must be against the current **master** htacg branch. Note no longer using the develop-500 branch.

It will probably be just the addition of a script in build/cmake (sh/bat/py) to pass extra parameters to cmake and run the appropriate compiler... Mainly setting `-DCMAKE_TOOLCHAIN_FILE=$EMSCRIPTEN/cmake/Modules/Platform/Emscripten.cmake`...

And maybe an `option` in CMakeLists.txt, like BUILD_USING_EMCC, or something, to aid in the detection of the emscripten SDK, but it seems the above sample did not need this...

Hope you get a chance to revisit this... I think there would be quite some interest in a javascript version of tidy... making it easy to put up a full latest 'tidy5' web service...

geoffmcl:
Since no further comments in months, am closing...

In any case the fork needs to rebase to master, since the merge has conflicts...

PR #140, opened 2015-01-16
README.md: Code highlighting now balanced

sebgod:


balthisar:
I think we might not start accepting pull requests willy-nilly without some internal alignment, but in the spirit of the “new” Tidy (and given the harmlessness of this PR), let it be known that the first accepted PR under the new regime is the insertion of backticks into the documentation. This ushers in a new era for Tidy.

PR #141, opened 2015-01-16
Makefile: fix spelling of .PHONEY -> .PHONY

sebgod:
IMHO the "PHONY" option is written .PHONY


balthisar:
Given that we've moved to `cmake` can you confirm that we can close this PR with no further action?

PR #165, opened 2015-02-21
Use HTML5 mapping for entities &lang; and &rang;

fred-wang:


geoffmcl:
@fred-wang have merged this change. Thanks...

PR #170, opened 2015-03-05
Docs Tweaks

pedromorgan:
Added some links and items to menu bar, dynamic version, and some basic pages..

this is the same caper as I use in other doxygen projects..
eg
http://api-docs.freeflightsim.org/
http://fgx.freeflightsim.org/

So can develop this more.. preferably in markdown.


balthisar:
I'm considering accepting this, as it offers a lot of good, extra features. I have a couple of concerns, though.

- The tabs' text are unreadable on my 5K display with the color changes that you made.
- Are all of the documentation styles necessary? We still have to address coding style, and that should include documentation style for the source code. For example the auto-briefs.
- Example.1.c is showing up as a source file.
- I think the headers are the only thing that should be documented. If there's something important to the API that's buried in source code, then let us know so that we can move it into a header. We can leave this as is for now, but non-header files are meant to be private.
- The quickref link in the tabs section is broken.

Don't let my nit-picking get you down. This is a _definite_ improvement. I'll leave the PR open for the next 24 or 25 hours in case you want to push any more tweaks, and then I'll accept it unless anyone voices any objections.

pedromorgan:
thanks for the comments and nit picking is good ;-)

tabs - I have no comment on this, I just made the colors greyscale so its neutral. Maybe it need more staturation/gamma whatever its called..

doc styles - I've switched more stuff on to see what "comments" are there and work, and still trying to figure out whether its a doxygen setting or a comment style problem. But indeed they an be silenced down to the "public api"

example = an oops, need to move examples to a new directory as `build/documentation/` is added as INPUT, will fix this tonite by creating `build/documentation/examples/`. But is there anything in testing we can include as examples ?

quickref = Probably `xsltproc` is not installed.. ? the script has been modified so that qucikref is generated, and then "copied" added to the site as a static document..
https://github.com/pedromorgan/tidy-html5/blob/pedro/build/documentation/build_docs.sh#L117

pedromorgan:
Got nothing much more to add to this atmo..

Unless.... 
- we move some of this stuff  - http://www.html-tidy.org/documentation/
  into doxygen also ??

Also maybe we can build a zip file of the completed temp/html adn a "downlaod the manual" ?

pedromorgan:
OK... Tweaked it until all issues I had are done...

Please review !!!

balthisar:
@pedromorgan,

Well, obviously my merge strategy wasn't perfect -- I thought it would show up here, but it's merged into Master now.

pedromorgan:
Err.. U havent updated your local branch to match pedro..

Your missing a few of the final commits...

balthisar:
The first thing I did was `git pull git://github.com/pedromorgan/tidy-html5.git`

What's missing? Looking at the diffs seems to capture all of your changes as of about an hour ago or so.

pedromorgan:
I think your at "merge with head" in the bottom block above"..1d56d54

So the pages are split, and examples/ subdir etc

PR #173, opened 2015-03-05
Travis setup

pedromorgan:
This is the travis setup. and preview is at
- https://travis-ci.org/pedromorgan/tidy-html5/builds

Also changed the README to emphasise what it does..

Enable
=========
Too enable, one of the admin needs to gointo.. 
https://github.com/htacg/tidy-html5/settings/hooks

and add service "travis"..

it will then appear here..
- https://travis-ci.org/htacg/tidy-html5


balthisar:
@pedromorgan, can you open issue for this? I'm not certain what this site does, and their own description is pretty threadbare. Wikipedia says they do online continuous integration. What problem does this solve for Tidy? Don't answer here, let's take it to issues, but those are the types of questions some of us will ask. It looks harmless enough until you say enable hooks (which is probably harmless).

If this affects everyone's workflow in some way, it's especially important to understand what you're asking us here.

Thanks!

pedromorgan:
Maybe this answers your question.
http://stackoverflow.com/questions/22587148/trying-to-understand-what-travis-ci-does-and-when-it-should-be-used

Re "hooks"... don't panic, its just a switch that makes github trigger a build on travis 

Travis should not mess things up as is only appears in the "repos" as
- a .travis.yml = travis config file
- a link/image in readme to the "build banner"
eg
README = https://github.com/fgx/fgx/tree/next 
doxygen = http://fgx.freeflightsim.org/developer.html

pedromorgan:
Moot,,, nobody want it !

vielmetti:
Can we reopen this? Now that we have a release, it would be useful to make sure that there's some way to ensure that subsequent changes don't cause regressions.

vielmetti:
Thanks to @pedromorgan 's excellent start, I now have something running using Travis in my development version. You can see that file at https://github.com/vielmetti/tidy-html5/blob/master/.travis.yml

I converted the installation so that it runs in one of Travis's containers and thus the whole thing builds in about 90 seconds.

Of note for the next person to come around to this is that the version of `cmake` on Travis is too old to build `tidy-html5`, and thus this dependency is brought in via the `kalakris-cmake` apt source.

pedromorgan:
@vielmetti Halleluiah..

Some olde style devs around here are nervous about config of travis + build etc..

There was a Jenkins instance also to build a project.. but no fluff there..

We need to automate this process though, so that any changes to master build the docs,.. whoever is doing it in future moving forward.. automajically...

I got the stuff here.. am familiar with this.. only problem is upstream devs need to either embrace, or delegate..

vielmetti:
@pedromorgan thanks. I don't fully understand Travis yet & would be happy to work with someone familiar with the system.

Yes, building the docs automatically so that they track changes is important. There's a page full of doxygen errors and warnings at https://travis-ci.org/vielmetti/tidy-html5/jobs/80146632 that are all probably easy to fix. 

The other first next thing I want to try to tackle is to update the test infrastructure, so that I can catch regressions and (especially) differences in the build between x86 and ARM architectures. My try on the Raspberry Pi 2 came up with about 100 lines of `diff` output (cf https://gist.github.com/vielmetti/2f50364c103919f59ae9) . Refactoring that so that tests either pass or fail and don't all have to be looked at by eyeballs every time. The testing README at https://github.com/htacg/tidy-html5/blob/master/test/README.md scares me just a little bit.

vielmetti:
@pedromorgan embrace, delegate, or do the work in a fork; the main thing I want to work on is not the code itself (which seems in good hands) but the build/test/deploy infrastructure so that eventually it's dirt-easy to make `apt-get install tidy` or its equivalent do exactly the right thing for all of the systems I have to deal with and not have to fiddle with things by hand and not have to worry that the next release has broken something.

pedromorgan:
Ok @vielmetti were in..
I got a few RPI here,, and we make a build farm for fun..

What travis does, I suppose is compile and make sure it compiles..

Thats why I had a jenkins instance.. So i could create my slaves..
32bit+ububtu = local
64bit+debian = remote
rpi-arm6 = a job..

So this is all in the front end.. and I am ready.. and waiting..

Snag is that upstream.. them dev's are harcore c coders and dont understand et all..
So I am definately not wasting my time again.. for it to be kicked out..

You can't teach an olde horse to suck gravel

I have a clean vision on how it goes, but upstream/masters have different debates and plans more concered witht the binary.. the docs is a second thought..

So thats why I almost give up.. pain in the ass init..
Kinda unserstand why Dianna Ross left Motown.. 

I give them the patches, setup a website, and later its kinda..
how do I do this ??
So am not wasting ma enengy anymore..
Will submit patch,, Jenkins it tehre and all setup..

What more can I do.. ?????????????????????




pissed of pedro who's submitted all thsi stuff already.. !!!

Now go figure it out yourselfves.. I cant be bothered anymoere.. to turn an olde goat around.... quite frankly.....!"!!

vielmetti:
@pedromorgan - regarding Jenkins -

I have a Pi that I can throw into the Jenkins mix as a remote, I think, but the last time I looked at setting it up for myself the complexity was too enormous. (Or more to the point, I wasn't working with someone who knew it cold.)

My last go at Travis for building docs using a modern-ish version of doxygen looks like it also did work, at least for the trivial case of not generating errors and only a few warnings. I picked up on what Travis was all about through the `Node-RED` project, those IBM guys know their build tools cold and it's a pleasure to work with them.

vielmetti:
More notes on travis + doxygen for what it's worth: https://github.com/travis-ci/apt-source-whitelist/issues/40 we'd like a newish doxygen made available inside a supported Travis container so that the build is as efficient as possible.

vielmetti:
I now have a Travis config at https://github.com/vielmetti/tidy-html5/blob/master/.travis.yml which successfully builds `tidy` and its documentation on both Mac and Linux, with both clang and gcc, with dependencies installed on both systems.

It doesn't do any `make test`, but that's a next step. I don't yet know the right test harness, but that would be a good separate issue.

geoffmcl:
@vielmetti, @pedromorgan, zoot, while pondering whether this needed re-opening, there was a great flurry of comments ;=)) good to see...

Did not **like** some, nor agree with all the comments given, but to touch a few...

It was here that I saw some yaml that did `cp tidy tidy5`, before calling documentation\build-doc.sh... can not find this `build-doc.sh` script to fix it... but please do not perpetuate the `5`...

> It doesn't do any `make test`

That's because there are no real `unit` tests that can be done, or at least identified... 

Do you propose adding some, maybe using some selected files from `test/input`? 

Look forward to that...

Issue #177, opened 2015-03-06
no command line manual page


anEffingChamp:
At least for my sake `tidy` could benefit from a manual page, such than `man tidy` would render some thing. It could even just start as the same output as `tidy -?`, but it is another common path to exploring command line options.

geoffmcl:
@anEffingChamp as mentioned in #178 we are certainly working towards bringing this back... the previous man tidy pages were generated using xsltproc which is not a very crossplatform way... any help appreciated...

pedromorgan:
An observation re packaging

This is what cpack currently includes in the deb, missing is the man file ```tidy.1```
```
./usr/bin/tidy5
./usr/include/buffio.h
./usr/include/platform.h
./usr/include/tidy.h
./usr/include/tidyenum.h
./usr/lib/libtidy5.a
```

So the snag is that ```tidy.1``` needs to be included 

`tidy.1` is created with the `build_docs.sh` and requires  xsltproc
```
$TIDY_PATH -xml-help > "tidy-help.xml"
xsltproc "tidy1.xsl" "$tidy-help.xml" > "$OUTP_DIR/tidy.1"
```

So this is a bit chicken and egg in that..
1. tidy needs to be compiled, to output the `tidy-help.xml`
2. xml is then processed into man page

So the sequence needs to be..
1. build tidy
2. generate man page
3. run cpack and include man page for install

pedromorgan:
http://stackoverflow.com/questions/3375341/man-page-generation-packaging-installation-with-cmake

geoffmcl:
@pedromorgan Thanks Pete. As the link you gave points out this can **ALL** be done in CMakeLists.txt, which is where is SHOULD be done, NOT as a separate process, including -

  1. finding if xsltproc is installed
  2. running the just built tidy executable to generate the tidy-help.xml and tidy-config.xml
  3. running xsltproc to generate the tidy.1 and quickref.html
  4. installing tidy.1 and quickref.html appropriately along with the rest
  5. then cpack would pick them up and include them in the deb and rpm

As a quick guess that would be something like, added after the exe builds, before the cpack stuff, but this is obviously untested -

````
if (UNIX)
    find_program( XSLTPROC_FOUND NAME xsltproc )
    if (XSLTPROC_FOUND)
        set(TIDY1XSL /path/to/tidy1.xsl)
        set(TIDYHELP ${CMAKE_BINARY_DIR}/tidy-help.xml)
        add_custom_target(man ALL)
[...]

PR #181, opened 2015-03-07
Create man page as part of cmake build

pedromorgan:
See issue
https://github.com/htacg/tidy-html5/issues/177

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,9 +175,10 @@ endif ()
 if (UNIX)
     find_program( XSLTPROC_FOUND xsltproc )
     if (XSLTPROC_FOUND)
-        ## NOTE: man name must match exe ie currently `tidy5.`` not `tidy.1`
+        ## NOTE: man name must match exe ie currently `tidy5.1` not `tidy.1`
         ## also could use `manpath` command output to determine target install path
         message("*** Generating man ***")
+        set(TIDY1_MANFILE tidy5.1)
         set(TIDY1XSL ../documentation/tidy1.xsl)
         set(TIDYHELP ${CMAKE_BINARY_DIR}/tidy-help.xml)
         add_custom_target(man ALL)
@@ -194,17 +195,17 @@ if (UNIX)
         add_custom_command(
             TARGET man
             SOURCE ${TIDYHELP}
-            COMMAND xsltproc ARGS ${TIDY1XSL} ${TIDYHELP} > ${CMAKE_BINARY_DIR}/tidy5.1
-            OUTPUTS ${CMAKE_BINARY_DIR}/tidy5.1
+            COMMAND xsltproc ARGS ${TIDY1XSL} ${TIDYHELP} > ${CMAKE_BINARY_DIR}/${TIDY1_MANFILE}
+            OUTPUTS ${CMAKE_BINARY_DIR}/${TIDY1_MANFILE}
             VERBATIM
             )
         add_custom_command(
             TARGET man
             SOURCE man
-            DEPENDS ${CMAKE_BINARY_DIR}/tidy5.1
+            DEPENDS ${CMAKE_BINARY_DIR}/${TIDY1_MANFILE}
         )
         # repeated for quickref.html - may need different targets...
-        install(FILES ${CMAKE_BINARY_DIR}/tidy5.1 DESTINATION local/man/man1)
+        install(FILES ${CMAKE_BINARY_DIR}/${TIDY1_MANFILE} DESTINATION local/man/man1)
         # add the quickref.html install
     endif ()
 endif ()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,45 @@ if (BUILD_SAMPLE_CODE)
     # no INSTALL of this 'local' sample
 endif ()
 
+#==========================================================
+# Create man pages
+#==========================================================
+if (UNIX)
+    find_program( XSLTPROC_FOUND xsltproc )
+    if (XSLTPROC_FOUND)
+        ## NOTE: man name must match exe ie currently `tidy5.`` not `tidy.1`
+        message("*** Generating man ***")
+        set(TIDY1XSL ../documentation/tidy1.xsl)
+        set(TIDYHELP ${CMAKE_BINARY_DIR}/tidy-help.xml)
+        add_custom_target(man ALL)
+        
+        # run built EXE to generate xml output 
+        add_custom_command(
+            OUTPUT ${TIDYHELP}
+            COMMAND ${CMAKE_BINARY_DIR}/tidy5 -xml-help > ${TIDYHELP}
+            COMMENT "Generate ${TIDYHELP}"
+            VERBATIM
+            )
+        # repeated for tidy-config.xml
+        # run xsltproc to generate the install files..
+        add_custom_command(
+            TARGET man
+            SOURCE ${TIDYHELP}
+            COMMAND xsltproc ARGS ${TIDY1XSL} ${TIDYHELP} > ${CMAKE_BINARY_DIR}/tidy5.1
+            OUTPUTS ${CMAKE_BINARY_DIR}/tidy5.1
+            VERBATIM
+            )
+        add_custom_command(
+            TARGET man
+            SOURCE man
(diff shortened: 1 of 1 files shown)


balthisar:
I'm trying to troubleshoot this PR... CMAkesLists.txt specifies cmake_minimum_required (VERSION 2.8), but it breaks with 3.1.1 because of the deprecated use of SOURCE in the add_custom_command.

Since you know a lot more about Cmake than I do, can you suggest a workaround?

I'm hacking at it now, but I might take all day...

pedromorgan:
>> Since you know a lot more about Cmake than I do, can you suggest a workaround?

err no, am a complete muppet and copied the guru's (geoff) snippets in #177 and modified...

Geoff wrote the code essentially, I only tweaks it.. ;-)

balthisar:
That's okay, I keep learning a lot about cmake little by little. I'll merge this once I can get it working on 3.1.1. I see the other PR for dox, and I'll get to it in a bit, too.

geoffmcl:
@pedromorgan, @balthisar, thanks for adding this... 

I did a test in my Ubuntu 14.04, and it worked fine... and tidy5.1 got added to the deb...

Added a small tweak... all output should be kept to the 'build' folder, the so called CMAKE_BINARY_DIR, so it can be .gitignore'd, and not polute the source tree.

And that would be true for any added cmake docs generation, so it also does NOT become part of the source. Preferrably into a sub-directory, like say `${CMAKE_BINARY_DIR}/docs` so it can all be ignored by one entry to .gitignore.

Of course the needed components, like the tidy1.xsl, .cfg, etc have to be part of the source, BUT not any generated items...

And as previously expressed hope these can soon be moved to says a 'docs' directory, or somewhere, but out of the build directory...

Still to think on a window equivalent... any ideas appreciated...

balthisar:
@geoffmcl said:

> And that would be true for any added cmake docs generation, so it also does NOT become part of the source. Preferrably into a sub-directory, like say ${CMAKE_BINARY_DIR}/docs so it can all be ignored by one entry to .gitignore.

What a timely simultaneous posting. I requested a PR on the cmake_documentation branch for your code review. It moves documentation out of build, and only builds into a temp directory, and cleans up its temporary files.

If you pull it, I can clean it up tomorrow and redirect the build to build/temp in instead of documentation/temp, as you're right: that makes a lot more sense.

Also I wasn't able to test on Windows... I have a feeling it might break because some DOS commands aren't the same as Unix.

geoffmcl:
@pedromorgan, @balthisar, just did an install test, and now think the line 
`install(FILES ${CMAKE_BINARY_DIR}/${TIDY_MANFILE} DESTINATION local/man/man1)`
should be just `... DESTINATION man/man1)`

Seems cmake already adds the prefix `/usr/local` so the man page ended up in `/usr/local/local/man/man1/tidy5.1`, but I am certainly NOT an expert on where things should go in unix ;=(( Just tried it, did install and `man tidy5` brings up a 4.9.20 page ;=))

As for windows there is no 'man' but as part of the SDK or MSVC there is a help file compiler, that if I remember correctly begins as a .rtf then uses something like a hcrtf.exe to convert to a winhelp .hlp file, but so long since I did this, need to re-discover the process...

And will pull the branch for further testing soonest... but not sure why there is a PR and a branch... but will try to work it out ;=))

pedromorgan:
Re the destination
https://github.com/htacg/tidy-html5/blob/master/CMakeLists.txt#L179

Manpath on my machine = ubuntu
```bash
$  manpath
/usr/local/man:/usr/local/share/man:/usr/share/man
```

So `/usr/local/man` is there the cutom compiled stuff seems to go

`/usr/share/man` seems to be kinda "different" as its full of directories, eg en, es, ch, and man1, man2, and tar.gz files.. containing the man file itself..

When I installed the deb here, the destination `local/man/man1` worked, `main/man1` didnt and could figure out where it ended up

Maybe we an use manpath output and split it, same style as the version.txt

pedromorgan:
Maybe be should be doing this in develop branch, so as not to break for users,, until we got it figured out

geoffmcl:
@pedromorgan, that is strange that two linux cases do different install things!

First, I did not use the deb to install it. I just used -  
$ sudo make install

And for some idiot reason can not get the deb generated at the moment... something not found and can't write some file... all too weird... it worked yesterday using  
$ make package, or  
$ cpack

But if you run -  
$ dpkg -c tidy5-4.9.20-64bit.deb  
where does it show it will install the tidy5.1 - the full path is shown...

As stated with `local/man/man1` in CMakeLists.txt, I found it in `/usr/local/local/man/man1`...

If you want to try `manpath` and split the output that is also fine. If cmake finds a FULL path, like `/usr/local/man/man1` it knows not to add an install prefix, which is `/usr/local` by default...

And right at this moment to me master IS the development branch, as it usually should be...

When we get to a 5.0.0 release, then it is my idea to create say a rel-5.0.0 branch, which people checkout if they want the last 'stable'. master is always the leading edge of develpment and should not garanteee anything... but we are not there yet...

geoffmcl:
@pedromorgan, ok, got the deb to build... using cpack... just closed and re-opened a terminal???

This is using a modified CMakeLists.txt, with DESTINATION man/man1

$ dpkg -c tidy5....deb

shows it will install in `/usr/man/man1/tidy5.1`???? WHAAAA! What's going on here!?!?!? 

The cmake_install.cmake CLEARLY shows -

````
IF(NOT DEFINED CMAKE_INSTALL_PREFIX)
  SET(CMAKE_INSTALL_PREFIX "/usr/local")
````

Then a little later -

````
 FILE(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/man/man1" ...
````

So that means `$ sudo make install` will go to the right place, while using the DEB will send it to some other BAD place??????????

Now, on the other hand, if I alter the CMakeLists.txt to the absolute path `/usr/local/man/man1' then as I thought, cmake_install.cmake AND the DEB agree ;=)) as they should...

But the cmake file adds a curious WARNING or ERROR if the user defines CMAKE_[WARN|ERROR]_ON_ABSOLUTE_INSTALL_DESTINATION? But I suppose we can ignore this...

So maybe using `manpath` to set this absolute path is the way to go...

Also note reading around that appearently Fedora uses `${CMAKE_INSTALL_PREFIX}/share/man/man1`... which is also output as one of the paths from 'manpath'... so maybe this is the best choice?

And another titbit read during my searching and that is what package should be generated for what system...

It said add `include(SystemSpecificInformations)` to your CMakeLists.txt, then SPECIFIC_SYSTEM_PREFERED_CPACK_GENERATOR will contain the preferred generator for the system. And suggested the list would be - 

 1. Windows - NSIS
 2. Debian - DEB
 3. Ubuntu - DEB
 4. Fedora - RPM
 5. SuSE - RPM
 6. other linux - TGZ

I will experiment in my cpack-test repo, which I suppose is where true 'experimentation' should take place...

You can learn more every day ;=))

geoffmcl:
@pedromorgan, UGH, experimenting in cpack-test and reading some more seems you need to more or less write your own SystemSpecificInformations.cmake as it seems not part of the standard cmake distribution, so forget that last stuff!

Let's just get the man page setup correctly...

balthisar:
@geoffmcl, I'm fine with all of the suggestions. It was mostly an effort to get the documentation out of build/, and the learn a lot more about CMake. It quite literally took my entire day to make those changes. Overall it was worth it as it will never take that long again.

We can avoid all of the xsltproc dependencies by reverting some of the changes that @pedromorgan introduced. There's no absolute need to include quickref.html in developer documentation, for example.

For Unixes xsltproc is standard, or at least very easily installable. The typical user _expects_ a man page, whereas on Windows a typical user has no such expectation for command line applications. I don't see a need to generate a Windows help file, either, because those are most often associated with GUI applications.

From the strategy standpoint, is there any value in installing the quickref.html document at all? `tidy -help` Works on all platform, Unices will have man pages, We can still make sure it's on the website. I don't see that it has value as part of the distribution and it frees us from xsltproc dependency on non-Unix platforms.

pedromorgan:
xsltproc is required to create the man page, 
and the quick ref is very useful, so I don't want that to go if possible

so no xsltproc == no man or quickfref

PR #184, opened 2015-03-08
Cmake documentation

balthisar:
I'm issue this as a PR instead merging in directly.

@geoffmcl, as our CMake guru, would you take a look at the changes for generating the documentation? It gets the job done, but I'm as yet unfamiliar with the "CMake way" of doing things.


geoffmcl:
@balthisar have had a look at the cmake_documentation branch. But not the PR? The CMakeLists.txt looks fine but some questions...

Why do you separate building libtidy/tidy and the documentation? That is if I add -DBUILD_DOCUMENTATION=ON you switch off building tidy, BUT the tidy exe is needed for part of it...

To me building tidy should ALWAYS be ON. That is what the project is about... and should NOT need to be switched off just because something else is also built...

Now building the man page tidy5.1 is ONLY for unix, NEVER windows... as explained, windows uses an entirely different 'help' system... different tools, different source, different output...

And if cmake is always building tidy exe than it is a 'target' so cmake 'knows' where it put a target output... we do not need to find it where ever it was output...

Just for information in windows the output is in `build/cmake/<configuration>/tidy5[d].exe`, where the configuration can be Release, Debug, plus some others... and not in the root build/cmake as it does in linux... and note the 'd' added for the Debug version...

Building the doxygen generation can also be done in window, if you have doxygen installed, which I do, but the tidy binary would not be found as you wrote it, but as suggested above we should not need to 'find' tidy, if it is already a 'target'. 

So trying to separate the things...

  1. Let's always build tidy, no matter what else is being built.
  2. Let's always use that built tidy to output to say `${CMAKE_BINARY_DIR}` tidy-help.xml, tidy-config.xml, tidy.help.txt, tidy.config.txt.
  3. Have already added the first 2 to .gitignore. Just need to add the other 2, and no need to clean up.

This would be done in ALL platforms, since it only requires the built product, tidy5 binary, where ever it is put...

Now if (UNIX), then we can check for xsltproc, and if found generate the man page to be part of the install... and deb,... with the two files already available...

And finally, if (BUILD_DOCUMENTATION) then we already have some of the files needed. Only need to check if doxygen exists... and the 4 files are available, BUT in windows, remember NO xsltproc, so may need some switches for that...

Using a 'temp' directory is fine, but again why not in `${CMAKE_BINARY_DIR}/temp/` then you do NOT need any clean up... but can if you want... You will also note the root .gitignore already tells git to ignore everything temp*

So to experiment in windows I just added -

````
set(TIDYHELP ${CMAKE_BINARY_DIR}/tidy-help.xml)
set(TIDYCONFIG ${CMAKE_BINARY_DIR}/tidy-config.xml)
set(TIDYHELP2 ${CMAKE_BINARY_DIR}/tidy.help.txt)
set(TIDYCONFIG2 ${CMAKE_BINARY_DIR}/tidy.config.txt)
add_custom_target(docs ALL DEPENDS tidy5)
add_custom_command(
    TARGET docs
    COMMAND tidy5 -xml-help > ${TIDYHELP}
    COMMAND tidy5 -help-config > ${TIDYCONFIG}
    COMMAND tidy5 -show-config > ${TIDYCONFIG2}
    COMMAND tidy5 -h > ${TIDYHELP2}
    COMMENT "Generate ${TIDYHELP} ${TIDYCONFIG} ${TIDYHELP2} ${TIDYCONFIG2}"
    VERBATIM
    )
````

Inserted it just before the 'Create man pages' block, and it worked in windows... I got the 4 files created... and just tried the same in linux, and it worked there too... and of course in both cases I had to make sure there was no other tidy5 in the path... that is tidy5 -h yielded an error...

Hope this helps... 

And BTW I am not actually a cmake Guru ;=(( Just a very good cmake documentation searcher using google/yahoo/etc... and always find someone else has answered my question when, sometimes by chance, I phrase it correctly... sometimes need a few tries at this...

geoffmcl:
@balthisar hi, long time no hear ;=((

Now that documentation building is now in the root CMakeLists.txt, and it seems to sort of work - some inputs, outputs could be arranged better - but is this PR needed now?

Likewise for the cmake_documentation branch...

geoffmcl:
As the documentation generation is now merged into the `master` branch, am closing this for now...

The generation needs some more testing, but seems to mostly work...

geoffmcl:
And as github offers to delete this `cmake_documentation` branch will try that..

Issue #196, opened 2015-04-09
Tidy does not allow block elements in <caption>


willydee:
According to http://dev.w3.org/html5/spec-author-view/the-caption-element.html#the-caption-element arbitrary flow content is allowed within `CAPTION` elements, and the following code passes the validation without warnings. 
```html
<!DOCTYPE html>
<html>
<head>
  <title></title>
</head>
<body>
<table>
  <caption>
    <nav><button>Test</button></nav>
  </caption>
  <tbody>
    <tr><td>Lorem ipsum</td></tr>
  </tbody>
</table>
</body>
</html>
```
Tidy output:
```
line 8 column 3 - Warning: missing </caption> before <nav>
line 9 column 5 - Warning: <nav> isn't allowed in <table> elements
line 7 column 1 - Info: <table> previously mentioned
line 10 column 3 - Warning: discarding unexpected </caption>
line 8 column 3 - Warning: trimming empty <caption>
Info: Document content looks like HTML5
4 warnings, 0 errors were found!
```

willydee:
Probably related to https://github.com/htacg/tidy-html5/issues/169

geoffmcl:
@willydee, thanks for the pull request #197. Your patch certainly does the job ;=))

The only thing I noted is maybe we should 'expand' the comment in `lexer.c(2750):` to include this issue when calling `AdjustTags`... but quite minor...

And yes, we can expect more dynamic table adjustments as we move forward supporting html5++ while still maintaining full html4-- support.

Out of time to finalise and test a little more today, but will get to this soonest... Thanks again...

geoffmcl:
@willydee, Thanks for this PR... merged and closed, bumped version, and added comments in lexer.c

Now tidy5 passes this, like the validator, without problems...

And constructed a backwards compatible test -
````
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
"http://www.w3.org/TR/html4/loose.dtd">
<html>
<head>
  <meta http-equiv=Content-Type content="text/html; charset=utf-8">
  <title>Tidy does not allow block elements in caption #196</title>
</head>
<body>
<table>
  <caption>
    <nav><button>Test</button></nav>
  </caption>
  <tbody>
    <tr><td>Lorem ipsum</td></tr>
  </tbody>
</table>
</body>
</html>
````

and as expected the 'warnings' return, just like in the validator -
````
[...]

Issue #169, opened 2015-03-04
<p> moved out of <a>


ralfjunker:
Revision: 2e383c6
Date: 28.02.2015 20:32:38

Tidy moves a &lt;p&gt; tag out of an &lt;a&gt; tag.

Example:
```html
<a href="l1"><p>one</p></a>
```
Tidy processes this to:
```html
<!DOCTYPE html>
<html>
<head>
<title></title>
</head>
<body>
<a href="l1"></a>
<p>one</p>
</body>
</html>
```
Testing with FireFox 36, the &lt;p&gt; tag stays within the &lt;a&gt; tag.

Should not Tidy behave the same?

geoffmcl:
@ralfjunker, yes this is a real **problem!** It stems from trying to keep HTML4-- parsing while trying to also correctly parse HTML5++

Others have tried to 'fix' this by redefining the TidyTag_A, bringing it into line with HTML5++

`{ TidyTag_A, "a", ... (CM_INLINE|CM_BLOCK|CM_MIXED), TY_(ParseBlock), ...`

But this led to other problems when parsing HTML4--, so I have reverted it to -

`{ TidyTag_A, "a", ... (CM_INLINE), TY_(ParseInline), ...`

For sure the following HTML5 sample, passes the W3C validator with no problem -

````
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<title>p moved out of a #169</title>
</head>
<body>
<a href="l1"><p>one</p></a>
</body>
</html>
````

But make a few changes, bringing it back to HTML4, and the validator will complain strongly -

````
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" 
"http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8">
<title>p moved out of a #169</title>
</head>
<body>
<a href="l1"><p>one</p></a>
</body>
</html>
````

And of course most browsers will display them the same, with the 'one' as a link... they have no role or interest in validating or fixing the document...

One **answer** to all this would be to have **two** separate tidy libraries - html4 and html5 - but I, and maybe others, hope it does not come to that...

For such a dual mode tidy the important question is at what point does tidy choose it is in html5 mode. There are some obvious indicators, like if the document has an earlier doctype, or the user has chosen a --doctype of not html5, then it should warn, and try to fix to suit. And if in html5 mode, should leave this alone...

[...]

PR #197, opened 2015-04-09
Fix for #196: HTML5 allows block elements in <CAPTION>

willydee:
@geoffmcl Here is a patch for the issue reported earlier. Solved it the same way you did in #169.
Best regards, willydee


geoffmcl:
@willydee, thanks for the pull request #197. Your patch certainly does the job ;=))

The only thing I noted is maybe we should 'expand' the comment in `lexer.c(2750):` to include this issue when calling `AdjustTags`... but quite minor...

And yes, we can expect more dynamic table adjustments as we move forward supporting html5++ while still maintaining full html4-- support.

Out of time to finalise and test a little more today, but will get to this soonest... Thanks again...

geoffmcl:
@willydee thanks for this... now merged...

PR #206, opened 2015-04-26
Support build with MinGW on Windows

d1vanov:
Hello, 

Thank you very much for tidy-html5! I've recently found that current version from master requires only two small tweaks to make it perfectly buildable with MinGW on Windows, here they are.

### DIFF
--- a/src/mappedio.c
+++ b/src/mappedio.c
@@ -102,7 +102,7 @@ void TY_(freeFileSource)( TidyInputSource* inp, Bool closeIt )
 #include "message.h"
 
 #include <errno.h>
-#if _MSC_VER < 1300  /* less than msvc++ 7.0 */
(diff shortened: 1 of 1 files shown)


geoffmcl:
@d1vanov looks perfectly reasonable... now merged and thanks...

geoffmcl:
@d1vanov I thought I still had **MSVC6** still installed somewhere to check if Tidy still compiles with that **OLD** version, which is the only reason for that macro in the first place, but it seems **not**.

But for now will leave it there, just in case. Does no harm...

Good to hear Tidy compiles ok with MinGW...

d1vanov:
@geoffmcl yeah, this version of MSVC is indeed **OLD** :) I just didn't want to come with breaking changes so I decided to make no harm in the pull request. Perhaps these macros can be removed right before releasing 5.0 or when you see fit.

PR #236, opened 2015-08-09
Add PHP support section

skynet:


geoffmcl:
@skynet I have no problem with this, so why not merge it?

In general additional information to help users with integrating `libtidy` with other specific packages is very welcome...

If the number of different packages grows large we may want to consider adding say a specific package readme, like `README.<package>.md` for each...

But this is an **important** header file name change, and certainly `deserves` to be noted in the main general readme...

PR #242, opened 2015-08-25
Drop unnecessary semicolons from includes in example.

b4hand:
These semicolons are almost certainly harmless, but they are not idiomatic in C.

None of the real code appears to follow this convention of having semicolons after each `#include`, so it seemed strange for them to be included in the example code.

### DIFF
--- a/build/documentation/examples/example.1.c
+++ b/build/documentation/examples/example.1.c
@@ -1,9 +1,9 @@
 
 
-#include <tidy.h>;
-#include <buffio.h>;
-#include <stdio.h>;
-#include <errno.h>;
+#include <tidy.h>
+#include <buffio.h>
+#include <stdio.h>
+#include <errno.h>
 
 int main(int argc, char **argv )
 {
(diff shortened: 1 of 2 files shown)


geoffmcl:
@b4hand agree... good catch, and thanks...

Could I ask you do one more thing before I merge this PR, and that is to change the header name from `<buffio.h>` to the new correct name `<tidybuffio.h>` in the first example as it has already been done in the 2nd...

Of course there is the question why we have this example file twice, and several other files... I think we intended to drop the `build/documentation` folder in favor of the `documentation` folder, or something... 

But I note that the `build/documentation/README.md` has lots more information than its counterpart, so some information should be moved before such a deletion... 

And the CMakeLists.txt still refers to `build/documentation/tidy1.xsl` which is slightly different to its counterpart... will check these later...

If you could just add that header name change will merge the results... thanks...

b4hand:
As requested, it now says `#include <tidybuffio.h>`.

geoffmcl:
@b4hand now merged into master, and thanks...

PR #243, opened 2015-08-27
Bug fixes and stability improvements

dmit25:
Got some nasty errors while parsing incorrect html pages.
Found workarounds to prevent tidy from failing on this pages.

Specific incorrect html support added (prevents failures)
+ some new tags added (frequently used non-standard html tags)

I'd thank u if you could support INDEX, NOINDEX and AUTHOR tags
I know that these tags are non-standard but they are frequently used by some hand-broken programmers and I would like to use some meta information about them (especially content-mode).

I am not familiar with project's test system and build tidy using MSVC 14 through VCXPROJ (not c-make) and use NUnit in my project that invokes tidy through some wrapper, that is why no new tests are added, but ofc if it's necessary I can adapt my tests (most likely it will be just copy paste and rename)

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -396,6 +396,8 @@ typedef enum
   TidyTag_NOLAYER,  /**< NOLAYER */
   TidyTag_NOSAVE,   /**< NOSAVE */
   TidyTag_NOSCRIPT, /**< NOSCRIPT */
+  TidyTag_NOINDEX,  /**< NOINDEX */
+  TidyTag_INDEX,    /**< INDEX */
   TidyTag_OBJECT,   /**< OBJECT */
   TidyTag_OL,       /**< OL */
   TidyTag_OPTGROUP, /**< OPTGROUP */
@@ -445,6 +447,7 @@ typedef enum
   TidyTag_NEXTID,   /**< NEXTID */
 
   TidyTag_ARTICLE,
+  TidyTag_AUTHOR,
   TidyTag_ASIDE,
   TidyTag_AUDIO,
   TidyTag_BDI,
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -2008,9 +2008,13 @@ static Node *GetCDATA( TidyDocImpl* doc, Node *container )
             else if (c == '/')
             {
                 TY_(AddCharToLexer)(lexer, c);
-
-                c = TY_(ReadChar)(doc->docIn);
-                
+                /*incorrect tag end <script> </ script> */
+                /*incorrect tag end </![cdata */
+                while ((c = TY_(ReadChar)(doc->docIn)) &&
+                    (c != EndOfStream) &&
+                    (TY_(IsWhite)(c) || c == '!')){
+                    lexer->lexsize -= 2;
+                }
                 if (!TY_(IsLetter)(c))
                 {
                     TY_(UngetChar)(c, doc->docIn);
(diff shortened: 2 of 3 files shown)


geoffmcl:
@dmit25 thanks for your interest in tidy and a suggested extension and 'fixes'...

It seems there are several different parts of your patch... trying to divide it up...

 1. addition of three elements, index, noindex, author
 2. add CM_BLOCK to table of the iframe element
 3. deal with `</ script>` and `</![cdata...`
 4. deal with a incorrect `<?>` tag
 
#### 1. Adding new elements

This question is addressed by the relatively new add custom tags options...

```
new-blocklevel-tags         Tag names                                         
new-empty-tags              Tag names                                         
new-inline-tags             Tag names                                         
new-pre-tags                Tag names                                         
```

See http://www.htacg.org/tidy-html5/quickref.html#new-blocklevel-tags

While this will not remove the `warning` it does remove the `error`, thus the file gets written without using `--force-output yes`...

And such new tag lists can be added to your common tidy config file to make them semi-permanent each and any time you run tidy...

Now whether we **should** support certain tags just because `they are frequently used by some hand-broken programmers` is a separate question. 

On these three, I do not really know their use frequency, but I think I would vote `no` to including them in tidy's internal table, which, in effect, should conform to the W3C tags lists... 

But am open to ideas, and comments of others... 

#### 2. Adding `block` to 'iframe'

This may be the case, but I would like to see some samples of what this change actually fixes.

If there is a **bug** here then this should be separated to its own issue, with the sample code... and not bundled in here...

#### 3. Deal with misformed end tags

Hmmm, just reading the code you changed a single next byte read to a `while()`, and if it matches your criteria - is white or '!' - you decrease the lexer size by `-= 2;`, each time, which may be correct...

Again I need to construct some sample code with the shown faults, and test it further...

Maybe you could add any samples you have... please just the essence to show the problems rather than a cut-and-paste from a document...

#### 4. Deal with incorrect tag `<?>`

Hmmm, it is always difficult just `reading` code, but it seems you are reading another character from the stream after the `<?`, and if it is a `>`, decrease lexer size by 4?. 4! Maybe that is correct, but again would need to setup a sample, and test...

But if it is not a `>` you do not seem to unget this character, so it would be lost... Again would need to test with some samples to determine exactly what will happen...

### This Pull Request

Well, does the ability to add `new` tags through a config option solve the first problem? Can these addtitional tags be removed from the PR?

And you can see while the others may be legitimate issues, they should be separated into individual issues and dealt with separately... separate testing... discussion...

This keeps the git history separate making it possible, for example, to revert just one at any time, rather than a whole block...

So as stated, thanks for your work on Tidy!, that is always good, but at this stage can only reject this as a single PR...

Concerning a `project test system` there is something in the `test` directory, but most of this was accumulated long ago... it still needs some work to try to make it some sort of `unit test`... a WIP...

And good to hear you have no problem using MSVC 14 to compile Tidy! The README.md does mention that is is also possible to use the MSVC IDE load of the solution file for building, while only showing the command line use...

geoffmcl:
@dmit25 just a rough sample input using the 3 tags... but not sure how to eactly use them since they are `undocumented` elements...

```
F:\Projects\tidy-html5\test>type input5\in_243-1.html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<title>Test 1 - Issue #243</title>
</head>
<body>
<author>Smith</author>
<index>index</index>
<noindex>noindex</noindex>
</body>
</html>
```

Output:

```
F:\Projects\tidy-html5\test>C:\MDOS\tidy-issue-228.exe --tidy-mark no --show-info no --new-blocklevel-tags "author,index,noindex" input5\in_243-1.html
line 8 column 1 - Warning: <author> is not approved by W3C
line 9 column 1 - Warning: <index> is not approved by W3C
line 10 column 1 - Warning: <noindex> is not approved by W3C
3 warnings, 0 errors were found!

Info: Document content looks like HTML5
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<title>Test 1 - Issue #243</title>
</head>
<body>
<author>Smith</author>
<index>index</index>
<noindex>noindex</noindex>
</body>
</html>
```

The `errors` are reduced to an `appropriate` set of warnings...

And using the `--show-warnings no` option, this can be reduced to 1 line -
```
3 warnings, 0 errors were found! Not all warnings/errors were shown.
```

This of course is using the very latest in the `issue-228` branch, namely `HTML Tidy for Windows version 5.1.7.D231`... but this new tags support is unchanged for quite some time now, well before 5.0.0.RC1...

Issue #59, opened 2012-08-07
Code cleanup


petdance:
Since we're starting a new project, let's take the time up front to clean up some of the artifacts from the old tidy:

* Standardize tabs/spaces and indenting, including adding vim mode lines (and emacs if anyone wants them)
* Clean all spurious whitespace so that we don't have false diffs down the road.
* Remove all CVS cruft.  These have no value to us:

```
   CVS Info:
     $Author: arnaud02 $
     $Date: 2007/02/11 09:45:08 $
     $Revision: 1.9 $
```

balthisar:
I'm in favor of biting the bullet and fixing this as our last operation before merging into Master. Future PRs will be affected, so a code freeze is a pretty going point to do this, too.

geoffmcl:
@petdance, @balthisar, did a quick analysis of the 52 source files, 50,352 lines, approx. 2.7MB... trying to determine what, if anything, needs 'fixing' in this regard.

Some 1.4% of the lines, 722 lines, in 34 files, have 1 or more trailing spaces. Does not seem like a giant problem ;=))

Regarding tabs, only 4 files contain tabs  

 1. tidylib.c - 2144 lines, 4 w/tab, 4 tabs
 2. sprtf.c - 440 lines, 45 w/tab, 142 tabs
 3. parser.c - 4898 lines, 124 w/tab, 341 tabs
 4. tidy-int.h - 155 lines, 6 w/tab, 19 tabs

The sprtf.c is my windows ONLY debug addition, so this should not be a problem. This file is not included/compiled in the release build, and is only ever 'viewed' in my MSVC IDE which is set to display tabs as 4 spaces, so it still all lines up neatly...

parser.c is because I (or someone?) imported some code from another repo that used tabs. I did remove some, but a few remain... At 150K bytes it is the 2nd largest source file in tidy.

The largest is attrdict.c, 13,827 lines, 1.5MB, with no tabs, and only 131 lines with 1 or more trailing spaces.

But with a grand total of just 506 tabs, 0.01% of the total bytes, again it does not seem a problem.

As for relining, and indenting, having been through and through most of the source over the last few months, I have NOT seen anything noticeably bad...

In fact I **LOVE** that doing a big diff between cvs/sf tidy and current tidy5, there are still thousands of lines that are totally **UNCHANGED**. This is particular good for doing code reviews, comparison, etc...

So in summary, maybe I missed something, but I would not recommend this be done at this time.

[...]

PR #244, opened 2015-08-27
Whitespace cleanup

b4hand:
You can see this is only whitespace changes here:

https://github.com/htacg/tidy-html5/pull/244/files?w=1

### DIFF
--- a/build/documentation/examples/example.1.c
+++ b/build/documentation/examples/example.1.c
@@ -44,4 +44,3 @@ int main(int argc, char **argv )
     tidyRelease( tdoc );
     return rc;
 }
-
--- a/build/documentation/examples/example.1.c
+++ b/build/documentation/examples/example.1.c
@@ -18,7 +18,7 @@ int main(int argc, char **argv )
     printf( "Tidying:\t%s\n", input );
 
     // Convert to XHTML
-    ok = tidyOptSetBool( tdoc, TidyXhtmlOut, yes );  
+    ok = tidyOptSetBool( tdoc, TidyXhtmlOut, yes );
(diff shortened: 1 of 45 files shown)


geoffmcl:
@b4hand thanks for this PR, but we have discussed this before, I think in #59, and maybe others, but it was eventually **not** done!!!

As mentioned back then, from my perspective this change is not good. It is still possible to do a big diff between earlier versions of tidy, and the current, and get a meaningful diff showing the important additions, subtraction without the massive clutter of just pure space changes...

Yes, I know diff has a `-w` switch to suppress some space different but then you also lose intentional space changes; Unfortunately `$ git diff` does not seem to support this option; Then there is a great WinMerge GUI, which I often use to compare files, and my own file comapare [FC4](https://github.com/geoffmcl/FA4) and directory compare (DC4W) utilities... each dealing with pure space change in different ways...

So I certainly prefer we do **not** do such a clean up!!!

And I just do not see much gain... I quickly re-ran my tab and trailing space [chktabs.pl](https://github.com/geoffmcl/scripts) script and it seems little has changed since the #59 summary... Processed 58 files, 58473 lines, appx 2998185 bytes...

I added a check for leading and trailing newlines in the source files, and only found a few...

In general I think tidy code is already quite clean ;=))

Anyway, thanks for thinking of tidy... there are some open issues you could work on...

Will leave this open for a while, in case others have a strong argument one way or the other...

geoffmcl:
Since no further comments for a few weeks, am closing...

Also the fork would need to rebase to master, since there are now conflicts that need to be resolved...

PR #249, opened 2015-09-08
fixed: tidy-html5 debug build fails on VS2015 x64

sria91:
The debug build fails when building on Windows 10 x64 using Visual Studio 2010 (64-bit build).

### DIFF
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -86,7 +86,7 @@ void TIDY_CALL TY_(filesink_putByte)( void* sinkData, byte bv )
   FILE* fout = (FILE*) sinkData;
   fputc( bv, fout );
 #if !defined(NDEBUG) && defined(_MSC_VER)
-  if (fout->_file != 2)
+  if (_fileno(fout) != 2)
     SPRTF("%c",bv);
 #endif
 }


geoffmcl:
@sria91 thanks for looking at this... a bit `sneaky` of MS for changing this structure, but...

But using a structure member name was very **sloppy** coding on my part!

In general the `FILE` structure should be considered opaque, and one should **not** relie on specific member names, which could changes, as it seems MS has done with stdio.h that comes with VS2015 (_MSC_VER == 1900)... changing the `struct _iobuf`...

Can you try the following patch which uses the `_fileno()` function. This patch works for my VS2010... and should be good for **all** MSVC versions...

```
diff --git a/src/fileio.c b/src/fileio.c
index a1c14c9..f2e11e7 100644
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -86,7 +86,7 @@ void TIDY_CALL TY_(filesink_putByte)( void* sinkData, byte bv )
   FILE* fout = (FILE*) sinkData;
   fputc( bv, fout );
 #if !defined(NDEBUG) && defined(_MSC_VER)
-  if (fout->_file != 2)
+  if (_fileno(fout) != 2)
     SPRTF("%c",bv);
 #endif
 }
```

That way we do not have to have the extra `#if (_MSC_VER) == 1900` switch code... and in case they change it again...

We could also use the POSIX compatible `fileno()`, minus the leading underscore, but in this case it is already inside a _MSC_VER macro... so is presently not compiled in other than Windows...

If you can test this use of `_fileno()`, and modify the PR, I would merge this... thanks...

sria91:
Thank you @geoffmcl. I've modified the code as you have suggested and tested the same using the scripts under the `test` directory. Following are the outputs.

Output of `test\alltest.cmd`

```...
Tidying 'input\in_837023.html'
R=1 C=1: Returning starttag node <html> stream
line 1 column 1 - line 1 column 1 - Warning: missing <!DOCTYPE> declaration
Entering ParseHTML...
R=2 C=1: Returning starttag node <head> stream
Enter ParseHead...
R=3 C=1: Returning starttag node <title> stream
R=3 C=8: Returning text TextNode [[ 837...]43 stream
R=3 C=51: Returning endtag node <title> stream
R=4 C=1: Returning endtag node <head> stream
Exit ParseHead 1...
R=5 C=1: Returning starttag node <body> stream
Enter ParseBody...
R=9 C=1: Returning text_node TextNode [Just ...]11 stream
line 9 column 7 - line 9 column 7 - Warning: discarding malformed <!DOCTYPE>
...
Tidying 'input\in_1072528.html'
line 1 column 1 - line 1 column 1 - Warning: discarding malformed <!DOCTYPE>
...
ERROR TESTS [ 837023 1072528] ...
```

Output of `test\acctest.cmd`

```...
FAILED --- test '13.10.1.1' not detected in file 'accessTest\13-10-1-1.html'

Accessibility Checks: Version 0.1

accessTest\13-10-1-1.html:1:1: accessTest\13-10-1-1.html:3:1: accessTest\13-10-1-1.html:7:1: accessTest\13-10-1-1.html:7:1: No warnings or errors were found.

About HTML Tidy: https://github.com/htacg/tidy-html5
Bug reports and comments: https://github.com/htacg/tidy-html5/issues
Or send questions and comments to: https://lists.w3.org/Archives/Public/public-htacg/
Latest HTML specification: http://dev.w3.org/html5/spec-author-view/
Validate your HTML documents: http://validator.w3.org/nu/
Lobby your company to join the W3C: http://www.w3.org/Consortium
FAILED --- test '13.10.1.1' not detected in above
Appears ALL tests ran fine ...
```

Output of `test\alltestc.bat`

```...
Tidying 'input\in_837023.html'
R=1 C=1: Returning starttag node <html> stream
line 1 column 1 - line 1 column 1 - Warning: missing <!DOCTYPE> declaration
Entering ParseHTML...
R=2 C=1: Returning starttag node <head> stream
Enter ParseHead...
R=3 C=1: Returning starttag node <title> stream
R=3 C=8: Returning text TextNode [[ 837...]43 stream
R=3 C=51: Returning endtag node <title> stream
R=4 C=1: Returning endtag node <head> stream
Exit ParseHead 1...
R=5 C=1: Returning starttag node <body> stream
Enter ParseBody...
R=9 C=1: Returning text_node TextNode [Just ...]11 stream
line 9 column 7 - line 9 column 7 - Warning: discarding malformed <!DOCTYPE>
R=9 C=7:
...
Testing 1069553, expect 0, got 0
Tidying 'input\in_1072528.html'
line 1 column 1 - line 1 column 1 - Warning: discarding malformed <!DOCTYPE>
R=1 C=1:
...
Completed all test... full output written to temptest.txt
```

Output of `test\html5\testhtml5.cmd`

```...
Done 30 files. See tidied output in tempout...
With NO WARNINGS or ERRORS! This is a *** SUCCESS ***.
```

##### Additional issue

   1. Executing `test\html5\cleanup.dat` erroneously deletes `test\html5\template.org.html`. It looks like the problem is in line 15 `@del temp* >nul`

geoffmcl:
@sria91 thanks for testing the use of `_fileno()`, but in the PR it seems you have just done that fix on **top** of your earlier fix...

So if I merged this, then we would get **two** commits ;=(( -

 - https://github.com/sria91/tidy-html5/commit/48cea8c221fe06bc92c283ba55f97cde7fed5951
 - https://github.com/sria91/tidy-html5/commit/a45f34ade59753e410e47c958ea3c54792a9e90f

I would prefer this be just one commit, taking the code from what it is now directly to using _fileno()... no intermediate step...

Also then there is this additional 3rd commit in the PR -

 - https://github.com/sria91/tidy-html5/commit/60883b8d1fa6f5135db063ca04ed701f3938010d

which in fact has already been applied to master... so may do nothing, but I would prefer it not be there...

As I see it you need to -

 1. backout/revert **both** your fileio.c fixes.
 2. regularly rebase your fork to htacg/master - important!
 3. apply a single simple _fileno() fix only.

The second item, the rebase, is something **all** forks need to do regularly, but the process is quite simple - from say here https://robots.thoughtbot.com/keeping-a-github-fork-updated,  or many other references -

 1. In your local repo create a one time remote reference calling it say 'upstream'
    - $ git remote add upstream git@github.com:htacg/tidy-html5.git
 2. Then Each time you want to update your local master branch:
    - $ git fetch upstream
    - $ git rebase upstream/master

This should give us a simple single line PR ;=))

#### alltest.cmd

Also thanks for the `test` info and testing... 

Coincidently I have been doing some tidying up in there... the above rebase will pull all those changes into your fork. I have added a little more informational output, and put it all in the 'temptest.txt' file...

And these tests are not meant to be run using the `Debug` version of tidy, which with MSVC adds a **lot** of extra debug only information...

So in running 'alltest.cmd' I do **not** get the two errors you show of ERROR TESTS [ 837023 1072528]??? When I run alltest.cmd these two have no problems - extract from my `temptest.txt` -

```
...
Doing: '..\build\cmake\Release\tidy.exe -f temp-5\msg_837023.txt -config input\cfg_default.txt        --tidy-mark no -o temp-5\out_837023.html input\in_837023.html  
Testing 837023, expect 1, got 1, msg temp-5\msg_837023.txt 
...
Doing: '..\build\cmake\Release\tidy.exe -f temp-5\msg_1072528.txt -config input\cfg_default.txt        --tidy-mark no -o temp-5\out_1072528.html input\in_1072528.html  
Testing 1072528, expect 1, got 1, msg temp-5\msg_1072528.txt 
...
```

So not sure why you got those two errors???

#### acctest.cmd

And I had not tried `acctest.cmd`, but in just trying that first had to correct the EXE name, and then it ran fine... That is I get **no** failed tests... Not sure why you show failed 13.10.1.1???

For me it ends with `Appears ALL tests ran fine`...

#### alltestc.bat

The 'alltestc.bat' is really just a **repeat** of 'alltest.cmd', with the addition of a compare of the output with the output in the 'testbase' folder... like `alltest.cmd`, it runs fine except for test 661606...

In both tests, test 661606 seems to show up as a `binary` compare???

```
Doing: 'diff -u testbase\out_661606.html temp-5\out_661606.html' 
Binary files testbase\out_661606.html and temp-5\out_661606.html differ
Got a DIFFERENCE between testbase\out_661606.html and temp-5\out_661606.html 
```

Hmmm, why is `diff` choosing this/these as `binary`... Other file compare utilities, like my own fc4, WinMerge, ... show the files as **exactly** the same...

They are encoded in `shift_jis` and apparently diff can not handled that... 

But I am convinced they are the same, so there is no problem here...

#### cleanup.bat

It has been a long time since I looked in the test\html5, and thanks for the heads up about `cleanup.bat`. It will indeed eroniously delete template.org.html. As a quick fix have renamed it to atemplate.org.html... 

##### Meantime, look forward to a clean PR for the fileio.c/_fileno() fix ;=))  thanks...

sria91:
Thanks @geoffmcl, I had made a second commit only to preserve the history. Though, as you have suggested, a rebase along with a single commit appears to be more appealing.

According to https://git-scm.com/docs/git-revert, using a `git revert <commit>` would revert the changes but register an additional commit. Then I would have to apply the _fileno() fix after doing a `git rebase upstream/master` and make one more commit. So, in effect I will be recording **five** commits which is **not desired**.

Hence, I presume you asked me to do a [`git reset <commit>`](https://git-scm.com/docs/git-reset) before doing a rebase and applying the fix. This will record a single commit which is **desired**. The steps used are as follows.

1. `git reset 50e69dbcb1bfdb9b68a6b1e8dfe9a1552c6b0556` reset to commit just before I made the changes
2. `git reset --hard` discard the changes
3. `rm test/report-5-1-8.txt` this file was added to tracking after my first commit. So had to be removed manually before rebase
4. `git rebase upstream/master`
5. apply the `_fileno()` fix and commit
6. `git push -f` force push (generally this seems to be partially discouraged)

#### alltest.cmd & acctest.cmd

It appears that the two errors w.r.t. files **837023, 1072528** and one error w.r.t **accessTest\13-10-1-1.html** occur only with the Debug build. The Release build indicates no error. I'm wondering why is it so?

#### alltestc.bat

In my system, running `alltestc.bat` gives the following output

```
...
Got a DIFFERENCE between testbase\out_500236.html and temp-5\out_500236.html
...
Got a DIFFERENCE between testbase\out_661606.html and temp-5\out_661606.html
...
```

And yeah both the files are the same when viewed in a text editor, except that the files in **testbase** (on the left) have different line endings `CR+LF` than the files in **temp-5** (on the right) `LF`.

out_661606

Using  [`diff -ua --strip-trailing-cr %TIDYBASE% %TIDYFILE%`](http://www.gnu.org/software/diffutils/manual/html_node/Binary.html) seems to solves the problem. (the -s option just prints when both the files are identical)

diff_out_500236_661606

The `--strip-trailing-cr` option is available in **diffutils 3.3** bundled with Git 2.5.0 for Windows :-). But it isn't avalable in *diffutils 2.7* bundled with GitHub Desktop 3.0.5.2 :-(.

Additionally `test/html5/testhtml5.cmd` line number 5 is `@set TMPEXE=..\..\build\cmake\Release\tidy5.exe`. Shouldn't it be `tidy.exe`

**Anyways, heads up for cleanup to the test scripts. The output is now pretty and readable. 
Now we have a single line, single commit PR ;-)
Hope you could merge the pull request now.
Thanks :-))**

geoffmcl:
@sria91 have now merged the PR... thanks for persisting with this... 

And yes, now that I think about it `$ git reset --hard` is the right thing to discard commits, not `revert`, which effectively adds a new commit...

And, yes, I have the older `diff` 2.7 which does not support the `--strip-trailing-cr`! 

But I find if I add the -a (treat as text) option diff finds them equal... This is only needed in my case on 661606, which as mentioned contains shift-jis characters... that is byte sequences like 82 BA 82 F1 82 D4 81... and is treated by git as `binary`... And as a binary file, there are no line endings changes done... 

The `testbase` is git maintained, and I remember a long time ago doing `$ git config --global core.autocrlf true` in my windows system which I thought made all downloads CR/LF, but only uploaded as a LF... or something like that...

And then the new output by the current tidy to `temp-5` should again have native windows default of CR/LF when run in windows... This is what I get for the files checked...

Except I now note 2 configs, a cfg_647255.txt with `newline: LF` and cfg_649812.txt with `newline: CR` so would expect tidy to honor that in those cases...

Just did a quick line-end check of 447 files in `temp-5` and see 446 have CRLF, and 1, out_647255.html, has UNIX LF, which sort of agrees with the first of the above. Why none with just CR???? Could be just a bug in my perl script...

On the other hand of the 447 files in `testbase` 444 have CRLF, and 3, out_500236.html, out_647255.html, out_661606.html with UNIX LF... 647255 was requested in config... 661606 is treated as binary, as mentioned above. Why 500236? Again maybe a scirpt bug???

Anyway, as stated, except for the line endings, the two ouputs of 661606 are the same... so no problem with tidy that I can see... 

Not exactly sure why the second, 500236, does not show up for me, as it does seems to have  different line endings! But again see no problem with tidy...

And like you, running the Debug version of tidy, tidyd.exe, does show some problems with 837023 and 1072528. Not sure WHY, but the debug output is massive, and hard to pin it down... but again I am not really concerned about this... but will still try to explore at some time...

And thanks for the test/html5/testhtml5.cmd fix. Have now changed and pushed that...

sria91:
@geoffmcl thanks for replying and of course merging the request.

It's odd that on my system `diff -ua` still indicates the two files are not identical, so I had to use ` --strip-trailing-cr`.

I can understand that it is not of much concern. Still I couldn't help but fix the code causing the problem with 837023 and 1072528. My recent commit on the forked repo 2ed0f6555c6972f82aff763675855e9cedcb5e98 makes it possible for the Debug build to work with the test cases without crashing :smile:. If you deem it as required now, it could be merged.

It was some great learning experience for me though. This being the first time me using git, GitHub, and Markdown. :smiley: :smiley:

Thank you again. :smiley:

geoffmcl:
@sria91, wow 4 smilies... must be doing something right ;=)) (my special smile!)

And thanks for looking into the debug crashes on 837023 and 1072528... 

What you have done would work fine, but I prefer to fix the Show_Node() debug service to handle a NULL node. It may be important to know that at that point a NULL node is returned, and with the GTDBG macro used in some 20+ cases, we need to protect against a NULL node in **every** case...

A bit of a quick messy patch, it is just debug stuff, but now I can run tidyd.exe, BUT it does **not** produce a clean `diff -ua testbase temp-5`, while there is no diff with the Release version. 

Apparently the debug version has some other slightly different output, but no time to exactly explore why...

It is sufficient that the Release version is clean... Thanks again...

PR #255, opened 2015-09-12
Dockerfile for tidy-html5

vielmetti:
My first pass at a Dockerfile for this project. It downloads the 5.1.9 release, brings in all of the dependencies, builds all of the binaries and libraries, creates documentation, and builds an RPM and DEB for the platform it's on.

A couple of todos are left:
- it should probably build from the git "master" branch instead of a specific release
- if you were doing this in production you'd strip a lot of stuff out of it
- the base image could be much smaller than Ubuntu

This does completely automate a build from scratch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+from ubuntu:14.04
+MAINTAINER Edward M. Vielmetti <edward.vielmetti@gmail.com>
+
+# install prerequisites
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    cmake \
+    curl \
+    zip \
+    xsltproc \
+    doxygen \
+    rpm \
+    git
+
+# grab the tarball
+WORKDIR /root
+RUN curl -L -o 5.1.9.zip https://github.com/htacg/tidy-html5/archive/5.1.9.zip && \
+    unzip 5.1.9.zip
+
+# do the build
+# build everything I possibly can
+
+WORKDIR /root/tidy-html5-5.1.9/build/cmake 
+RUN cmake ../.. && make 
+
+WORKDIR /root/tidy-html5-5.1.9/build/documentation
+RUN cp ../cmake/tidy ../cmake/tidy5 && ./build_docs.sh
+
+WORKDIR /root/tidy-html5-5.1.9/build/cmake 
+RUN make package && make install 
+
+# run the command
+CMD "/usr/local/bin/tidy"


geoffmcl:
@vielmett ok, this looks like a `script` build... No particular problem with that...

Some small points, ideas, comments -

First, as a suggestion! To me a PR should not be for your `first pass`, plus fixes that you required on testing... It should be the final, tested product, clean of say spelling mistakes... in this case one new file... no history of patches on it before it even arrives...

Others:

 1. `MAINTAINER` - should this not be the HTACG group? Yes, by all means add something like `
 # First added by Edward M. Vielmetti <edward.vielmetti@gmail.com> on...`
 2. `# install prerequisites` - well ok, if you really want to **mess** with someone else system... I'd prefer this 'switchable', at least to be **off**...
 3. `# build everything I possibly can` - Hey, WHY? As a user why would you want the documentation, and packaging??? They are of no use to him/her! No gain...

For 2. I do not know if it supports any `if-else-endif` logic, but my idea would be say in pseudo code -

```
#IF NOT ENV $DOUPDATE EQU NO
    # Update the system AND prerequisites
    RUN apt-get update && apt-get install -y \
    build-essential \
    cmake \
    ... etc ...
#ENDIF 
```

so I can do `export DOUPDATE=NO` to stop this...

And it would be nice of the list of `install` items was also externally configerable... and maybe current system dependence, if any needed...

On the latter 3. accepting to not build **eveything** would certainly reduce the `prerequisite` system invasion ;=))

To build the default, assuming you have cloned the source, why not just -

```
# change to the git source
WORKDIR $HOME/projects/tidy-html5/build/cmake
# and build tidy...
RUN ./build-me.sh
```

And you will note the `build-me.sh` chooses some defaults but has some user options... one of which is to generate the stripped release, production version...

And I certainly agree with with another of your TODO comments, namely using the git cloned  source, or even some other specific branch, other than the default `master`, like `release/5.0.0`, or even some earlier tagged versions, not just a **fixed** zip version...

Otherwise why is git installed? And if not using zip, then no `curl` install needed...

But either way, this `Dockerfile` does not really seem to be part of the tidy-html5 source repo... 

If you have already cloned the repo to get it, why is it then choosing to download another zip, **or** another clone, to a `WORKDIR /root` no less!... Does not really make sense to me...

Maybe this type of thing could be hosted somewhere else as `<...>/tidy-html5/Dockerfile`... as a sort of `do-it-all` script... if people want, need that... maybe added to http://www.html-tidy.org/#get_tidy... as an alternative 'full-download-build-install' script... 

Or maybe we need another repo, say `htacg/meta`, which could hold such `download-and-compile` scripts, and other `meta` data... I have an `updtidy-html5.sh` script which does the cloning, but could be expanded... or something...

So, for the reasons above, I personally see no Tidy! gain in this PR at this time... sorry...

Will certainly listens for other comments, ideas... 

Maybe I am missing something here? Maybe `Dockerfile` is the new `holy grail` ;=)) 

I do note it is **not** installed by default in my Ubuntu 14.04!

vielmetti:
Closing this pull request as it was rejected by the package maintainer.

PR #262, opened 2015-09-14
Change "tidy5" to "tidy" in documentation build script

vielmetti:
Remove last vestige of old "tidy5" binary name.

With this change, automated documentation builds will work from source.


geoffmcl:
@vielmetti, done and thanks... I certainly hope it is the last ;=))

As can be seen in the #173 i searched for a non-existant `build-docs.sh`, otherwise this would have already been done...

PR #278, opened 2015-09-28
Fix cmake for use as subproject.

KingDuckZ:
In CMakeLists.txt, replace <code>CMAKE_SOURCE_DIR</code> and <code>CMAKE_BINARY_DIR</code> with <code>CMAKE_CURRENT_SOURCE_DIR</code> and <code>CMAKE_CURRENT_BINARY_DIR</code> respectively.

This fix allows tiny to be used as a sub-project in a cmake project (ie: it can be included via add_subdirectory from a top-level CMakeLists.txt).

Also fix a warning about tidy-config.xml not being found (not sure if I caused it, but I think it should be fine now).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,15 +213,15 @@ if (UNIX)
         ## also could use `manpath` command output to determine target install path
         set(TIDY_MANFILE ${LIB_NAME}.1)
         message(STATUS "*** Generating man ${TIDY_MANFILE} custom commands...")
-        set(TIDY1XSL ${CMAKE_SOURCE_DIR}/build/documentation/tidy1.xsl)
-        set(TIDYHELP ${CMAKE_BINARY_DIR}/tidy-help.xml)
-        set(TIDYCONFIG ${CMAKE_BINARY_DIR}/tidy-config.xml)
-        add_custom_target(man ALL DEPENDS "${CMAKE_BINARY_DIR}/${LIB_NAME}")
+        set(TIDY1XSL ${CMAKE_CURRENT_SOURCE_DIR}/build/documentation/tidy1.xsl)
+        set(TIDYHELP ${CMAKE_CURRENT_BINARY_DIR}/tidy-help.xml)
+        set(TIDYCONFIG ${CMAKE_CURRENT_BINARY_DIR}/tidy-config.xml)
+        add_custom_target(man ALL DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME}")
  
         # run built EXE to generate xml output 
         add_custom_command(
             TARGET man
-            COMMAND ${CMAKE_BINARY_DIR}/${LIB_NAME} -xml-help > ${TIDYHELP}
+            COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME} -xml-help > ${TIDYHELP}
             COMMENT "Generate ${TIDYHELP}"
             VERBATIM
         )
@@ -229,7 +229,7 @@ if (UNIX)
         # run built EXE to generate more xml output 
         add_custom_command(
             TARGET man
-            COMMAND ${CMAKE_BINARY_DIR}/${LIB_NAME} -xml-config > ${TIDYCONFIG}
+            COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME} -xml-config > ${TIDYCONFIG}
             COMMENT "Generate ${TIDYCONFIG}"
             VERBATIM
         )
@@ -238,17 +238,17 @@ if (UNIX)
         add_custom_command(
             TARGET man
             DEPENDS ${TIDYHELP}
-            COMMAND xsltproc ARGS ${TIDY1XSL} ${TIDYHELP} > ${CMAKE_BINARY_DIR}/${TIDY_MANFILE}
+            COMMAND xsltproc ARGS ${TIDY1XSL} ${TIDYHELP} > ${CMAKE_CURRENT_BINARY_DIR}/${TIDY_MANFILE}
             COMMENT "Generate ${TIDY_MANFILE}"
             VERBATIM
         )
 
-        install(FILES ${CMAKE_BINARY_DIR}/${TIDY_MANFILE} DESTINATION ${MAN_INSTALL_DIR})
+        install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${TIDY_MANFILE} DESTINATION ${MAN_INSTALL_DIR})
         
         if (BUILD_DOCUMENTATION)
             find_program( DOXYGEN_FOUND doxygen )
             if (DOXYGEN_FOUND)
-                set( WRK_DIR ${CMAKE_SOURCE_DIR}/documentation )
+                set( WRK_DIR ${CMAKE_CURRENT_SOURCE_DIR}/documentation )
                 set( EXP_DIR ${WRK_DIR}/examples )
                 set( OUT_DIR ${WRK_DIR}/temp )
                 set( DXY_DIR ${WRK_DIR}/temp/tidylib_api )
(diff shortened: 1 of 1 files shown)


geoffmcl:
@KingDuckZ hmmm, reviewing the diff shows **many** lines changed by what looks like just space changes...

Could you please adjust your editor to not do this, to make it easy to see and review what is being changed... I expected just `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` substitutions...

That also made it hard to see what you fixed concerning `tidy-config.xml`... which may be a good thing... when I can see what you have done...

And then why is `build/documentation/tidy1.xsl` in your diff at all? In fact this particular file should be deleted, and only `documentation/tidy1.xsl` should be used... so could accept a change for that...

For the reasons above not prepared to merge this yet... can you address the issues raised?

KingDuckZ:
If it's ok with you, for me it's easier to drop trailing whitespaces in one commit and re-apply my patch on that. Conflicts should be resolved automatically that way, and you should see more clearly what the real patch is about.

KingDuckZ:
I see github picked up the new pull request by itself after I did a forced push, so you should see the two separate commits now.
I don't know what build/documentation/tidy1.xsl is, but it was causing the warning I fixed.
CMakeLists.txt referencese it from line 216 (before and after my patch) so that bit hasn't changed. The difference is that before you were using the .xsl from your working copy, now cmake generates it in the build dir after replacing the right path in it. The warning I was getting was "../cmake/tidy-config.xml" not found, which is because tidy-config.xml is a generated file afaict.

I'm not sure if before my patch it would find ../cmake/tidy-config.xml for some specific use cases, or if it was just always broken, but this fixes the warning and I think it does whatever the original code was trying to achieve.

geoffmcl:
@KingDuckZ first, let me say I have no problem with the cmake variable substitution... I have often wondered about exactly what is the difference between these two cmake variants... and if it lets tidy be a git submodule, and still be the sole, primary repo then no problem... that is **good**!

In fact, I would appreciate a few clues how to make tidy a submodule of another github repo  that uses tidy, like say edbrowse, and/or maybe others... just to experiment with how easy/hard this is... understand its benefits... pros/cons... document it... etc...

And yes, perhaps I should have warned you that if you just reverse the space change, github will just create yet another patch and join that to the others in the existing PR... 

While the final result in CMakeLists.txt is now clear, as just a `substitution`, the spaces changes will also be in the history...

That's not particularly a problem, but it would be better to return to the htacg HEAD in your fork and just apply just the substitution cleanly... and sometimes the only way to achieve that is to in fact delete your fork, and fork cleanly again...

And now I can see what the second, unrelated change is, and also this looks **very** good in principal, but...

You have changed `build/documentation/tidy1.xsl` to deal with the missing tidy-config.xml... you have renamed it `tidy1.xsl.in` and added a `configure_file(...)` to write it, doing the substitution of `@TIDYCONFIG@`... that is all ok...

But as mentioned, this is in the wrong folder in the first place...

The file `build/documentation/tidy1.xsl` needs to be deleted, and CMakeLists.txt adjusted to use the file `documentation/tidy1.xsl` or a `documentation/tidy1.xsl.in`, to do the name substituion... but this is nothing to do with this issue `Fix cmake for use as subproject`.

Can they be separated? 

Now take care here... github will always see what you do to in your `master` branch as a single PR to the upstream `master`... 

The only way I know to separate things is to either wait until other PRs are cleared, merged, and you have done a rebase, and push, or to do this 2nd change in a `branch` you create...

So, in summary -

 1. On the first, I could do this simple replacement, still attributing it to you in the message, or you could present a single clean PR for this, which will then have your name on the change...

 2. On the second issue, thanks for this, but can we separate it? 

 I do not want to make a change in `build/documentation` when later this will just be deleted. Seems creating history for nothing... Can we fix cmake to use `documentation/tidy1.xsl`, or better `documentation/tidy1.xsl.in`?

 3. And 3rd, on the suppression of trailing spaces even this should be considered... it has been discussed before a few times, like Code cleanup #59, but not done... but this too should be a separate issue, carefully timed to do **all** source files in one hit... with no **other** change what-so-ever...

 We should be able to do an external diff with -rw and not see a single `diff`...

Sorry for this carrying on, on simple issues, but I would like to keep the repo history as clean as possible, with issues clearly, and cleanly separated...

Really thanks for your efforts, but we are not quite there yet... can you help with a few more steps? How can I help to make that easier?

KingDuckZ:
Is this ok? There are no more whitespace changes and the fix for the warning is also gone. Either way as long as commits have a logical meaning I don't think having multiple commits each doing its thing in a single pull request is a big deal.
Also, git can easily handle millions of commits, so don't worry too much about the number of commits growing.

As an example on how you can use tidy-html5 as a submodule in another project, have a look at my project here https://bitbucket.org/King_DuckZ/duckscraper/src I think I'm doing exactly what you are describing in your post. I hope you find it interesting :)

geoffmcl:
@KingDuckZ this is now a single issue, and merged, having trust that this change of a CMake variable has no effect when tidy is a sole, primary repo... I think, hope not...

Thanks for this... merge done...

Do I have to explain why multiple commits are not particularly good for a single change, or why `git can handle millions of commits` is an argument for allowing more commits than neccessary to effect a desired change? I hope not!

Concerning using submodules you're pointing me to your repo that uses tidy (and it seems curl) as submodules... thanks for this... 

But this does not seem to help in the git command sequences that sets up these 3rdParty libaries as a submodule of `duckscraper`... and how to keep them updating on each build...

I can see there is a .gitmodules file containing, among other things -

```
[submodule "lib/tidy"]
	path = lib/tidy
	url = https://github.com/htacg/tidy-html5.git
```

That looks a good start, and I suppose creating such a .gitmodule file with these entries may help, but maybe there are git command to create this?

But then where is the `git pull 'tidy'` to make sure this source is the latest? And re-build if there is a change...

I have read some things about git submodules that are good, but I am still missing some basics about setting it up for other repos that depend upon the latest tidy... maybe you can help?

It seems maybe is starts with something like -
```
$ git submodule add https://github.com/htacg/tidy-html5.git
```

But then how to enable this submodule, and make sure it is updated, pulled, each time the primary project is built?

Sorry for these newbie questions on submodules, but all the reading suggests it is a good way to go if you depend on the latest 3rdParty source... 

But not quite getting it all...

But again thanks for your persistance and effort on this cmake gitmodule change...

KingDuckZ:
Thanks for accepting it. CMAKE_CURRENT_*_DIR are relative to the CMakeLists of the current project, whereas the non-CURRENT version are relative to the top-level cmake project. When you build tidy-html5 as a standalone, it is the top level project *and* the current project all the time, so these variables are equivalent. Refer to this for details: https://cmake.org/Wiki/CMake_Useful_Variables

As for adding submodules you guessed right, you create that git file by doing git submodule add. You can start a dummy project to try it out:

    git init dummy
    cd dummy
    git submodule add https://github.com/htacg/tidy-html5.git #this can also be a relative address
    git submodule update --init --recursive
    git status

I think you are being confused about the update mechanics though. Submodules don't update automatically. That makes perfect sense, because say I write my code to use a library the whose public interface changes at some point. I don't want to pull in the latest version automatically, or my code will break. Git instead ties your main repo to a particular commit of the submodules. That means that if you clone my duckscraper project and init the tidy-html5 submodule, you will always end up at the commit before my pull request until I update it. How do I update it? There may be better ways, but here is what I do:

    cd duckscraper/lib/tidy
    git fetch origin #origin in this subdir is github/htacg, not bitbucket
    git reset --hard origin/master #again, we're not affecting duckscraper's working copy
    cd ../..
    git add lib/tidy
    git commit -m "Update tidy-html5"
    git push --all

Practically speaking, git stores the sha1 of the commit of the submodules. When you checkout a specific version of your code, you also check out that file with the sha1, and then git checks out the submodules at the sha1 written in that file. You can see that eveen from [here](https://bitbucket.org/King_DuckZ/duckscraper/src/bf3b85498b6aed982662f26637b3d69f95512cf8/lib/?at=master) you can go to a specific commit of tidy-html5.

Also, if you are actively working in your submodules you can create branches, add remotes, rebase, do all the normal stuff.

geoffmcl:
@KingDuckZ thank you for the cmake link, and the explanation of git submodules... I will get around to experimenting with this some more...

Of course, when I think about it, updating a submodule should not be automatic... that should only be when I decide an update is needed... which could in turn be scripted to be done on each build if so desired... or not...

Looks like fun... thanks again...

PR #279, opened 2015-09-30
Use tidy.xsl from documentation/ instead of build/documentation.

KingDuckZ:
As we discussed, this is the second pull request to fix the warning during the build. I'm not sure I can delete the build/ directory yet. I tried and the project still builds fine, but I don't know if it's used elsewhere. If it can be safely removed, just let me know and I'll make one more commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,10 +213,14 @@ if (UNIX)
         ## also could use `manpath` command output to determine target install path
         set(TIDY_MANFILE ${LIB_NAME}.1)
         message(STATUS "*** Generating man ${TIDY_MANFILE} custom commands...")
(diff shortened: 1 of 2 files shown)


geoffmcl:
@KingDuckZ thanks for following through on this... now merged...

I too do not know whether the `build/documentation` can be deleted yet, but that is a separate issue...

The ultimate aim is to remove `build/documentation`, `build/win64`, and `build/cmake`, and the contents of the current `build/cmake` be just in `build`, but that can be further down the road...

Again thanks...

geoffmcl:
@KingDuckZ oops, missed that you had created a **new** `CMAKE_CURRENT_BINARY_DIR/documentation`! This seems just adding unnecessary depth, and more to clean up...

And then telling git to ignore this now built tidy1.xsl...

Still to fully test the API documentation build... but all looks good so far...  thanks...

PR #290, opened 2015-11-04
Fix indent-with-tabs for library use.

balthisar:
Posted as PR for code review.

When using libtidy the `config.c`'s `ParseTabs` (and other parsers) aren't used when setting properties with `tidyOptSetBool` (and other `tidyOptSet`ters), so this caused `indent-with-tabs` not to work. 

`tidylib.c` `tidyDocSaveToStream` are where most of the other option conditionals already exist so this seems like a likely home.

Fix verified using a `tidy.cfg` on the command line, and in Balthisar Tidy using `libtidy.dylib`.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1229,10 +1229,8 @@ Bool ParseTabs( TidyDocImpl* doc, const TidyOptionImpl* entry )
         Bool tabs = flag != 0 ? yes : no;
         TY_(SetOptionBool)( doc, entry->id, tabs );
         if (tabs) {
-            TY_(PPrintTabs)();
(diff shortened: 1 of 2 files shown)


balthisar:
I'll add, we should bump the version with this because it affects API behavior and libtidy users might want to check against it.

geoffmcl:
@balthisar this is a great patch... a good catch... thanks...

And for sure `tidyDocSaveStream` is the best home, since every `save` method uses this internal common output service, be it to a buffer, a file, or just tossed into the air...

Definitely a version bumper ;=)) one BUG less...

An absolute **must have**

balthisar:
Merged and version bumped.

Issue #295, opened 2015-11-11
Build with cmake 2.8.11 on Linux fails to install libraries into /usr/lib64


mcepl:
I am trying to update Fedora tidy to 5.0.0 version and current (very experimental, unfinished, etc.) https://mcepl.fedorapeople.org/tmp/tidy-5.0.0-1.fc22.src.rpm fails because `make install` puts libraries (`libtidy*`) into `/usr/lib/` instead of `/usr/lib64/` directory. Whatever magical CMake variables I am setting nothing leads to the desired solution.
Using cmake 2.8.11 on RHEL-7.
Build log is [build-5.0.0-1.fc22.txt](https://github.com/htacg/tidy-html5/files/31684/build-5.0.0-1.fc22.txt)

mcepl:
Also, even though I have `-DBUILD_TAB2SPACE:BOOL=ON` I don’t see `tab2space` installed in `/usr/bin/`.

geoffmcl:
@mcepl good to see someone looking after a Fedora build, install and packaging...

As you have found, and as your build log shows, lines 47-55, none of your `magical` Cmake variables do anything about this ;=()

The install sub-directory is controlled around [Lines 168+] (https://github.com/htacg/tidy-html5/blob/release/5.0.0/CMakeLists.txt#L168). You would need to change the `DESTINATION` to `lib64` if that is where you want it installed...

Now it would be possible to make this a `user` option, like say -

```
if (NOT LIB_INSTALL_DIR)
    set(LIB_INSTALL_DIR lib)
endif ()
```

Then you could add an option `-DLIB_INSTALL_DIR:PATH=lib64`. And likewise perhaps also a `-DBIN_INSTALL_DIR=bin`...

So then we could have -
```
    install(TARGETS ${name}
        RUNTIME DESTINATION ${BIN_INSTALL_DIR}
        ARCHIVE DESTINATION ${LIB_INSTALL_DIR}
        LIBRARY DESTINATION ${LIB_INSTALL_DIR}
        )    
```

Now this would then be relative to a `-DCMAKE_INSTALL_PREFIX=/usr`, which I note you do add...

And IIRC if the user gives absolute paths here, like `-DLIB_INSTALL_DIR=/usr/lib64`, then cmake will use that absolute path... need to experiment with that...

So this could be fixed, but we would be unlikely to push such a small cmake only fix back into `release/5.0.0`, making it `release/5.0.1`, especially since I hope we will have another release within a few months...

[...]

PR #297, opened 2015-11-11
Add tab2space manpage.

mcepl:
Certainly something better and more complicated could be done (e.g.,
generate the manpage from tab2space -h output), but it all seems
excessively complicated and unnecessary (I don't think there will be
many changes in this rather straightforward utility).

### DIFF
--- a/console/tab2space.1
+++ b/console/tab2space.1
@@ -1,0 +1,39 @@
+.TH TABS2SPACE "1" "March 2013" "tab2space" "User Commands"
+
+.SH NAME
+.B tab2space
+[\fIoptions\fR] [\fIinfile\fR [\fIoutfile\fR]] ...
+.SH DESCRIPTION
+.PP
+Utility to expand tabs and ensure consistent line endings options.
+.PP
+When the \fIinfile\fR is not specified or is set to \fI-\fR (dash) the input
+file is read from standard input.
+.PP
+This man page is rewritten from \fI\-help\fR output.
+
+.SH OPTIONS
+.TP
+\fB\-h\fR, \fB\-help\fR
+display the help message
+.TP
+\fB\-dos\fR, \fB\-crlf\fR
+set line ends to CRLF (PC-DOS/Windows - default)
+.TP
+\fB\-mac\fR, \fB\-cr\fR
+set line ends to CR (classic Mac OS)
+.TP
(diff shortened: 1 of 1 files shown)


geoffmcl:
@mcepl as you point out in #295 the local utility `tab2space` is **not** presently installed, so I do not think we need a man page for it! See the cmake comment on [LN192](https://github.com/htacg/tidy-html5/blob/release/5.0.0/CMakeLists.txt#L192)

In other words we are **not** offering a universal utility to the world at large, supported by us like tidy, so the `BUILD_TAB2SPACE` option should be `off`, the default, when building distribution packages.

It is only for tidy developers who might have not adjusted their editors to use spaces instead of tabs... it was a quick way to convert the file to conform to the tidy `no tabs` code style.

I would go on to add, while the previous SF tidy maintainers had a `horror` of tabs in the source, I feel the current maintainers, at least me, are not so strict. And at this moment there are a few tabs in one or two source files... but there are other reasons we do not immediately run them through this purely local utility...

Of course, I still try to check source code changes by others for massive tabulation, and maybe fix that, but like I say, does not seem so important these days...

So thanks for the PR, but I think not at this time...

Will leave this open for a while in case others want to comment...

mcepl:
I am a Pythonista in my heart so I consider tabs to be a tool of devil, however it seems to me that carrying a special tool just for the conversion (when ``sed -e 's/\t/    /g'`` works just as well) is a bit excessive. However, I didn’t want to leave any patches in Fedora repo which couldn’t be considered upstream (see also #298).

geoffmcl:
@mcepl ok, had to look up what is a `Pythonista` ;=))

Of all the code/scripts I have written, over so many years, Python is probably the least... but with its `indent` **has** meaning, to denote context, I can see why `tabs` may not be your friend.

I too thought this utility **a bit excessive**, so when I wrote the CMake script, I made it optional, and defaulted it to `OFF`.

And as mentioned directly in #298, just a brief delay on what `looks` like a good patch... but now merged...

So, am closing this. Please feel free to re-open, or comment

mcepl:
So, isn't The Right Thing™ to get rid off ``tabspace`` utility itself, when you don't want ti?

geoffmcl:
@mcepl certainly removing it entirely is an option... but wait, it is an option already ;=))

PR #300, opened 2015-11-13
allow modified install paths

rdieter:
and make them consistent between static/shared builds.

Long term, consider using:
https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

issue #295

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,6 +138,15 @@ if (MSVC)
     list(APPEND LIBHFILES ${SRCDIR}/sprtf.h)
 endif ()
 #######################################
+
+if (NOT LIB_INSTALL_DIR)
+    set(LIB_INSTALL_DIR lib${LIB_SUFFIX})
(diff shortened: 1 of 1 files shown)


geoffmcl:
@rdieter thanks for this... will certainly look at it...

What about the idea of adding INC_INSTALL_DIR?

rdieter:
indeed, I can add support for INCLUDE_INSTALL_DIR too

geoffmcl:
@rdieter thanks...

@mcepl what do you think of this?

Seems in your case would only need to add `-DLIB_SUFFIX=64`, which you already have...

mcepl:
> @mcepl what do you think of this?

Certainly @rdieter knows (comparing to me) what he is talking about with regards to ``CMake``.

geoffmcl:
@rdieter @mcepl now merged... thanks...

As this in only a cmake change, did not bump the version

PR #319, opened 2015-11-23
Fix a tiny buffer overflow.

CMB:


geoffmcl:
@CMB thanks have merged it...

But was not able to use 'wget' to get the page you showed, like I usually can do...

edbrowse (in windows) did better and I got -

> edbrowse  https://www.indiegogo.com/projects/translation-of-the-book-play-to-live-volume-6
no ssl certificate file specified; secure connections cannot be verified
1139
warning: http error 416, Requested Range Not Satisfiable
4438
957

I did that before and after I merged but no crash here in either case, despite the obvious over-run before...

Hope you get a chance to update tidy to 5.1.26 and test again... and thanks again...

CMB:
@geoffmcl Thanks Geoff.  You can wget the page from here as well:
http://the-brannons.com/broken.html
Sorry I wasn't able to pair this down to something smaller.  In this
case, it was easier to just find and fix the bug.

geoffmcl:
@CMB Thanks Chris. Yes, the wget worked fine on [broken.html](http://the-brannons.com/broken.html)

It turned out we just needed the line 3 `<head prefix="og: http://ogp.me/ns# etc... >` to test and exercise that allocation. So was able to reduce that 900 line, 150 KB file to a 12 line, 285 byte sample -  http://geoffair.org/tmp/in_eb10-2.html - that exercises that particular piece of allocation code, now fixed.

I suppose I could add that sample to our test cases, as a RDFa sample, and one that exercises at least the `prefix` attribute. Will try to get around to that, maybe expanding it to do some more RDFa testing...

Meantime I installed `valgrind` in my Ubuntu, and ran tidy 5.1.26 on my in_eb10-2.html file, and it reported `0 errors`, compared to an `invalid write of size 1` for tidy 5.1.25, so I hope all fixed. Thanks again...

Issue #321, opened 2015-11-25
Infinite loop


tspivey:
Originally found when trying to tidy http://www.fimfiction.net/

Pared it down to this which seems to loop.
```
<!doctype html>
<html>
<body>
<ul>
<a><li></li></a>
<a href="www.example.org"><li></li></a>
</ul>
</body>
</html>
```

benkasminbullock:
Commits prior to 2388fb017565bf09b65a400175de800154909955 don't have this problem, so the issue was introduced in that commit.

benkasminbullock:
I think this fixes the problem:

https://github.com/benkasminbullock/tidy-html5/commit/71d963844897a6864097d4bc36501a97ff314e40

Can someone check that and pull if it is OK?

geoffmcl:
@tspivey thanks for reporting...

@benkasminbullock added your patch and can confirm this removes the infinite loop from this sample...

Can you do a PR? I tried the usual fetch from your fork and cherry pick the commit to keep you as the author but this failed! Or I can just push the patch?

CMB:
@geoffmcl Something like the following should work to pull and apply
the patch.  Current directory assumed to be the tidy-html5 source:

wget -O commit.patch https://github.com/benkasminbullock/tidy-html5/commit/71d963844897a6864097d4bc36501a97ff314e40.patch
git am commit.patch

And voila, patch applied, with proper credit.

I can also confirm that the patch from @benkasminbullock fixes this.

benkasminbullock:
You can also pull the patch into a local repository using

    git pull git@github.com:benkasminbullock/tidy-html5.git

then push it to the github repository using

    git push

or whatever command you usually use to update the github repository from your local copy.

geoffmcl:
@CMB thanks for that option... will keep that in mind as an alternative in future...

@benkasminbullock well I tried a similar approach -

```
$ git fetch git@github.com:benkasminbullock/tidy-html5.git
$ git log # to find the commit I wanted - this failed!
$ git cherry-pick commit
```

So as indicated this failed in this case! Not sure why? It usually works. `$ git pull` would auto merge **all** the fork commits, which in this case was only one, so would have worked fine...

But thanks for the PR #322, now merged... and version pushed to 5.1.27

geoffmcl:
@tspivey if you get a chance to pull 5.1.27, and if fixed, close this... thanks...

geoffmcl:
Will close this... feel free to reopen, or add a new issue...

PR #322, opened 2015-11-25
Don't push back non-A tokens.

benkasminbullock:


geoffmcl:
@benkasminbullock I do not yet know if this is the complete solution to this, but certainly we now have 2 reports in #321 that it works, so that is good enough for now... patch merged, and thanks...

PR #323, opened 2015-11-26
Allows null value css-prefix to be used in a config file without issu…

balthisar:
…ing a warning.

Fixes a case when an auto-generated tidy config file with a null `css-prefix` generated a warning about a malformed option value. Other tidy strings can accept null values.


geoffmcl:
@balthisar looks fine to me...

PR #325, opened 2015-11-28
Progress callback

balthisar:
Adds a callback to the pretty printer for tidylib users to report where items in the source text ended up in the destination text.

A potential use case is for GUI applications to synchronize scrolling of "before" and "after" text editors.

### DIFF
--- a/include/tidy.h
+++ b/include/tidy.h
@@ -631,6 +631,9 @@ TIDY_EXPORT Bool TIDY_CALL tidyInitSink( TidyOutputSink* sink,
 TIDY_EXPORT void TIDY_CALL tidyPutByte( TidyOutputSink* sink, uint byteValue );
 
 
+/****************
+   Errors
+****************/
 /** Callback to filter messages by diagnostic level:
 **  info, warning, etc.  Just set diagnostic output 
 **  handler to redirect all diagnostics output.  Return true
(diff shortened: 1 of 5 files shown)


geoffmcl:
@balthisar pulled this `progress_callback` branch, and can confirm it compiles in Windows, no problems... 

Concerning your use case, for sure it does give back an `interesting` mapping between the input `node->line`, `node->column` to now an output line number, which could be used in such a `synchronized` scroll view in a "before" and "after" GUI compare... **neat**...

Note you do not add this for `PPrintXMLTree`! But maybe not needed for that output?

Do not yet have any app to test this, but will consider something...

So it compiles, and does **feel** like a new feature ;=)) Go for it...

geoffmcl:
@balthisar just for fun - love coding! - wrote a [tidy-view](https://github.com/geoffmcl/tidy-view) gui app implementing your progress call back to match lines in the `left - before` view and add them to the `right - after` view of a html file.

As with [tidy-gui2](https://github.com/geoffmcl/tidy-gui2) I use Qt4 to do the windows in a cross platform way. In general I think you will find that Qt implements quite a good OS specific look and feel, like using native file selection dialogs, etc... Anyway just another reason to install Qt4 and give it a try...

Seems the input line value is not always updated correctly by tidy - may be a bug? - but generally works... sort of... Have not yet done sychronised scrolling, but could... And because I use a QTextEdit can not yet see how to easily align the left/right lines... would need to pad the data with newlines to get it all lined up??? feels quite difficult... 

At present the color coding is for C/C++ text, so that is another TODO - implement HTML specific color coding.

Anyway, just a FUN tidy viewer ;=)) Usage $ tidy-view index.html

balthisar:
I'll try to give it a spin this weekend (Qt, etc.).

My plan won't be to use blank lines to line them up and synchronize them but to find the container coordinate of the middle line of the container being scrolled, and telling the other contain to scroll into view using the coordinate of the related line. I'll be using Cocoa, and although I don't know anything about Qt yet, I'm certain that it must provide similar functions.

I've been working on the localization project for Tidy so I've not actually built a Cocoa app to demonstrate this yet!

geoffmcl:
@balthisar yeah, great idea - just synchronize on the middle line. This should not be too difficult.

I have another very powerful gui file compare, [winmerge](http://winmerge.org/?lang=en), and my own dc4w, similar purpose, which do line them up line by line, but these use a list type display so is easy to insert blank lines... and that was what I was thinking of...

But yes, just the middle line, lined up, would look good...

And for sure Qt has that capability... lots of example code... hope you get a chance to try it...

Still to get a chance to look closer at your tidy localization, but will get there...

PR #328, opened 2015-11-29
Localized strings

balthisar:
- Added a new means of retrieving Tidy strings with an eye toward localization.
- Sample string is added to demonstrate functionality.
- There may be some edge cases to adjust to later.

Don't actually merge this, please!

In this "preview" I've written a substitute string retrieval system that can be used to localize Tidy. Currently the only output merged to the new system is the missing semicolon entity error, but I imagine all of the output can transitioned. I've also replaced a lot of the defines with an enum, and will probably do the same for some of the other sets, as well as add lookup keys for them (currently some of them are used as bit fields, so I can't use their values to look up strings).

In the long run localize.c probably becomes message.c (it looks like it may have been originally), and all strings are loaded from a common source.

The approach I took might seem a bit odd at first, and it sacrifices library size as languages are added for the simplicity of not having to curate separate language dylibs. String lookup speed should be nearly identical to current methods but could be optimized in the future (someone early this year added a lot of hashing functions to libtidy).

Anyway, have a look.


balthisar:
As of right now, there are no more loose strings at all in messages.c (né localize.c). I've yet to scour the entire library for stray strings that might be output, but there's the obvious console application that I might get around to tackling if we decide to move forward.

Also I hope some of our international users take an interest in helping us localize, since this will make it really very easy to do so.

Still have to make the library actually switch languages, though, when used with `--language`. Autodetecting the locale is best suited for the console application; the library should only be told what to do.

The sample en_gb is low-hanging fruit. I can go through and change all occurrences of "color" to "colour" for example, and there may be some -izes to -ises changes available.

benkasminbullock:
Possibly worth noting that "tidy_out" can be made static. This might help to "localize" the strings.

balthisar:
>Possibly worth noting that "tidy_out" can be made static. This might help to "localize" the strings.

It's declared static in the forward declaration, so that should make it static in the actual implementation.
I'm actually a little bit more worried about all of my string pointers, particularly in the console application. Clang's analyzer says I'm okay, but I don't trust myself, as strings in pure C without a library used to cause me a great deal of headaches!

Anyway I've taken a cursory glance at _all_ of the remaining source code. It seems that other than debug strings and strings that can't/shouldn't be localized, right now the task of getting all of the strings into the 'en' localization is complete. There's still a strategy issue I'll add to the next comment (just to keep things separate).

balthisar:
So my original idea was to tie into setting the `language` option, but a more careful look at the source code makes me think (1) this is a stupid idea, and (2) we should delete the `language` option entirely.

Addressing point (2) first, I can't think of a single use case for its existence. It's tied to `SUPPORT_ASIAN_ENCODINGS` but actually has no function, and for point (1) we obviously can't tie the output language to a compiler conditional. It also seems like a bad idea to specify a program's language as being an "option."

This leaves us with a couple of options, but I'm not sure what expectations that `libtidy` users have over libtidy's behavior. On systems I develop on (Mac OS X) frameworks automatically support the user's locale if the framework includes that localization, regardless of the application using the framework (a framework is an abstraction for a set of libraries).

• Should libtidy, then, try to determine its current locale and set the output language accordingly? Or,
• Should libtidy use the default `en` and only switch output languages when told to by an implementing application?

I kind of favor the second option is it gives the application developer a choice.

Then what about the console application? I think it should automatically adapt to the locale, perhaps with an option to force the default English.

Perhaps this whole thing is short sighted in that we don't actually _have_ any localizations available yet, but I'm hopeful we can attract international volunteers when it becomes essentially effortless to implement a localization (which the current state of this experimental branch manages quite effectively).

Anyway, this is a pretty significant change to some of Tidy's architecture, so some feedback is welcome.

geoffmcl:
@balthisar sorry I have not had a chance to try this branch, so do not have specific comments now, but some generalizations only...

For me there is always a concern about Windows support! I suppose it depends how it is done, but in quite a number of projects have always had trouble with the unix `locale` implementations... NSL_LANG or is that a different thing?

The simple way `edbrowse` does it does work everywhere, including in windows. They have a `translated` file for each set of messages in each language, naturally defaulting to English if no user preference given. It does bloat the app/library, having each language set built in, but maybe this is not a big problem... and allows dynamic changing, if that is needed, desired...

It does become a distribution addition if the languages are in a separate library, or in *.po files (gettext) implementations... again is that the same or different?

Naturally I do agree Tidy should be `international`, and it depends on finding people to do each of the translations... and making that **easy** for them to contribute...

Again in the edbrowse case, they already have 6 message files done... they have each in a single file in the source, and use a cmake run perl script to combine each file into one,, msg-strings.c?, for the compile...

Anyway will naturally comment more when I have had a chance to catch up a little, and give it some thought... and look at some other projects to see how they have done it... etc...

balthisar:
@geoffmcl,

The last thing I'm still investigating is how to determine locale. Platform-specific implementations are just an `#ifdef` away, but every platform approaches it differently. The design question then becomes do we do this in the library, or have library users implement it in their application and simply tell `libtidy` which language to use?

The way I implemented the languages uses an integer key to retrieve a string, and I've also made sure that a lot of old `#define`s are now enums (where possible), and changed some enum starting values where not accessible outside the library. Unfortunately some of them are exposed and rather than break things, I've done workarounds (like mirroring the existing enum).

Adding a language is merely a header file, and adding the structure defined in the header file to a languages array in language.c. Nothing at all has to be done to CMakeLists.txt, unless we want to export the individual language headers in the lib.

This does mean that every language becomes part of the binary. Right now the English header file is 2K, so each language would add less than 2K to the library. This was a deliberate choice to avoid having to build language libraries and installing them, too.

balthisar:
Current state:
Tidy will set the locale automatically assuming that it's set in your OS. In English my default local is en_US, but when I switch to Spanish my default locale is C. When setting my OS to English UK it switches back to en_US, but this is easy to change for testing on Linux/Unix (export LC_ALL, export LANGUAGE). Not sure about Windows.

Whatever language the library is set to Tidy will try to fallback intelligently. Supposing both es and es_mx are available, if a string is unavailable in es_mx it will be sought in es, and if unavailable there the always present en "master" string will be retrieved.

On Mac OS X Tidy 5.1.24 as installed by our HTACG installer is 549K. Adding en_gb (mostly) and the sample from es increase it to 751k, but this is using XCode instead of CMake, so I may have left on debugging symbols or something. (Actually I may not have specified "release" in the latest HTACG installer).

Assuming there are no glaring bugs, about the only thing I would do is add a bunch headers for common languages, and "localize" the non-quiet output with an additional message at the end asking for users to help contribute a translation!

balthisar:
Current state is looking pretty good. I eliminated all of the memory leaks related to my previously poor string management, and greatly improved the `--help-option` output. I suppose `--help-option` could be decoupled from the localization support effort, but it seemed worth it to get it in now as it was a great way to experiment with long, localized strings.

It had to get into a lot of Tidy's source files to be all encompassing, and so it would be great it library and command line users could start testing this branch. I'll keep it current with **master** so that bug fixes are included in this branch.

gagern:
> Naturally I do agree Tidy should be `international`, and it depends on finding people to do each of the translations... and making that **easy** for them to contribute...

That might be helped by using some well-established message file format, like e.g. [GNU gettext PO files](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files). That way, experienced translators could stick to [the tools they know](http://translationproject.org/html/software.html) to write their translations.

If you want to use `.h` files instead of the gettext library and format, that's fine since it should be easy to cook up some tool to convert between these formats. Essentially all you have to do is switch between enum constant and english string as the key for a message, and perform some slight changes to escaping and surrounding boilerplate.

One thing e.g. gettext handles that the current implementation with its simple lookup table would have problems with is [plural forms](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html#Plural-forms) from different languages. It is quite surprising how much these vary!

balthisar:
@gagern said:
>If you want to use .h files instead of the gettext library and format, that's fine since it should be easy to cook up some tool to convert between these formats.

Ah, this isn't a bad idea at all. I'd looked at GNU's library's and thought that they were overweight for Tidy's meager needs, but a quick and dirty conversion script between PO files and headers is simple enough to do.

I especially like the PO file because it's so _very easy_ to turn them into .strings files that I use on Mac OS X.

I'm working on regression testing scheme right now during the holiday downtime, so I'll look at adding PO support after a week or two. Of course PRs are welcome in the meantime.

geoffmcl:
Wow, as a primary windows user, I really do not know 
if I should weight in on this (unix/MAC) discussion!

===
Re: Use of the GNU gettext library

I absolutely understand the `norm` I see in unix is
`getext/PO`, ie PO translation files... **BUT**, I have 
yet to see this **well** implemented in a Windows 
environment! 

Yes, I have seen that there are some binary install 
options for windows, but the last time I checked, the 
GNU GetText for Windows -
http://gnuwin32.sourceforge.net/packages/gettext.htm
is a 10 years old project... maybe it still works,
but I for one will probably **NOT** install such an 
OLD system... 

And that is only 32-bit! What about if I want a 
64-bit install, or heaven forbid, a UNICODE 32 
and 64 bit version?

I hope you really understand gettext() has some 
serious dependencies... especially in Windows...

Do you have a link to something more recent?

And all that is in CMake terms... Yes, there is a
FindGetext.cmake in the distribution, and I `pass`,
having somehow installed `msgmerge`, `msgfmt` executables!

So, for instance, for the almost defunct [fgrun](https://sourceforge.net/projects/fgrun/) project,
I can successfully build the `gettext/PO` files... but 
I wonder how many other Windows developers have that?

Even I do not remember well how I got there!

For sure, I want internationalization of Tidy!!! Who 
in their right mind would be against that!

But it is not just finding the people to provide 
the message 'translations', but how to implement 
that in a cross-platform way, **WITH** the minimal 
of additional dependencies, especially in Windows.

But this is still mute at this time since we only 
have one translation, and that is English... I 
suppose we could easily mess up an UK/USA difference...

I look forward to taking the time to checkout this 
`localize` branch, just to check, show that what has 
been implemented still works fine for that language
in Windows... 

At present language.c fails to compile in Windows,
but I am sure this can be easily fixed... yes, the 
use of the gcc only `.` structure init... 
https://github.com/htacg/tidy-html5/blob/localize/src/language.c#L34
easily fixed... do not use that gcc ONLY struct init 
form... use what works in K&R C...

But on the broader topic, it seems when we have 2 or 
more translations would be the time to reconsider the 
possible **cross-platform** options...

That might include having cmake find_package(Getext),
but also having a backup translation plan if that 
fails...

Tidy should support **every** language under the sun!
As said, finding, encouraging, helping text translators
give us alternatives... 

If they want, need, the text in some familiar, easy to 
use format, then I for one would help write some scripts 
to produce that format, and when it is back, into what 
tidy is currently using, if required...

Let us **Go International**

balthisar:
I'll modify the struct to K&R.

I share a lot of the same concerns. My thought was to use the PO files _only_, and write a shell script to convert them back and forth to the header format, _not_ in an automated fashion and _not_ as part of the build process. This would provide a familiar format for the people willing to translate, while keeping things simple and as cross-platform as today.

geoffmcl:
@balthisar thanks for the quick K&R struct fix for language.c ;=)) That now compiles with MSVC...

The next thing is tidy.c! For a long time now gcc has allowed 'late' variable declaration, even after a code statement, like in C++...

MSVC sticks to the original K&R where, in any given context, all variables MUST be declared before code statements.

I tried for find a gcc dialect option that reverted this behavior, but found none! 

Thought `-ansi`, which is `-std=c89`, might do it, but that undefined `uint` which is used everywhere ;=(( 

So there seems no way to **check** this using gcc... maybe I missed something...

So you can see the diff, attached below is a tidy.c.diff. As can be seen it is just a case of bubbling variable declaration to near the top of the context. Does not change anything else (I hope!)...

And yes, I have no problem using the PO format to store the translated strings, and using a script to get them ready for a compile...


```
<tidy.c.diff>
diff --git a/console/tidy.c b/console/tidy.c
index 35ba8e6..f5344ea 100644
--- a/console/tidy.c
+++ b/console/tidy.c
@@ -275,7 +275,7 @@ static const CmdOptDesc cmdopt_defs[] =  {
  */
 static tmbstr stringWithFormat( const ctmbstr fmt, ... )
 {
-	va_list argList = {};
+	va_list argList;
 	tmbstr result = NULL;
 	int len = 0;
 	
@@ -1060,6 +1060,8 @@ static tmbstr cleanup_description( ctmbstr description )
 	states state = s_DATA;
 	charstates charstate;
 	char c;
+	int j = 0, k = 0;
+	transitionType transition;
 
 	/* Process the HTML Snippet */
 	do {
@@ -1088,9 +1090,6 @@ static tmbstr cleanup_description( ctmbstr description )
 		}
 
 		/* Find the correct instruction */
-		int j = 0;
-		transitionType transition;
-
 		while (transitions[j].state != s_LAST)
 		{
 			transition = transitions[j];
@@ -1130,7 +1129,7 @@ static tmbstr cleanup_description( ctmbstr description )
 					case a_EMIT_SUBS:
 						name[i_name] = '\0';
 						i_name = 0;
-						int k = 0;
+						k = 0;
 						writer = "";
 						while ( replacements[k].tag )
 						{
@@ -1357,6 +1356,7 @@ static void xml_options_strings( TidyDoc tdoc )
  */
 static void xml_strings( void )
 {
+    uint i;
 	ctmbstr current_language = tidyGetLanguage();
 	Bool skip_current = strcmp( current_language, "en" ) == 0;
 	Bool matches_base;
@@ -1364,7 +1364,7 @@ static void xml_strings( void )
 	printf( "<?xml version=\"1.0\"?>\n"
 		   "<localized_strings version=\"%s\">\n", tidyLibraryVersion());
 
-	uint i = tidyFirstStringKey();
+	i = tidyFirstStringKey();
 	do	{
 		printf( "<localized_string id=\"%u\">\n", i );
 		printf( " <string class=\"%s\">", "en" );
@@ -1423,6 +1423,7 @@ int main( int argc, char** argv )
 	ctmbstr cfgfil = NULL, errfil = NULL, htmlfil = NULL;
 	TidyDoc tdoc = tidyCreate();
 	int status = 0;
+	tmbstr locale = NULL;
 	
 	uint contentErrors = 0;
 	uint contentWarnings = 0;
@@ -1433,7 +1434,6 @@ int main( int argc, char** argv )
 //	tidySetPrettyPrinterCallback(tdoc, progressTester);
 
 	/* Set the locale for tidy's output. */
-	tmbstr locale = NULL;
 	locale = tidySystemLocale(locale);
 	tidySetLanguage(locale);
 	if ( locale )
</tidy.c.diff>
```

So now the `localize` branch compiles in WIN32. using MSVC... Would push this change if ok... But still to do some actual tidy testing...

balthisar:
@geoffmcl, looks okay to push the change.

I'll try to remember to stick to K&R... I'm mostly following the existing style. I wonder, though, MSVC doesn't support C99?

I'm using Clang, so I know there's an option somewhere to prevent C99 and newer; I'll make sure I set that in my environment. It largely implements GCC's front end, but it wouldn't be a different compiler if it weren't different. ;-)

balthisar:
The  PO<->HEADER back and forth converter is working pretty well. I _really_ like the tools that are available for PO files, and I thank @gagern for really making the point stick that it's a format with wide support.

Strategically speaking I don't see the value in using gettext directly, and the current proposed solution works well across platforms and keeps the file size small, and most importantly, portable!

I'm certain that we could make the effort to automate the PO file conversion as part of the build process, but I'm going to vote against that, too. It doesn't take a huge effort to generate the header using the converter tool, and I don't predict that it will be a major time sink.

On a side note, as part of testing I've been using a program called "POEdit" that is really, really just awesome and provides really good suggestions in Spanish. I think I can offer a full Spanish localization in the next week or two. I'm not a native speaker, but I'm good enough that the machine translation makes me really, really fast. I may look for a simple way to convert Mac .strings files so I can use this tool to localize my pet Mac application, too.

gagern:
> I *really* like the tools that are available for PO files, and I thank @gagern for really making the point stick that it's a format with wide support.

I'm glad my comment turned out to be this useful.

> One thing e.g. gettext handles that the current implementation with its simple lookup table would have problems with is [plural forms](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html#Plural-forms) from different languages. It is quite surprising how much these vary!

One possible approach to that is the following: for every string with a plural distinction, reserve a range of consecutive enum constants. E.g. `STRING_ERROR`, `STRING_ERROR_PLURAL1 = STRING_ERROR + 1` and so on. There doesn't seem to be a reasonably common language with more than 6 different plural forms, so `STRING_ERROR_PLURAL5 = STRING_ERROR + 5` could be the last. That way, you can build something that works essentially like [`ngettext`](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html#index-ngettext) in that it selects the appropriate plural form for the current language. In the PO file, the plural forms should of course be represented using [the appropriate PO format](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#index-plural-forms_002c-in-PO-files).

You might want to be specific up front whether you support [argument reordering](https://www.gnu.org/software/gettext/manual/html_node/c_002dformat.html) for printf-style format strings. Not doing so may force translators to choose clumsy formulations in order to stick to the english order of substitutions.

I think you should make the `.pot` file (translation template) readily available to likely contributors, e.g. by committing them to git. You should also avoid absolute paths in there. I'm not yet decided whether it would be wise to add the [`c-format` flag](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#index-c_002dformat-flag) to strings containing `%` as part of the automatic conversion. Doing so might be a useful hint to translators, and may enable additional helpful features in some tools. Getting this wrong would be rather confusing, though, so you should only do this if you are sure the heuristics are correct.

balthisar:
@gagern, this is really like going down a rabbit hole! I didn't intend to become gettext guru, especially given that we're against actually _using_ gettext, but I'll look into the items you mention.

I'm not keen on re-writing all of gettext's functionality (for determining plural forms, etc.). It looks like most of gettext is LGPL, so maybe there's a case for bringing the code in directly if it's not too large. Tidy is portable first, and we're not likely going to release anything that we can't compile and distribute ourselves. Dependency hell sucks!

gagern:
> I'm not keen on re-writing all of gettext's functionality (for determining plural forms, etc.).

The functionality for determining the appropriate plural form is a single C expression which is part of the PO file itself. So as long as you provide the flexibility of choosing among a varying number of plural forms, picking the right one would be fairly easy.

> It looks like most of gettext is LGPL, so maybe there's a case for bringing the code in directly if it's not too large.

If you can use the gettext sources unmodified, you could include them as a git submodule. See [here](http://savannah.gnu.org/git/?group=gettext) for its git repository. Note that it in turn uses [gnulib](https://www.gnu.org/software/gnulib/), a sourcecode-only compatibility library, as a submodule, so you may need to operate on git submodules recursively. By the way, properly using gnulib should be enough to provide argument reordering support on pretty much any platform.

The gettext build system is designed around the [GNU autotools](https://en.wikipedia.org/wiki/GNU_build_system). This (particularly in combination with gnulib) will provide a lot of portability for *nix systems. But building things on Windows [will likely require Cygwin or similar](http://git.savannah.gnu.org/cgit/gettext.git/tree/README.woe32). I'd consider that a reasonable requirement, but it's up to you whether you are willing to accept that. I'd be slightly worried that path name conventions on Windows differ radically from [those on *nix](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). But the [`--enable-relocatable`](http://git.savannah.gnu.org/cgit/gettext.git/tree/INSTALL) switch to `configure` may help there.

I also think that gettext is designed around the `LC_MESSAGES` facet of the current locale. Switching language per document or per thread is likely not well supported. If you want to offer such functionality, you might have to mix gettext with your own code. Since [its implementation](http://git.savannah.gnu.org/cgit/gettext.git/tree/gettext-runtime/intl) has essentially one file per function, that might still be feasible without modifying the sources, by selectively including files instead. It looks as though substituting [`localename.c`](http://git.savannah.gnu.org/cgit/gettext.git/tree/gettext-runtime/intl/localename.c) might be enough to use a different locale. Perhaps there even is some machinery in there to set the locale per thread; I haven't really read all of that file yet.

balthisar:
> The functionality for determining the appropriate plural form is a single C expression which is part of the PO file itself. So as long as you provide the flexibility of choosing among a varying number of plural forms, picking the right one would be fairly easy.

Oh, that will be easy for the translator, certainly, but then Tidy has to implement a C expression parser to figure out which of the strings to pick. I'm looking through `msgmerge.c` to figure out how to use `plural-exp.c` from the gettext package to see if there's a simple way to re-use that code without reinventing the wheel.

The only other thing that could present an issue is the portability of POSIX format string reordering.

> By the way, properly using gnulib should be enough to provide argument reordering support on pretty much any platform.

I'm looking to pick through the code and re-use a (s)printf implementation directly rather than count on submodules. Unfortunately the code is littered with so many configure-time macros I'm having trouble grokking out what specifically is providing the sprintf. It looks like maybe `printf.c` or `vasnprintf.c`, but again all of the configuration macros are making it a bit unclear.

Currently Tidy has dependencies on only the standard C Lib and we hope to keep it that way. If it means bringing in LGPL source here and there I think the rest of the team can support that, but introducing a huge dependency is a non-starter for the time being.

We haven't really conducted an inventory on what we consider a "supported platform." Other than Classic Mac OS (long deprecated) and Windows, a cursory glance at `platforms.h` tends to make me think that the only non-POSIX platform is Windows, and _certain_ versions of Windows do seem to have sprintf-like functions that support positional specifiers (I just don't know how far back that goes).

@geoffmcl, MSDN only goes back as far as MSVC 2005... does the sprintf family on Windows support positional parameters universally as far as you know? It looks like every other Tidy platform is safe in this regard.

Really, the only two challenges to offering gettext-like functionality (using tokens instead of strings) are:
- Parsing the C expression within Tidy
- Implementing POSIX-style positional parameters in sprintf.

gagern:
> then Tidy has to implement a C expression parser to figure out which of the strings to pick.

I don't see why. You are creating header files, which already may contain C code. So how about this in `language.h`:

```c
typedef struct {
  int (*pluralForm)(int n),
  languageDictionary messages
} languageDefinition;
```

Then in e.g. `language_en.h` you could write:

```c
static int pluralForm_en(int n) {
  return n != 1;
}

static languageDefinition language_en = { pluralForm_en, {
  { TIDY_LANGUAGE, "en" },
  …
  { TIDY_MESSAGE_TYPE_LAST, NULL }
}};
```

The `n != 1` comes directly from the PO file header. The same pattern could be used for other languages as well.

If, for some reason, you feel that you have to parse these strings instead of copying their code, then I suggest you look at [this parser](http://git.savannah.gnu.org/cgit/gettext.git/tree/gettext-runtime/intl/plural.y) which apparently is part of the gettext runtime library. But I can't think of a reasony why you'd have to use this.

> Unfortunately the code is littered with so many configure-time macros I'm having trouble grokking out what specifically is providing the sprintf.

Looking at the [`sprintf-posix` module](https://www.gnu.org/software/gnulib/MODULES.html#module=sprintf-posix), [`sprintf.c`](http://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/sprintf.c) is the file for this. It calls [`vasnprintf`](http://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/vasnprintf.c) which indeed appears to contain the core logic here. But I concur: that file contains so many case distinctions that maintaining them manually may become really painful.

> a cursory glance at `platforms.h` tends to make me think that the only non-POSIX platform is Windows

In that case, you could probably set up some toy project with gnulib, make sure that builds on Windows, and use its preprocessor output to obtain an implementation of `sprintf` without having to resolve all the macros manually. Unless you can get that functionality out of Windows itself.

> certain versions of Windows do seem to have sprintf-like functions that support positional specifiers

Just for reference: the relevant MSDN article here appears to be [printf_p Positional Parameters](https://msdn.microsoft.com/en-us/library/bt7tawza.aspx), right?

gagern:
One other thing: what about character encoding?

The header files currently seem to use UTF-8 encoded string literals. Perhaps the PO-to-H step should replace these by hexadecimal escape sequences, so that you don't depend on the C compiler accepting UTF-8 literals. [This reference](http://en.cppreference.com/w/c/language/translation_phases#Phase_5) suggests that there is indeed a translation between source character set and execution character set, so assuming that arbitrary bytes in string literals will get passed on unmodified may be incorrect, as far as I can tell. With escape sequences, we'd have full control over the binary representation. (Note that [C11 has support for UTF-8 string literals](http://en.cppreference.com/w/c/language/string_literal), but using that would severely restrict portability).

I guess people may expect the setting of [`char-encoding`](http://api.html-tidy.org/tidy/tidylib_api_5.1.25/quick_ref.html#char-encoding) to affect error messages as well. Particularly if these error messages are written to a file. Are all the occurrences of message strings in locations where they will pass through a character re-encoding step before being sent to the output? Would it make sense to add `error-encoding` (or `diagnostics-encoding`) in addition to `input-encoding` and `output-encoding`, so that people could control the character set of diagnostic messages independently?

You might also want to check what kinds of hoops you'll have to jump through in order to make the command line tool `tidy.exe` use Unicode output on Windows. Unless that's already implemented somewhere, or not considered relevant enough for the first implementation of this feature.

Perhaps you should also clarify the semantics around the message filters. As far as I can tell, `mssgFilt` currently receives the localized message, and `mssgFilt2` the localized message format (and arguments). So the message filters don't have access to the unlocalized message, and therefore may have a hard time making reasonable decisions about any given message if they don't know all the translations. I'm not sure about the best solution here, but it might be worthwhile to introduce a `mssgFilt3` which receives some opaque type, and which may call a bunch of functions using which it can extract various things from that opaque type:

* English and the translated message format string
* English and the translated formatted message
* Positional arguments as `va_list`
* Message ID as well if that can be made available easily

balthisar:
Tidy uses UTF internally, exclusively. I have no problems forcing all output to be UTF-8 rather than having to encode to 100 different formats. Output HTML still has legacy use cases for non-UTF-8, otherwise I'd be happy to eliminate everything else from the world.

Ah, yes, mssgFilt2 should definitely _only_ return the native English string, which can be used for localization by library users (which is why I added it a year or two ago). LibTidy users can use the format string as a key for their own applications' gettext or NSLocalizedString or other schemes.

There's not currently a convenient, single opaque error type available, otherwise this would be especially handy to have. Maybe it's something that can be added in the longer term.

Every major compiler supports UTF-8 and most of them default to UTF-8 (probably with the exception of MSVC, but I don't know this), and I'm certain we can ensure that cmake will ensure the headers are used as such during build.

I think this is all reasonably do-able.

balthisar:
# Current status
Given that all current platforms except Windows support POSIX locational parameters, I've managed to get Windows builds to support these, too, with very few code changes. _A review of these specific changes is appreciated_, as I have very little historical experience with Tidy's coding style. The Windows changes are pragmatic and work.

In this branch all of the strings are now using positional parameters. Running the tests in the regression branch on Windows, Linux, and Mac are giving identical error output, so this is a good indication that this is working.

This puts Tidy in an _excellent_ position to work with gettext po files for providing localizations. I'll reiterate these points, though:
- We will _not_ use gettext, but the wide toolset available for gettext po files will be available.
- We will _not_ automate the inclusion of po files as part of the build. Conversion of po files to header files will still have to be performed, and this should be a simple, easy task.
- We/I will provide a Ruby script to translate between po files and header files. If I do this in Bash it will leave Windows without a solution, and at least Ruby can run everywhere if installed.
- Tidy with localization (this branch) now supports plurals in a gettext friendly manner. This means our po files will support plurals properly.
- Tidy with localization now supports positional parameters (e.g., %2u$) on all tested platforms (*nix, Windows), and all strings have been updated to positional format and tested. This means our translators can localize without regard to format string parameter input position.

As of right now the Ruby tool is not at all complete, but expect it to work in a day or two for testing the ability to add localizations. _In the meantime the existing Tidy code really, really needs to intensively tested!_ We pass all tests (except string where we changed strings) in testing branch, and my own testing seems okay.

Special thanks to @gagern for not letting me get lazy ;-)  -- Tidy will have _proper_ localization support!

geoffmcl:
@balthisar just push small K&R like change to compile with MSVC10...

Now over the coming days hope to get into some Windows testing... anything special to look at?

balthisar:
Oh, wow, thanks! I thought I was keeping an eye out for those, and I built on Windows both with Cmake and Visual Studio 2013 to be even more sure.

Time for me to look for a linter, I suppose!

For things to look at I mostly want to be sure that I’ve not broken anything. All of the current regression tests comparing error output pass (except in cases where I changed the string), so that’s a very good sign. I’m pretty sure my memory management is okay and I’m not finding any leaks where I create new strings. My C style is very much out of date due to Objective-C, etc., meaning I bring a lot of baggage and techniques from other languages (this isn’t a good thing in many cases!). Is there anything in the code that I should be embarrassed about??

geoffmcl:
@balthisar have not yet got to actual code testing... that will come...

It seems, maybe, by MSVC 2013, MSVC12, they, MS, have maybe dropped this quite stupid K&R rule that **all** variable declaration **must** be at the top of a context... good on them! about time! gcc has been doing this since the beginning of time! But on the other hand, it is not a **bad** rule... makes developer think about just adding this variable now... it is still **all** on the function stack...

Maybe if you adjusted your Windows VM to use [MSVC10](http://microsoft-visual-cpp-express.soft32.com/) then this would be clear... not sure...

And of course that brings up issues of backward MSVC compatibility in windows.

Traditionally, Tidy has supported back to MSVC6... not really suggesting that, but why not? Really, Tidy code has no special quirks... Tidy is simple, well formed C, not C++! It should be very easy to compile... no special compiler version required... in **all** systems... hopefully with **no**, or **very few** compiler specific MACROS!

It has some already, but what I see are most are related to only my _MSC_VER, no NDEBUG, code... the others should be removed where possible...

And you seem mix this with regression `testing`?

Here you have proposed a completely Ruby way to test... no problem... looks interesting... but you have control over the `expected`. You can, from what I have read, and what you told me, simply update the expected, to `canonical`. Now I know the **`terror`** in this!

Update one wrong file that has other than just `known` string changes, and you hide this fault, this possible regression, for some time... I do not like such an easy option. To change the `expected` should require some additional steps... deliberately make it difficult, not **easy**, to change the `expected`...

Simply, it is difficult to decide, case by case, excluding known string changes, but that is what is required! Only update when **very** sure!!! But update you must, to be able to see any **real** regression problems, without false hits...

That type of **difficult** choice should be carefuly considered, even discussed, if it is **not** just a known string changes...

But as stated, I will get around to testing both... including searching for memory leaks, and the like... 

We are moving forward...

balthisar:
I'll move discussion over to #347...

PR #337, opened 2015-12-14
Link to node-libtidy

gagern:
I've recently started a new [project with node bindings for libtidy](https://github.com/gagern/node-libtidy). Not just a complete tidy run, but lower level interfaces as well. This is a *very* young project, so I can understand if you might not want to add a link to that just now. It would be OK with me to keep this pull request open for a while as a reminder while you wait and see how things evolve.

On the other hand, having various language bindings listed here would in my opinion be a good thing, and from all the contenders I looked at, I think my implementation is already the one offering the most flexible connection to many parts of the library. So choose however you feel appropriate, and let me know if I should take this up again later in case you reject it now.


balthisar:
Hi @gagern, good luck with the project!

gagern:
@balthisar Thanks, for the merge and the good luck wish.

Issue #339, opened 2015-12-31
Character encoding results on different OS (case 445557 as current example)


balthisar:
I've been digging into why this file is different on Mac OS versus Windows and Linux (which I regard as proper output and improper on Mac OS X, where the generated HTML doesn't render properly).

Essentially Tidy is treating ASCII > 127 as "MacRoman", which according to `streamio.h:184`, replaces "_illegal SGML characters regardless of encoding._" The test case specifies **latin1**; I don't think Tidy should be making substitutions regardless of encoding.

A quick fix would be separately defining `DFLT_REPL_CHARENC` in `tidyplatform.h:90-98`, so that at the _very least_ Mac OS X behaves the same as Linux and Windows. I suppose for `MAC_OS_CLASSIC` we could leave in the archaic MacRoman, but we'd still be generating different output (although I'd suggest deprecating `MAC_OS_CLASSIC` altogether as I have no intention of setting up a C development system for testing in an VM).

The reason I'm not issuing an immediate PR for review is because **I'm probably misunderstanding some of the reasons this character substitution is performed in the first place.**

My opinion is summed up like this: _given an identical file on any platform using the same configuration options, the output should be identical_, e.g., if a file is MacRoman on a Windows or Linux system, it should give the same output regardless. If we say a file is **latin1**, we should trust the configuration option.

I've not looked into the other (probable) character encoding issues in the test cases yet, but some of them may be Windows-specific. Note that `streamio.c:467` makes specific checks for isWinChar, too, meaning that we can have the same issues by second-guessing the file encoding format.

As I dig into the other varying test cases, I'll add comments to this bug report.

A discussion about the couple of questions I asked above would be useful, I suggest.

balthisar:
Confirming that case 1632218 is also corrected on Mac OS X by not using MacRoman character substitution.

I'm tentatively thinking that the whole `ReplacementCharEncoding` mechanism isn't necessary. The appropriate substitutions are made if `in->encoding` is set, otherwise we're second-guessing the file format on different systems, even for the very same file.

The only other place I see `ReplacementCharEncoding` used is in `lexer.c:1040`, where again the WIN1252 replacement is used in all cases except for Mac OS.

This means that without understanding the reason for the original distinction, the whole mechanism isn't needed; invalid entities should _always_ be treated as CP1252 unless we want to introduce an entirely new configuration option (which would parallel the input-encoding option, and be largely redundant).

[...]

PR #340, opened 2015-12-31
Fixes Mac OS X encoding issues and harmonizes output across platforms.

balthisar:
Previously Tidy produced different output based on the compilation target, NOT based on
the file encoding and specified options. Every platform was equal except Mac OS. Now unless
the encoding is specifically set to a Mac file type, all encoding assumptions are the same
across platforms.

### DIFF
--- a/include/tidyplatform.h
+++ b/include/tidyplatform.h
@@ -92,9 +92,6 @@ extern "C" {
 #define MAC_OS
 #define FILENAMES_CASE_SENSITIVE 0
 #define strcasecmp strcmp
-#ifndef DFLT_REPL_CHARENC
-#define DFLT_REPL_CHARENC MACROMAN
-#endif
 #endif
 
 /* Convenience defines for BSD like platforms */
@@ -360,13 +357,6 @@ extern "C" {
 
 #ifdef NEEDS_UNISTD_H
 #include <unistd.h>  /* needed for unlink on some Unix systems */
-#endif
-
-/* This can be set at compile time.  Usually Windows,
-** except for Macintosh builds.
-*/
-#ifndef DFLT_REPL_CHARENC
-#define DFLT_REPL_CHARENC WIN1252
 #endif
 
 /* By default, use case-sensitive filename comparison.
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -1037,10 +1037,8 @@ static void ParseEntity( TidyDocImpl* doc, GetTokenMode mode )
                 uint c1 = 0;
                 int replaceMode = DISCARDED_CHAR;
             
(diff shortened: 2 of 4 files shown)


balthisar:
Addresses #339.

include/tidyplatform.h
```diff
@@ -92,9 +92,6 @@ extern "C" {
 #define MAC_OS
 #define FILENAMES_CASE_SENSITIVE 0
 #define strcasecmp strcmp
-#ifndef DFLT_REPL_CHARENC
```
balthisar: This is the actual fix.

include/tidyplatform.h
```diff
@@ -362,13 +359,6 @@ extern "C" {
 #include <unistd.h>  /* needed for unlink on some Unix systems */
 #endif
 
-/* This can be set at compile time.  Usually Windows,
```
balthisar: None of these are needed as no provision anywhere in the code makes for other allowances once we treat Mac OS X the same.

src/lexer.c
```diff
@@ -1037,10 +1037,8 @@ static void ParseEntity( TidyDocImpl* doc, GetTokenMode mode )
                 uint c1 = 0;
                 int replaceMode = DISCARDED_CHAR;
             
-                if ( TY_(ReplacementCharEncoding) == WIN1252 )
```
balthisar: Assuming cp1252 has always been the case, except in the broken Mac implementation.

src/streamio.c
```diff
@@ -476,10 +473,10 @@ uint TY_(ReadChar)( StreamIn *in )
                 in->doc->lexer->columns = in->curcol;
             }
                 
-            if ( isWinChar )
-                c1 = TY_(DecodeWin1252)( c );
-            else if ( isMacChar )
-                c1 = TY_(DecodeMacRoman)( c );
+            if ( isMacChar )
```
balthisar: Still require this logic if the actual input encoding is set to MacRoman, though, otherwise there can be no other assumption than cp1252.

src/streamio.c
```diff
@@ -748,14 +745,7 @@ void TY_(WriteChar)( uint c, StreamOut* out )
 ** Miscellaneous / Helpers
 ****************************/
 
-/* char encoding used when replacing illegal SGML chars,
```
balthisar: Not needed.

src/streamio.h
```diff
@@ -181,12 +181,6 @@ int TY_(GetCharEncodingFromOptName)(ctmbstr charenc);
 #endif
 
 
-/* char encoding used when replacing illegal SGML chars,
```
balthisar: Not needed.

PR #347, opened 2016-01-15
Localization Support added to HTML Tidy

balthisar:
- Languages can now be added to Tidy using standard toolchains.
- Tidy's help output is improved with new options and some reorganization.


balthisar:
I rebased my development branch and issued a PR on this one, and closed the first PR. No sense in adding a bunch of development crud to our master branch.

### Feature Complete

I'm judging the current PR as "feature complete." There may be some bugs and hiccups in the PO<->H conversion process, but that cannot in any way affect the Tidy code.

I've been testing the changes to Tidy's core. In the end they're not all that complex at all and doesn't gut Tidy. I've not found any incompatibilities, leaks, poor output, or other bad behavior at all. Aside from the strings I changed, it's virtually a transparent replacement for current `master` branch.

Unfortunately there are also several changes to tidy.c that I took the opportunity to make that I wasn't able to add to a feature-specific branch. Mostly developer documentation and some tweaks to the output, especially the `-h` output. I've also introduced help for the new language features, and added a couple of new XML output features related to language.

Special thanks to @geoffmcl for making sure I stick to K&R!

### gettext Support

Special thanks to @gagern for the gettext suggestions. While it complicated things just a bit, it really does give Tidy the opportunity to offer world-class language support. Using the Ruby `potool.rb` we can convert between standard PO files and Tidy's language header files.

While gettext's tools won't work directly with Tidy, we can officially support PO files, and this opens us to a large world of potential translators to use their favorite tools. Tidy will support PO files even if they use positional format parameters, and will support the various plural forms of different languages. I never realized, for example, that French used different plural forms (does @geoffmcl know enough French to give us a French translation?)

If anyone wants to get started on a translation, then `tidy.pot` is now officially available in the `localize\translations` directory!

I look forward to merging this branch in a few days unless someone discovers some last-minute showstoppers.

gagern:
I'm looking at [the POT](https://github.com/htacg/tidy-html5/blob/2900d82decd6423f32254af65e9f5ae122328143/localize/translations/tidy.pot).

Not a showstopper for this PR, but it would be nice if the pot could contain `#:` source reference comments, indicating where in the sources the enum constant is being referenced. In some cases this could help clarify the meaning of a short string.

    #. Translator, please ignore following: ##SUPPORT_ACCESSIBILITY_CHECKS##

Here it's unclear to me whether the comment refers to the `##SUPPORT_ACCESSIBILITY_CHECKS##` or to the message as a whole. In the latter case, why are they in here in the first case?

balthisar:
Hi @gagern, I considered adding the reference comment. I may do that yet, as I know it can be helpful. It's just that because we can't use `xgettext` directly it's going to be a manual effort unless I write my own `xgettext` to parse all of Tidy's source code. This means updating the reference information when we make code changes.

As for the developer comment I will try to make it more clear, such as "Translator, please ignore this comment." Because the PO format doesn't allow any metadata to be stored with the strings, I'm using those developer comments to store metadata myself. Hopefully the comments won't frustrate the translators. The particular metadata in your example categories that particular string as being part of a #if SUPPORT_ACCESSIBILITY_CHECKS ... #endif block, which is required information when building the H from the PO.

Now that I think about it, I can probably change this. I have to parse the built-in `language_en.h` anyway in order to translate the PO to H, so I can just get the if-group assignments then without cluttering the PO/POT. You keep making me rethink things and making them better.

I'll keep the PR as is, and implement the enhancement after merging. I've been ignoring too many of the other issues lately.

geoffmcl:
@balthisar ok, thanks for flattening the development...

First it compiles fine in WIN32 with MSVC10, both as 32-bits, and 64-bits... 

But it failed to compile in MinGW32 - some problem with -
```
libtidys.a(win_vsnprintf.c.obj):win_vsnprintf.c:(.text+0x1a): undefined reference to `_vscprintf_p'
libtidys.a(win_vsnprintf.c.obj):win_vsnprintf.c:(.text+0x5c): undefined reference to `_vsprintf_p'
```

Will need to look deaper into this... maybe needs some `#ifdef __MINGW32__` in that source, or maybe `if (MINGW)` in CMakeLists.txt... not sure... still trying to get with this MinGW build, which uses its own `gcc 4.8.1` WIN32 port, and lots of other tools... I would like at some stage to offer MinGW 32 and 64 builds in our binaries, as alternative to the MSVC10 files now offered...

But onto testing the WIN32 64-bit EXE... I tried a few known tests, and at least nothing crashed... I did compare the HTML output with `testbase`, since this HTML output should **not** change... and in the limited tests done with tidy exe 5.1.33localizing, everything appeared ok... as it should - localization should no effect html output...

Of course comparing the error output file shows up all the good changes made in the text... but this is a separate question... and I have yet to also get into the new rudy testing you have offerered...

Over the coming days I will try to do a fuller testing...

Concerning the tranlation of the POT file, wow, I am new to this... Yes, I know perhaps enough french to put the messages into say google translate, and apply some corrections to the results, and pass the results to a french friend to finalise... together we could get there...

But what is this PO toolchain? Favorite PO editor? I do not seem to have these in WIN32. What are they in linux? But the README.md does seem to say it can be done manually, if you are careful... 

Do I understand it correctly that for every `msgid` (english), there must be a `msgstr` in the language of choice, if there is a translation...

Is that it? If it is, then I could develop my own toolchain to extract each `msgid`, work on a translation, with my friend, and put it back together... and then use the poconvert.rb to generate the .h file... But maybe there is more to understand?

Concerning running ruby... I have Ruby installed, and can easily setup (modify the PATH) to use it, but in each new case it exits with warning to run `bundle install`... but when I do that it also alters the Gemfile.lock... but then I can run `ruby`... it is always confusing to have an altered file, that git notes, that probably should **not** be pushed to the repo...

But then the readme says to run [potool.rb](https://github.com/htacg/tidy-html5/tree/localize_flat/localize#potoolrb), but I can not find that tool... I seem missing something... confused a little...

balthisar:
I'll try to look into MinGW32 -- it doesn't define _WIN32? Or maybe it does, but doesn't export MS' `_vscprintf_p`? Maybe I need to capture MSVC instead of MinGW32 -- I don't know enough about MinGW32 yet.
* * *
For the POT, apparently it's a "standard" format for Unix translations (really gettext). A lot of the tools I see support POT and `.strings` (a Mac OS X and iOS format). It seems like a reasonable format, and as it's text based, I guess it's easier than editing the header files directly for people that offer translations. There also appear to be a _lot_ of online services that offer support for the format.

The "tool chain" is gettext specific, something that we're avoiding. The poconvert.rb kind of attempts to replicate some of it specific to Tidy, hopefully in a way that's familiar to gettext users. This is untested because I'm not a gettext user. In any case, the POT file should be familiar to translators, and they use this to generate a PO file that is the actual translation. "Favorite editor" I suppose is up to each translator. There are online tools (including Google) that use the format and provide machine translations. I've been playing around with something called "POEdit" that makes it very easy to translate.

Similar to you, I can speak and write Mexican Spanish quite fluently, but I'm a non-native speaker. I think I'm good enough to edit the machine translations and deliver an acceptable Spanish localization, then await the bug reports or PR's for corrections.

Your understanding is correct about the `msgid` and the `msgstr`. The only thing to note for French, anyway, is the plurals. The `msgid` will also have an `msgid_plural` in English, but the translation, instead of `msgstr` will have an array-type format: `msgstr[0]`, etc., denoting the "case." The case is selected by the language's Plural-Forms, which for French is different than English: `nplurals=2; plural=n>1` means there are two forms (subscripts 0 and 1), and the plural will be applied if the case > 1 (e.g., in French, zero is singular vs plural in English). In the end, for French, it doesn't really matter: if you use [0] singular, [1] plural, Tidy will automatically choose the correct one if the PO (and header) includes the C pieces above (check out the sample langugage_xx.h header files to see how this is achieved).

Actually, French is easy. Some languages, it seems, have up to six plural forms!!!!

See also https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html#Plural-forms

* * *

For the poconvert.rb (oops, docs are wrong; I must have changed the name at some point), sorry about the Ruby. I know Bash, but haven't worked with DOS scripts in ages and ages and ages and ages, and I really want to provide something that works everywhere (assuming one installs Ruby). It's just a scripting language like Perl, and has a dependency manager like Perl.

I think your issue is that some of the dependencies aren't being met. The file `Gemfile` is the manifest for the non-built-in dependencies. The `Gemfile.lock` I should add to .gitignore. The Gemfile makes it simple to ensure dependencies are installed, but it's not required.

Hard way: install dependencies manually with `gem install some_dependency`
Easy way: `bundle install`
Of course, `bundle` is non-built-in, so you can `gem install bundler`, then `bundle install` in the `localize` directory. That should take care of the dependencies issue.

In Windows cmd it's necessary to run Ruby scripts using `ruby poconvert.rb`. If you're using bash provided by git or some other bash, then the standard `./poconvert.rb` will be fine.

On a side note, Ruby is pretty nice and very easy to get used to. I've only dabbled in it before, and recently only became familiar with it due to trying to work on the testing branch and as a conversion tool for this branch. I guess long time Rubyists might have a lot of criticism, but it seems to get the job done.

geoffmcl:
@balthisar, as you can see moved the MinGW discussion to #349

Thanks for expanding on the `msgid` and `msgstr`, and plurals... I think I understand better...

I was about to install a Windows POEdit, but then saw the source is available, but unfortunately it is not a cmake project, and the poedit.sln MSVC solution files given failed to build it, even after also installing all the submodules... a nearly 2 GB of sources... 75 thousand files... but there was one or more failed - eventually it build 27 of the 33 projects... 6 failed...

I will try to continue with this, but see it is mixed with gettext, iconv, i18n, wxWidgets, BerkeleyDB, boost, expat, zlib, and others, so this will be a **difficult** windows build ;=((

I have tried for **years** to get some of these dependencies compiled in native windows, have some successes, but it is tough going... cetainly not possible for the average user...

Maybe I will try their Windows binary installer when I feel **braver**! Usually the installer will also contain all the dependency libraries/DLLs... But yet another editor... why not!

Just a small note concerning using `ruby poconvert.rb`. It is possible to associate a `.rb` extension to use the ruby.exe to load, run it, but it is a tedious process if the installer does not handle putting that association in the windows registry, and `PATHEXT` environment variable... 

Like for example with python, perl, etc. If I have a `which.py` script, then for sure I can run that as `python which.py args`, but because the installer did all the registry/environment setup then I can also run it as `which.py args`, or even `which args`... so such a setup is possible, just not always done, or desired...

The undesirable part of these associations is that if say you already have a `which.exe` app in the PATH, then you can be confused with what exactly is being run, when, where... 

So no real problem about using ruby, in general... but would appeciate you putting the changed `Gemfile.lock` file in .gitignore if we can do without it being there... just to avoid a git modified notice after I run `bundle install`...

Ok I can understand the name change from `potool.rb` to `poconvert.rb`... no problem... and can see this utility converts between `po` and `h`... again no problem when I get there... 

But to start say a french translation I assume I just copy either `tidy.pot` to say `language_fr.po` and set to work adding french `msgstr` strings... or maybe copy `language_en_gb.po` to `language_fr.po`, or is that the same thing? This seems not exactly clear in the current `README.md`!

balthisar:
I'll update the .gitignore today.

> But to start say a french translation I assume I just copy either `tidy.pot` to say `language_fr.po` and set to work adding french msgstr strings... or maybe copy `language_en_gb.po` to `language_fr.po`, or is that the same thing? This seems not exactly clear in the current `README.md`!

I'll update the readme, too to try to make it more clear. Given that I didn't know a thing about POT files a couple of weeks ago, I shouldn't expect potential translators to know a thing about them, either!

In the case of French (because it shares the same number of plurals as English), then all of those actions amount to the same thing. You'll have to set the correct plural form manually in the "header" section of your PO, and set the language code:

~~~
"Language: fr\n"
"Plural-Forms: nplurals=2; plural=n > 1;\n"
~~~

Or you can:
`./poconvert.rb msginit --locale fr`
...and that will produce a nice language_fr.po file for your with the correct header information!

balthisar:
@geoff, which specific MinGW are you using? I see MinGW (which is 32 bit only), MinGW-32 (a different project, also 32-bit only), and MinGW-64.

Is the idea to target _any_ distribution of MinGW, or one in particular? Looking at bug reports, stackoverflow articles, etc., tends to make me think they're not all really in sync with each other.

I still don't have a working implementation to test, but I've run into this: `__mingw_printf()`; Assuming a vsn exists, it should be a simple patch. But now with the different versions of MinGW I wonder if they all define the same `__MINGW32__` macro (why would the 64 bit versions report this?)?

It may be easier to convince the whole world to speak English (negating the need for localization) than to implement a standard POSIX printf on Windows!

gagern:
Some notes while translating to German.

* Running script on Gentoo was a pain, since Ruby 2.3 doesn't have bundler available, and most gems are already installed but for other versions of Ruby. Running the script with Ruby 2.0 worked, though.
* Closely related strings, like `STRING_DISCARDED` and `STRING_REPLACED` are not grouped together in the PO file.
* The `were` in `STRING_ERROR_COUNT` works for German, but I imagine French or others might have problems there if there is just one error and zero warnings. Not sure how do plural forms here. Translators may have to fall back to a message which is not a complete sentence.
* If there are formatting characters inside a text which is expected to be wrapped to 78 characters, it would be nice to know what replacements to expect for these.

And her a list of messages which I could not translate without looking up context first. Perhaps you can add comments to these.

* `STRING_DISCARDING`. What exactly gets discarded? Which of the related words `skipping`, `removing`, `deleting`, `ignoring` would match? Come to think of it, the same holds for the formatted message: you don't know what exactly it applies to.
* `STRING_HELLO_ACCESS`. I wonder why the version number is included in this constant here, and not an argument.
* `STRING_HTML_PROPRIETARY`. This gets pasted into `STRING_CONTENT_LOOKS` which is unclear. Important to get the grammar right! The English term looks trange here, too: why `looks like HTML Proprietary` and not `looks like proprietary HTML`? If `HTML Proprietary` is intended as a fixed term, should it be translated at all?
* `STRING_SPECIFIED`. This gets pasted into `TEXT_VENDOR_CHARS` or `TEXT_SGML_CHARS`. On the one hand it would make sense to make the term “character encoding” part of the string, since for German it would be better to translate “specified character encoding” as “angegebene Zeichenkodierung”, but “MacRoman character encoding” as “MacRoman-Zeichenkodierung” (with a hyphen) or “Zeichenkodierung MacRoman” (with a different word order). On the other hand, the two texts use the term differently, leading to different cases in the translation: “Sie sollten die angegebene Zeichenkodierung vermeiden” in `TEXT_VENDOR_CHARS` but “… mit dem gleichen Byte-Wert in der angegebenen Zeichenkodierung” in `TEXT_SGML_CHARS`. I don't have a clean solution here.

I'm running out of time now, so I'm posting my notes so far even though I'm far from done with the translation.

One more note: the following comment at the end of the PO file might be useful for Emacs users, since it establishes the 78 column rule in the message edit window:

```
# Local Variables:
# mode: po
# eval: (add-hook 'po-subedit-mode-hook '(lambda () (setq fill-column 78)))
# End:
```

geoffmcl:
@gagern wow, thanks for sharing your German translation notes...

That enthused me to begin a French translation... So I installed the binary POEdit in Windows, and kicked off...

At first it looked like a breeze ;=)) 

Click on the 'Source Text' and POEdit provided one or more French translations, rated... Selecting, and sometimes modifiying slightly, and it was done, and I thought I would be finished in no time at all... 

But after about 10/20 or so translations POEdit started to **`nag`** about upgrading to the PRO version to 'remove the limitation'... Bah!, I am not ready to `buy` it yet, even at just 20 bucks, or so... so had to bring up google translate and get into copy and paste... a little **more** work but still very workable... got exhausted at about 10% done, but will come back later...

Some comments - 

I assume I must **not** alter the 'Source Text', the English, although POEdit lets you do that... My single purpose is to give a `msgstr` in French language_fr.po... Or else I would have to push my new english changed `msgid` back upstream to say tidy.pot. Is this the right idea? But in thinking about a translation, did see some better english phraseology...

And POEdit also output a binary language_fr.mo, which I have still to explore...

Have not yet got into running the poconvert.rb script, but assume it can only work properly if the original english `msgid` base is not change... but maybe it does not care? And works on the msgctxt sort of `identifier`. Please advise.

And like you say if we are to try to stay in the 78 character column width, in certain cases it would be good to know the length of the `%1$s` possible options for that item...

And finally, isn't 78 a bit narrow for today's world? Yes, back in the small screen days, not **so** long ago, 78 was probably a good choice, but now I could see that being say 100, or even 110... I now never work in a console of less than about 120 characters wide... in linux and windows...

But maybe poconvert.rb could do that? Adding a column width option, and it relines the text... remove the trailing `\n`, and joins following lines... but again how would it know the approximate size of the potential substitutions? 

Also started a crude perl poconvert.pl, since I know and use perl **all the time** ;=))

Also sort of like you, had to sometimes understand more of the context where the string is used, to provide a more meaningful translation... Sort of wanted to be able to run tidy on say a file STRING_DISCARDING.html, and tidy outputs an example of that warning/error... but I suppose that would require quite some work to set that up... many files... will try to experiment a little...

And concerning MinGW #349, so far found **no** cases, but only at 10%, where I would need to mess with the order of parameters... so at this time feel they could be dropped, and left as a future enhancement, if the need arises... What about in German? Have you needed to change any parameter order?

Anyway, just wanted to report a good start... seems all workable... thanks...

gagern:
> I assume I must **not** alter the 'Source Text', the English […]. Is this the right idea?

I'd agree. I'm surprised POEdit lets you do that; Emacs will by default prevent you from doing that. I must confess that with `tidy.pot`, `language_en.h` and `language_en.po` I'm not sure which of them is generated from which. I also just noticed that the paths are a bit confusing: the repo contains `src/language_en.h` and `localize/translations/language_en_gb.po` while the steps above (running `poconvert.rb msginit`) caused the creation of `localize/language_de.po`.

> POEdit also output a binary language_fr.mo

That would be relevant if tidy were using the gettext library, which it is not. Just ignore it.

> Have not yet got into running the poconvert.rb script, but assume it can only work properly if the original english `msgid` base is not change... but maybe it does not care? And works on the msgctxt sort of `identifier`. Please advise.

I'm interested in this point myself. Canonical gettext requires an exact match to apply a translation. If you change one character, you invalidate all existing translations. Which makes sense since that character you changed might have changed the meaning. Translators can run `msgmerge` to combine a new template with an existing set of translations. Lacking a perfect match, a fuzzy match can be added which will be marked using the `fuzzy` flag. Usually fuzzy translations are omitted from the installed `*.mo` files (unles overridden with the `--use-fuzzy` argument to `msgfmt`). So the idea is that translators check these fuzzy translations and remove the flag once they are satisfied.

To mimic this, `poconvert.rb` should ensure a perfect match, and should ignore translations marked as fuzzy. I haven't checked whether it does so already. In either case this kind of workflow should be documented.

The documentation probably should also advise people on how to test their newly added language. Which in my understanding involves a suitable `poconvert.rb msgfmt` invocation, combined with two additional lines in `language.c` (one for the `#include` and the other for the `languages` array element). Right?

> But now with the different versions of MinGW I wonder if they all define the same `__MINGW32__` macro (why would the 64 bit versions report this?)

Because it's not a 16-bit version of Windows (like Windows 3.11) but a 32-bit capable version of Windows which comes with a different API. I'd assume you can rely on it, but I'm no expert. Looking at the source, I see that [in this 2007 gcc commit](http://repo.or.cz/w/official-gcc.git/commit/0bbe9142cafba534f128e88ff39cdde5f2d2e18f), `__MINGW64__` was introduced *in addition to* the existing `__MINGW32__`. The latter is [from 1997](http://repo.or.cz/w/official-gcc.git/commit/b421d30de8a1093f4a0607af039e3fae282c1159). So I'd say that's a pretty reliable macro. I'm less certain about the `__mingw_printf` symbol, which presumably comes from the MinGW runtime library.

> It may be easier to convince the whole world to speak English (negating the need for localization) than to implement a standard POSIX printf on Windows!

Exactly this kind of trouble is the reason gnulib tries to solve.

balthisar:
From @gagern:

> Running script on Gentoo was a pain, since Ruby 2.3 doesn't have bundler available, and most gems are already installed but for other versions of Ruby. Running the script with Ruby 2.0 worked, though.

There are only a few dependencies, so it should work if you manually `gem install xxx` in Ruby 2.3.

> Closely related strings, like STRING_DISCARDED and STRING_REPLACED are not grouped together in the PO file.

I'll have a look at that. Most of them are grouped by where they are in the source code, and then sorted alphabetically.

> The were in STRING_ERROR_COUNT works for German, but I imagine French or others might have problems there if there is just one error and zero warnings. Not sure how do plural forms here. Translators may have to fall back to a message which is not a complete sentence.

I will look at changing the way this works in the source, e.g., "Tidy detected x error(s) and y warning(s)." Actually this probably works without changing the source code, but only the English string. The errors(s) and warning(s) have proper plural support.

> If there are formatting characters inside a text which is expected to be wrapped to 78 characters, it would be nice to know what replacements to expect for these.

Ah, yes, I can improve the developer comments with examples.

> And her a list of messages which I could not translate without looking up context first. Perhaps you can add comments to these.

I will do so, with some examples.

> `STRING_HELLO_ACCESS`. I wonder why the version number is included in this constant here, and not an argument.

The accessibility checking feature isn't really a "module" per se, and we're not currently versioning it. I suppose the original contributor(s) might have had the idea of treating the accessibility feature as a modular piece of Tidy, but it never seemed to pan out. I would suggest there's no reason to have a version number at all.

> `STRING_HTML_PROPRIETARY`. This gets pasted into `STRING_CONTENT_LOOKS` which is unclear. Important to get the grammar right! The English term looks trange here, too: why `looks like HTML Proprietary` and not `looks like proprietary HTML`? If `HTML Proprietary` is intended as a fixed term, should it be translated at all?

The origin goes back before me. In English this word order is used to emphasize the adjective, and gives a meaning similar to _proprietary_ HTML (including the emphasis). It also follows the word order pattern of "HTML 2.0," "HTML4," "HTML5", "HTML Proprietary." I'm guessing this is the reason, but not necessarily justifying it. We could change the English string, and I believe that this is a string that should definitely be translated.

> STRING_SPECIFIED. This gets pasted into TEXT_VENDOR_CHARS or TEXT_SGML_CHARS…

Let me see how these are used in the source. If it's logical to add another string, then I can add another string.

> One more note: the following comment at the end of the PO file might be useful for Emacs users, since it establishes the 78 column rule in the message edit window:

It should be a simple matter to have the parser ignore those. However many strings _aren't_ limited to 78 characters; will this have an undesired effect for emacs users?

From @geoffmcl:

> I assume I must not alter the 'Source Text', the English, although POEdit lets you do that... My single purpose is to give a msgstr in French language_fr.po... Or else I would have to push my new english changed msgid back upstream to say tidy.pot. Is this the right idea? But in thinking about a translation, did see some better english phraseology...

Source text doesn't matter, and it's strange the POEdit lets you change it; the Mac version doesn't allow editing that field.

For English I've been holding off on changes due to the regression testing scripts, at least for things that are output as errors and warnings (I did make some changes, now they cause regression test failures, naturally!). The _only_ place to change these is in `language_en.h`, and that's by design. It's akin to changing the source code in a normal gettext application.

You can change the English text in the POT/PO if you like, but it won't take effect anywhere. The only thing that matters is `msgctxt`, which is the key that holds everything together. This avoids all of the "fuzzy" stuff, too, because we can change English strings at will without affecting the PO files translators use (old PO will see old English strings, so what?).

> And POEdit also output a binary language_fr.mo, which I have still to explore...

Yeah, not needed. POEdit lets you turn that auto-generation off.

> Have not yet got into running the poconvert.rb script, but assume it can only work properly if the original english msgid base is not change... but maybe it does not care? And works on the msgctxt sort of identifier. Please advise.

Yup, as above, only need to leave the `msgctxt` alone.

> And finally, isn't 78 a bit narrow for today's world? Yes, back in the small screen days, not so long ago, 78 was probably a good choice, but now I could see that being say 100, or even 110... I now never work in a console of less than about 120 characters wide... in linux and windows...

I agree! I'm not sure if the current strings were designed for old consoles (back in the 640x480 days!), or if the original writers simply wanted to avoid going past 80 characters in the their source code editors. I'm comfortable with longer strings.

> But maybe poconvert.rb could do that? Adding a column width option, and it relines the text... remove the trailing \n, and joins following lines... but again how would it know the approximate size of the potential substitutions?

Actually given that it's console output, the *right* thing to do would be to remove all of the newlines completely, and let the console application handle line wrapping. The console application already does some of that (see print2columns, etc.), but it also limits the output to about 78 columns, too.

My pragmatic suggestion is to leave things as they are now, and I can work on that as a future enhancement, i.e., sniff out the console size and handle word-wrapping and the column routines that way, and modify the POTs to eliminate the hard-coded newlines.

> And concerning MinGW #349, so far found no cases, but only at 10%, where I would need to mess with the order of parameters... so at this time feel they could be dropped, and left as a future enhancement, if the need arises... What about in German? Have you needed to change any parameter order?

Being pragmatic, I'm starting to agree with this. If I find a promising solution today, I will try to implement it, otherwise leave it as a future enhancement.

@gagern again:

> I must confess that with tidy.pot, language_en.h and language_en.po I'm not sure which of them is generated from which. I also just noticed that the paths are a bit confusing: the repo contains src/language_en.h and localize/translations/language_en_gb.po while the steps above (running poconvert.rb msginit) caused the creation of localize/language_de.po.

There should be no `language_en.po`! There's a `language_en_gb.po` for UK localization, though.

The paths are hard-coded that way right now so that we don't screw up and overwrite files, so generate files always end up in the CWD, and existing files are renamed with a number. If we're satisfied with the resulting file, we'll have to manually move them into the correct directory. It's trivial to change the script, but for now I'm excercising an overabundance of caution!

I will update the documentation with workflows and descriptions. As per the normal gettext workflow, English the default, authoritative, canonical language as it exists in the source code, and that's the file `language_en.h`. _Everything_ is generated based on this file. When PO or POTs are generated by poconvert, they all reference this file (although if you prefer to translate from a different language, you can generate PO files from other header files -- see poconvert.rb --help).

Some other things to keep in mind:

- Instead of generating binary MO files, poconvert generates `language_xx.h` files. 
- This is why changing the `msgid` in PO files doesn't matter; they're not used when building the language header.
- This allows generating PO with `msgid` in a different language, e.g., to translate from `es` to `es_mx` in an easier fashion. Normal gettext frowns on this, but `msgid` is disposable so it seems like a handy feature.

> I'm interested in this point myself. Canonical gettext requires an exact match to apply a translation. If you change one character, you invalidate all existing translations. Which makes sense since that character you changed might have changed the meaning. Translators can run msgmerge to combine a new template with an existing set of translations. Lacking a perfect match, a fuzzy match can be added which will be marked using the fuzzy flag. Usually fuzzy translations are omitted from the installed *.mo files (unles overridden with the --use-fuzzy argument to msgfmt). So the idea is that translators check these fuzzy translations and remove the flag once they are satisfied.

Fuzzy translations aren't currently supported. I think I can support `#| msgid previous-untranslated-string` as flag that the English string changed in the event that a meaning changed. However our real key isn't `msgid` but `msgctxt`, which is inviolable and should not change. This means that the purpose of strings won't change, and so if we update the English string in source, it really shouldn't have an effect on the translation. There should be no cases of us changing the meaning of string `STRING_MISSING_MALFORMED`, for example.

> The documentation probably should also advise people on how to test their newly added language. Which in my understanding involves a suitable poconvert.rb msgfmt invocation, combined with two additional lines in language.c (one for the #include and the other for the languages array element). Right?

Yup, and I'm sure that it's in the readme.

gagern:
> Actually given that it's console output, the *right* thing to do would be to remove all of the newlines completely, and let the console application handle line wrapping.

I'm of two minds about this. On the one hand you're perfectly right, that would be the decent thing to do to adjust to arbitrary terminal widths. On the other hand, getting this right might be *very* tricky. You can't simply count bytes and look for the last space no later than column 78. You might have UTF sequences using multiple bytes per codepoint. You might have combining marks so you have multiple codepoints per grapheme. You might have double-width symbols in some ideographic language. You might have a language which doesn't use spaces between words. And you might have to use libraries like curses to correctly talk to the terminal to query its size. So while this would be nice to have, I'd definitely postpone it till the main feature has landed. Personally, I'm using my large screen for *multiple* terminal windows of 80 columns each, so I'm particularly happy with the current constraints.

> However many strings *aren't* limited to 78 characters; will this have an undesired effect for emacs users?

No, (re)wrapping the text is a manual operation (pressing Alt+Q), which can simply be omitted for texts which don't have to be wrapped.

balthisar:
**I did it!** It's not the prettiest solution, but we can build with MingW _and_ keep the positional format specifiers!

It's kind of tricky to convince MingGW to perform the build -- I'll figure out how to issue to PR to the MingGW guys so we don't have to do this, but here's the copy-paste from the readme in the new /build/mingw folder.

Building with MinGW
===================

Currently building with MinGW on Windows is a bit complicated, and these
assumptions are made:

1. Using MinGW as distributed from [http://mingw.org](http://mingw.org).
2. Using default install path (C:\MinGW)


Pre-Build Instructions
----------------------

**Copy the file `specs` into `C:\MinGW\lib\gcc\<version>\specs`**

This file was generated with `gcc-dumpspecs > specs`, and then modified to add
`-lmsvcrt_tidy` to the `*libgcc:` linking section.


**Copy the file `libmsvcrt.a` into `C:\MinGW\lib\`** (or someplace else you know
is included in gcc's library search path).

This file is the import library for `msvcrt.dll` that includes the missing
definitions not included in MinGW's own import library. It was generated via
`dlltool -v -d msvcrt.def -l libmsvcrt_tidy.a --dllname %windir%\System32\msvcrt.dll`.
The `msvcrt.def` file is available in the `mingw` directory, too.

You can inspect MS' `msvcrt.dll` with `dumpbin /exports C:\Windows\System32\msvcrt.dll`
in the Visual Studio Command Prompt (not CMD).


Build
-----

`cmake ..\.. -G "MinGW Makefiles"`
`mingw32-make`


Todo
----

Obviously this isn't a fun solution. Some help configuring CMake to include the
import library would be appreciated.

Also I'm not certain we should be distributing the import library as part of the
source. Would it be better to distribute the `msvcrt.def` as part of the source
and build the import library with `dlltool` as part of the build process?

balthisar:
> I'm of two minds about this. On the one hand you're perfectly right, that would be the decent thing to do to adjust to arbitrary terminal widths. On the other hand, getting this right might be very tricky. You can't simply count bytes and look for the last space no later than column 78. You might have UTF sequences using multiple bytes per codepoint. You might have combining marks so you have multiple codepoints per grapheme. You might have double-width symbols in some ideographic language. You might have a language which doesn't use spaces between words.

Yes, I think we should continue to leave this is the translators' hands for the time being. It looks like ICU has the most industry-standard implementation for determining word boundaries, but that's all C++ code, so for now, humans can determine the word boundaries!

gagern:
Yes, ICU would be my bet as well for a library handling this. It can detect grapheme boundaries as well, which should help measure the width of the text. Keep in mind for later reference.

balthisar:
I've had a little bit of Chinese help, so I've added Chinese for testing.
I'm now successfully building with both MSVC13 and MinGW, and Clang on Mac OS X, and gcc on Ubuntu.

There's an issue with Windows building using MSVC. It appears the MSVC compiler doesn't understand the UTF8 source code (affecting Spanish and Chinese), and so garbage is output. MinGW uses gcc and has no issues with the UTF8.

I'm using the following to force UTF8, and it's supposed to work:

~~~
#ifdef _MSC_VER
#pragma execution_character_set("utf-8")
#endif
~~~

International output (including Chinese) works perfectly on Mac OS X and Linux. On Windows compiled with MinGW, output is perfect except for Chinese -- I don't know if Consolas font in my terminal has Chinese characters or not (in Notepad++ the Consolas font has Chinese).

balthisar:
I can't help but laugh at this situation. If I change the files to UTF-16 BE:
- MSVC13 will compile them just fine and provide correct output (except Chinese, see below).
- MinGW now chokes with a cc1.exe out of memory error during compilation!
- Xcode (which uses Clang) and Cmake (Clang) on Mac OS X complain about UTF-16 not being supported.
- Cmake/Make on Ubuntu has more errors than my scroll buffer allows, but seems to be suffering from UTF-16 coding, too.

This is interesting: UTF-8 does **not** require a BOM, but this experiment delivers interesting results. If I set all of the language .h files to UTF-8 with a BOM:
- Windows/MinGW (gcc) works properly.
- Windows/MSVC2013 (msvc) works properly.
- Mac OS X/Code (clang) works properly.
- Mac OS X/make (clang) works properly.
- Ubunutu/make (gcc) works properly.

**Can anyone perform tests with other versions of MSVC?** It seems to be the weakest link here.

* * *

FWIW, the Chinese language in my Windows 10 CMD seems to be writing the correct characters if I C&P into Notepad++. I'll assume (lacking evidence) that Consolas in CMD is lacking the Chinese characters despite Consolas (in Notepad++) having them.

gagern:
I don't think there is anything in any of the C specifications which says anything about a BOM at the beginning of an UTF-8 encoded file. Since you already have the translation between PO and H files, I repeat my suggestion from https://github.com/htacg/tidy-html5/pull/328#issuecomment-168986240: Write the H files using hexadecimal escapes. That way you *know* what bytes you'll have in the executable, no matter the compiler, its configuration or the build environment. Though other approaches *may* work, I doubt you'll find any guarantees, so you'd end up with a positive list of supported compilers, with everything else unsupported unless proven otherwise.

gagern:
In the has-unicode package for node, I recently [read this](https://github.com/iarna/has-unicode/blob/fdd5de141a5564bdb5bc991d951209da40f6a598/index.js#L5-L8):

> Recent Win32 platforms (>XP) CAN support unicode in the console but don't have to, and in non-english locales often use traditional local code pages. There's no way, short of windows system calls or execing the chcp command line program to figure this out.

Do you want to support Windows XP? Do you want to work out the required system calls and perform the code page switch?

Looking around a bit further related to this, I found [this question on Stack Overflow](http://stackoverflow.com/q/9321419/1468366) which also mentions a difference between cmd and Notepad++. One answer there says:

> UTF-8 *should* be achievable with `chcp 65001`, but there are some nasty low-level bugs in the Microsoft C Runtime for this code page, which makes some apps unreliable and some not run at all.
>
> So yeah, I'm sorry, but UTF-8 is a second-class citizen under Windows. Anything that uses the 'ANSI' interfaces for IO, including anything that uses the C standard IO library, including the Command Prompt, won't be able to cope with it properly.
>
> The only reliable way to get Unicode output in Command Prompt is to use the Windows-specific `WriteConsoleW` interface to push Unicode strings directly.

[This answer](http://stackoverflow.com/a/11930722/1468366) to a similar question takes the same stance: UTF-8 and stdio on windows is broken; use [`WriteConsoleW`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms687401.aspx) if you know output is not redirected.

balthisar:
@gagern:
> Write the H files using hexadecimal escapes. That way you know what bytes you'll have in the executable, no matter the compiler, its configuration or the build environment. 

There's a lot to be said for this approach, and the script can be _very_ easily changed to allow this. I may add it as a command line switch simply as a preventive measure. I'm _really_ hoping to avoid this, though, because legible source code is valuable (although the PO file is fairly legible, so there's that).

> I don't think there is anything in any of the C specifications which says anything about a BOM at the beginning of an UTF-8 encoded file.

The UTF specification requires that BOMs be allowed, but also recommends against their use. I also recommend against their use as they serve no purpose. Well, except to tell MSVC not to treat a UTF file as the local code page! I'm mostly upset at MSVC for this. I should generate two sets of headers, and use the ugly escaped ones for MSVC builds only!

 > I found this question on Stack Overflow which also mentions a difference between cmd and Notepad++. 

Oh, thanks for that! I could conditionally use `WriteConsoleW` in the console app if it becomes necessary. Right now Tidy console sets the UTF CP if the user choose a language (this is temporary as it doesn't set the CP if the locale is already set), and I'm getting good UTF8 output with Spanish. I'll see if I can coax some Chinese out of it.

Yet another Windows-only workaround.

gagern:
> legible source code is valuable (although the PO file is fairly legible, so there's that).

I'd not see the headers as source, just as generated files which are committed to repo for convenience. Much like `configure` scripts, `config.h.in` and all that stuff in an autotooled package. You could even drop them from the repo and just include them in release tarballs. The PO files are the actual sources, imho.

> I should generate two sets of headers, and use the ugly escaped ones for MSVC builds only!

That just adds to the confusion, since now you have *three* files defining each language, which might get out of sync in even more bizarre ways.

gagern:
I could conditionally use `WriteConsoleW` in the console app if it becomes necessary.

[This answer](http://stackoverflow.com/a/9051543/1468366) quoting [this blog post](http://www.siao2.com/2008/03/18/8306597.aspx) suggests

```c
_setmode(_fileno(stdout), _O_U16TEXT);
```

followed by UTF-16 output via `std::wcout` or `wprintf`. There should be a matching `_O_U8TEXT`, but that might be just as broken as CP 65001 handling reportedly is. So I'd suggest trying `_O_U8TEXT` first. If it works, we can probably use the same code for both output to console and output to file. If not, we may have to distinguish between output to file and output to console, and use UTF-16 for the latter. The blog referenced above quotes a description for detecting redirected stdout using [`GetStdHandle`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683231.aspx), [`GetFileType`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364960.aspx) and [`GetConsoleMode`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683167.aspx). If that fails as well, `WriteConsoleW` might still be used in the console branch of this.

geoffmcl:
@balthisar did not get time to do more than a pull, and it compiled in MSVC10 (32 and 64 bit) without problems... and my mingw 32-bit...

And I too have problems with say Chinese in the standard `CMD` prompt, despite what I do with `chcp`... have never been able to solve this, while some editors, and browsers do successfully display the Chinese...

But the above comments by you and @gagern give some interesting ideas about using the Windows `W` wide char (unicode) functions... will experiment more...

balthisar:
>> I should generate two sets of headers, and use the ugly escaped ones for MSVC builds only!

> That just adds to the confusion, since now you have three files defining each language, which might get out of sync in even more bizarre ways.

Yes, I know; I was just being sarcastic. Not sarcasm aimed towards *you*, but towards Microsoft.

As @geoff says, I'll continue to look into it.

balthisar:
@geoffmcl, @gagern:
I've been experimenting with using unicode escape sequences instead of forcing the headers to be UTF8+BOM, and these are the results:

- [UCS](https://en.wikipedia.org/wiki/Universal_Coded_Character_Set) are only recognized as a C99 feature, and Tidy isn't a C99 program. Because MSVC never adopted C99 features this isn't going to change.
- Although I can generate UCS sequences for only "required" characters, Clang is complaining if I encode common characters as UCS.
- Using hexadecimal escapes doesn't work cross platform given the same source code. Clang on Mac OS X complains about invalid characters. GCC on Windows and Linux produce different results. It's as if each platform is interpreting the hex as its current locale, and that's a non-starter. There's no reason for Tidy to be converting character sets given a perfectly good UTF8 source.
- Whereas keeping the `language_xx.h` files as UTF-8 with a BOM has a 100% success rate with every platform and compiler that I've tested, even pre-C99, with zero effort.

gagern:
> * Using hexadecimal escapes doesn't work cross platform given the same source code. Clang on Mac OS X complains about invalid characters. GCC on Windows and Linux produce different results. It's as if each platform is interpreting the hex as its current locale, and that's a non-starter. There's no reason for Tidy to be converting character sets given a perfectly good UTF8 source.

I very much doubt it's interpreted according to current locale. It's more likely that compilers have different ideas about how many consecutive hex digits belong to the escape sequence. While things like `\xabc` might be seen as byte `0xAB` followed by character `'c'` on some systems, others might interpret this as `0xABC` which is out of range for `char`. Clang on OS X will give me an `error: hex escape sequence out of range` message for that. If that's what you got, then #358 should take care of this.

balthisar:
Current status: ready to merge.

For the time being we're not going to support positional parameters. If we get translator feedback we can possibly add that to a future to-do. However for the time being this is holding us back.

Although we have full support for hex-coded source files, for the time being we're going to support full UTF8 source instead. If this becomes unmanageable we can always change course in the future.

Also, Chinese output works, and it has been working for quite a while, and no special tricks are necessary on the part of Tidy. One only must set the Windows locale to a language that supports Chinese, and Chinese output works in Console as expected. I've also learned that PowerShell ISE offers full UTF8 support, so at least we can preview East Asian languages without setting our Windows locale. Note that setting Windows locale does **not** change your interface language.

I'll leave this in the current state for a couple of days in case anyone sees an showstoppers, and then I'll rebase this as a single commit against master (probably) before merging. Or maybe not, since @gagern's commits would be lost.

balthisar:
Closing in favor of #359, which I will merge presently.

Issue #346, opened 2016-01-12
<article> proprietary attribute "itemprop"


zanona:
Hey guys, just wondering why `itemprop` can only be applied to `div` and applying it to `article` would result in warnings?

```html
<div itemprop=employee itemscope itemtype=http://schema.org/Person>...</div>
```
`=>  OK`

```html
<article itemprop=employee itemscope itemtype=http://schema.org/Person>...</article>
```
`=> <article> proprietary attribute "itemprop"`

geoffmcl:
@zanona thanks for the report...

Well this simple answer is `TidyAttr_ITEMPROP` **is** included in the table `W3CAttrsFor_DIV[]`, along with some other `ITEMxxxx` attributes, while it is **not** included in `W3CAttrsFor_ARTICLE[]`.

Now is that deliberate? Or just an oversight? We have to check the W3C HTML5 specificaltions...

It would help if you could research and find some links to decide this... after that it is just a copy-paste operation... thanks...

zanona:
Thanks @geoffmcl,

I have done some really light research here and found the following:

From W3C I could spot this in regards the [`itemprop` attribute](http://www.w3.org/TR/microdata/#names:-the-itemprop-attribute)

> Every HTML element may have an itemprop attribute specified, if doing so adds one or more properties to one or more items 

As well as for [`itemscope`](http://www.w3.org/TR/microdata/#attr-itemscope) and, I presume, the remaining `itemxxx` attributes:

> Every HTML element may have an itemscope attribute specified. 

Under [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Attribute_list) `itemprop` is also listed as a [Global attribute](https://developer.mozilla.org/en/HTML/Global_attributes) with the observation:

> Global attributes are attributes common to all HTML elements; they can be used on all elements, though the attributes may have no effect on some elements.

Please let me know if these observations are valid.

Thanks again for your help.

geoffmcl:
@zanona thanks for commencing the research into this... it seems in the 'microdata' section, which is relatively new to tidy in the first place... And I would not look at references other than the W3C... Tidy is a W3C utiltiy... 

See #209, and maybe others... you will find a note in there to the effect "that microdata is NOT an approved W3C Recommendation, but it is supported by the validator so..."

That implementation was for RDFa - https://www.w3.org/TR/html-rdfa/ - it also notes "RDFa attributes are technically legal on ALL elements."...

[...]

Issue #342, opened 2016-01-05
Regression in 5.1.32


serjepatoff:
Hello,

We have found a regression that shows up when processing some htmls. When using version 5.1.32 (commit 99428561642b830267dad334868eacae837f2c39) content is corrupted. In ealier version 5.1.14 (commit 7e69ceb3d1fea1e6cbab7feacbfce7607ec71b49) everything is OK.

Options used are as follows:
```
tidyOptSetBool(tdoc, TidyDropEmptyElems, no);
tidyOptSetInt(tdoc, TidyWrapLen, 0);
tidyOptSetBool(tdoc, TidyXhtmlOut, yes);
tidyOptSetBool(tdoc, TidyMark, no);
tidyOptSetInt(tdoc, TidyDoctypeMode, TidyDoctypeUser);
tidyOptSetBool(tdoc, TidyNumEntities, yes);
tidySetCharEncoding(tdoc, "utf8");
tidyOptSetBool(tdoc, TidyForceOutput, yes);
tidyOptSetBool(tdoc, TidyShowWarnings, no);
tidyOptSetInt(tdoc, TidyShowErrors, 0);
```

There are 2 html content samples in attachment. For each sample there are 
(1) original html, 
(2) html after tidying with 5.1.14, 
(3) html after tidying with 5.1.32, 
(4) screenshot of content after tidying with 5.1.14 in iOS UIWebView, 
(5) screenshot of content after tidying with 5.1.32 in iOS UIWebView.

Here is tree listing of files in attachment, hope that naming would be self-explanatory to you. 

```
.
├── html
│   ├── Content1_original
│   ├── Content1_after_5_1_14
│   ├── Content1_after_5_1_32
│   ├── Content2_original
│   ├── Content2_after_5_1_14
│   └── Content2_after_5_1_32
└── png
    ├── Content1_after_5_1_14.png
    ├── Content2_after_5_1_14.png
    ├── Content1_after_5_1_32.png
    └── Content2_after_5_1_32.png
```

[Attachment.zip](https://github.com/htacg/tidy-html5/files/79039/Attachment.zip)

Thank you very much for your attention.

balthisar:
@serjepatoff, I'll track it down when I can. Thanks for the report.

geoffmcl:
@serjepatoff, yes thanks for the report... took me some time to set up a repeatable test using a 5.1.14 exe... compare the outputs... etc...

[...]

Issue #329, opened 2015-11-30
Space missing when opening tag at end of a line


yapper-git:
I use Tidy in a project, but I got a bug.
Below is a minimal input code example.

``` html
<p>This a minimal example for<strong>
TidyTidyTidyTidyTidyTidyTidyTidyTidy</strong> is another nice open source tool :)</p>
```

The problem is that a line ends with `<strong>`, but the white space (here newline) is in the `<strong>\n'.

All browsers displays a space between _for_ and _TidyTidyTidyTidyTidyTidyTidyTidyTidy_ (thanks to the newline).
While if you convert this HTML file with tidy (with default configuration).
Tidy outputs :

``` html
<!DOCTYPE html>
<html>
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 for Linux version 5.1.25">
<title></title>
</head>
<body>
<p>This a minimal example
for<strong>TidyTidyTidyTidyTidyTidyTidyTidyTidy</strong> is another
nice open source tool :)</p>
</body>
</html>
```

The whitespace disappeared...

geoffmcl:
@yapper-git thank you for moving the **bug** here... This is certainly where it may receive attention!

You might want to **update** your tidy output to show this problem still exists in current htacg tidy, current `master` branch, version 5.1.28 plus. Showing a 2009 output does not help the case.

I called this a `feature request` purely because tidy has been doing this **forever**! So it is a quite a fundamental change in the parsing...

The problem is, in the current parsing, in certain circumstances, tidy `hides` newlines, quietly swallowing them, so the caller does not know they even exist. So the fix is deeper than you think.

But that said, does not mean it is not a **bug**. I, for one, agree 100% that all the browser tests I have tried on your sample shows a space between `for` and the first `Tidy...`, and one of the **strong** aims of tidy is to render a **tidied** output that will **look** the same in a browser!

As stated, thanks for the report, and I, or others, will look into this soonest. But please update the output to reflect current tidy, and maybe remove the `.` which does not seem to be in the input... thanks...

yapper-git:
> You might want to update your tidy output to show this problem still exists in current htacg tidy, current master branch, version 5.1.28 plus. Showing a 2009 output does not help the case.

I updated the output above (with 5.1.25 version installed on ArchLinux).

geoffmcl:
@yapper-git thanks for updating the sample... as promised will look at this soonest...

OT: Just out of interest where did you get the tidy 5.1.25 version?

[...]

PR #350, opened 2016-01-19
Enforce Attributes, Address #346, Refactor attrdict.c

balthisar:
When discussing #346 we side-tracked a bit into optimizing `attrdict.c`. This PR addresses the original issue in #346, massively refactors `attrdict.c`, but raises a couple of questions that I need help with!

In the past Tidy kept a record of which attributes apply to which version of HTML, but in fact did nothing at all with this information. The current PR _will_ use that information and emit warnings when an attribute is used for a version of HTML that doesn't allow it.

This is _new_ behavior and naturally it breaks a _lot_ of our regression tests because Tidy is being strict about attributes that are allowed.

Questions:

- Is this the desired default behavior?
- Is it preferable to tie this to a new configuration option? For example, `mismatched-attributes`, default `no` using the old behaviour.
- As currently committed, we're using `drop-proprietary-attributes` to drop the illegal attributes.
  - This default is `no` so I'm not sure how unwanted this behavior is.
  - Alternatively we can leave them there and count on the warning being enough.
  - Or tie their removal to proposed `mismatched-attributes` configuration option.
  - Or add a new `drop-mismatched-attributes`
- As currently committed, we're using the same message as proprietary attributes (although I updated the string a bit). The plan is to give it a separate message once `localize` branch is merged (current #DEFINEs are a PITA).
- As currently committed we try to guess the doctype if one isn't provided, otherwise nearly all of the attributes are reported.

On the other hand we can simply _ignore_ the fact that attributes don't match the version of HTML in use as is done currently. This still gives us the benefit of addressing #346 and a massively improved `attrdict.c`, but still causes a regression in the text-based suite. This regression is due to `onload` being a universal attribute, and since it's now properly present in HTML5, it's not being called as a warning in the HTML4 test case.

@geoffmcl (and anyone else), would you help provide a sanity check here?


balthisar:
I think this is a strategy:
- Merge in `localize`, can continue to fix the MinGW issues in `master`.
- Also merge in this branch without the last two commits. It only causes a single regression, but only in emitted warnings and not HTML or exit status.

This will enable development of the configuration options and additional warning output with minimum of hassle due to all of the strings, option descriptions, etc.

_Much_ easier to add strings and not work with DEFINEs in `localize`! :-)

geoffmcl:
@balthisar sorry for the delay in my reply on this one... sometimes I forget to also check in PR's ;=() and the fact that it is partially in #346 and here did not help, but that is no real problem...

First a word about the `doctype` option, which I saw used in your example in #346, and the so called html4-- and html5++ modes, which effect what input tidy expects, thus reports. I have still to push it, but see the small [patch](https://github.com/htacg/tidy-html5/issues/342#issuecomment-170267062) to fix #342. The reason for that is I did it in my `issue-329` branch, which does fix #329, but introduces some other not so good side-effects, so can not yet merge it...

It drags up a newline character from the lower io layer, to be able to correctly get `<p>a <em>b></p>`, from `<p>a<em>\nb></p>` but presently causes lots of other not so desireable changes... but back to doctype...

If the doucment has a DOCTYPE, and if that is not html5, then tidy reverts to html4 mode - see [AdjustTags](https://github.com/htacg/tidy-html5/blob/master/src/lexer.c#L2832). That works fine...

But if there was no DOCTYPE, then tidy would stay in html5 mode, hence the above patch to allow the `doctype` option to also switch input modes.

But this is **incomplete**. For example what about when both are present - a doctype on the document **and** a doctype config option. In the end, I think a user config option of the doctype should over-rule any physical doctype, but was not completley sure about that... so did nothing...

Now I say all this because in your #346 [sample](https://github.com/htacg/tidy-html5/issues/346#issuecomment-172448333) you do have a DOCTYPE, so tidy would slide to html4 mode, but you also added a `--doctype transitional` config, which would do nothing... In your case it would not matter since they are the same... but it could matter...

As you point out while tidy has this big attribute table, with each attribute bit marked in where it is allowed, this is not really used. Well there is some version reduction/factoring, but that was only really used in an info output... Because your sample has attributes which were not allowed in html4, the only note by tidy is `Info: Document content looks like HTML Proprietary`... 

Tidy is not exactly clear why it is 'proprietory', except interally all html4 bits were removed, so it was left with sort of 'nothing fits'!

It seems your questions are exactly about this. I agree with you, that, like the W3C validator, tidy should output warnings -

```
Doctype:	HTML 4.01 Transitional
Line 6, Column 14: there is no attribute "ACCESSKEY"
Line 6, Column 47: there is no attribute "CONTEXTMENU"
```

Now I have not looked deeper, but I expected `--drop-proprietary-attributes yes` to remove those unexpected attributes. But it seems to have done nothing! The result is that tidy writes a tidied html4 document with them still there. That is it is outputting a known **invalid** document!

So I think that has to be changed... with or without additional options that may further control this...

I have still to checkout your `attrdict_346` branch, and look at what you have done... but will try to get to that soonest...

But have I answered your questions? In simple terms -

 - tidy **should** warn specifically about attributes not matching doctype
 - given the right options, tidy should try to output a fully **valid** docutment to match the doctype
 
As mentioned elsewhere, I feel a tidied document should have a `99%` plus chance of passing validation.
 
As usual, out of time tonight, but will bounce back with more soon...

balthisar:
@geoffmcl 
> But have I answered your questions?

Yes, I think it's clear how to move forward. This will cause a _lot_ of regressions, but that may be perfectly acceptable if it's the right thing for Tidy to do.

On the other hand we could "leave well enough alone" and introduce new options.

Right now the state of the current branch attempts to delete them based on TidyDropPropAttrs.

balthisar:
@geoffmcl,

This latest push adds a check in `lexer.c:GetTokenFromStream` to additionally check if an element is not used in the current version of HTML. It's added as an `else` so that that if an element is already reported as proprietary, it won't also be be reported as not applicable for this version.

This limited HTML5 test works as expected:

~~~
<!DOCTYPE html>
<html>
<head>
<title></title>
</head>
<body>
<marquee>Hello</marquee>
<font>Hello</font>
</body>
</html>
~~~

I'm worried that I'm still using `ApparentVersion()` to make the check, as I'm still uncertain where this become definitive.

It's also revealed an inconsistency/bug?

In `tidylib.c:tidyDocCleanAndRepair()` we're doing a `CheckHTML5()`, but _only_ if the doctype option is set to `html5`; this doesn't occur in `auto`, even if the input document is HTML5. This gives us different warning output:

`line 9 column 1 - Warning: <marquee> is not approved by W3C` due to the check in `lexer.c` (it's proprietary), but _not_ `line 9 column 1 - Warning: <marquee> element removed from HTML5` unless we set the doctype option.

Now if I run case 1004512 (as discussed in offline email), the actual doctype is strict, but `ApparentVersion()` is transitional, so we don't flag the use of font as we should.

Now I'm wondering if the lexer is even the appropriate place to be making these checks; at the point of lexing we seem to have indeterminate knowledge of the real HTML version. in `tidyDocCleanAndRepair` the `CheckHTML5()` seems to be doing the checks as a post-processing step, and so I'm wondering if we should refactor the proprietary and mismatched-version checks in the same fashion as `CheckHTML5`. I _think_ that by time we reach this point, we know the version of HTML that we're going to treat the input file as.

Any feedback?

balthisar:
Well, the _nice_ thing about having rolled in localize is that it's trivial to play with strings now. Anyway...

Current status (needs more testing, especially the HTML version detection!):

- We're now post-processing the tags to see if they're legal for the current version of HTML. This is new behavior.
- Detecting proprietary HTML is moved into the same post-process function as the above to help avoid duplicates.
- We're still detecting attributes that don't belong to the current HTML version, but also printing out the expected HTML version as part of the warning.
- REMOVED several of the HTML5 replacement warnings, which would be redundant now.

I didn't test on MSVC; I left several modern C comments in the code just to block them out; hopefully MSVC won't choke on them. These would be cleaned up.

**This is a can of worms** like so many things I try to improve, i.e., just fix #346, but I think that _if version detection is correct_, then this is really makes Tidy do the right thing.

balthisar:
As of 8c6c0e4, I like where this is going. I've not reworked the attributes yet (the cause of this whole thing), but the proprietary elements and not-in-this-version is working well with this commit. I'm using this logic as documented:
~~~
 * Check and report HTML tags that are:
 *  - Proprietary
 *  - Not supported in the current version of HTML, defined as the version
 *    of HTML that we are emitting.
 * Proprietary elements are reported as WARNINGS, and version mismatches will
 * be reported as WARNING or ERROR in the following conditions:
 *  - ERROR if the emitted doctype is a strict doctype.
 *  - WARNING if the emitted doctype is a non-strict doctype.
 * The propriety checks are *always* run as they have always been an integral
 * part of Tidy. The version checks are controlled by option `to-be-determined`.
~~~

Currently not using the non-existent `to-be-determined` option.

Just for testing I'm using this html, which seems to let me use all of the doctypes. I've also been alternating between doctype auto and doctype html5 configuration options.

~~~
<!DOCTYPE html>
<!-- <html xmlns="http://www.w3.org/1999/xhtml"> -->
<!-- <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"> -->
<!-- <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN"> -->
<!-- no doctype declaration, too -->
<!-- <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"> -->
<!-- <html xmlns="http://www.w3.org/1999/xhtml"> -->
<!-- <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"> -->
<!-- <html xmlns="http://www.w3.org/1999/xhtml"> -->
<!-- <!DOCTYPE html PUBLIC "-//IETF//DTD HTML 2.0//EN"> -->
<!-- <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN"> -->
<!-- <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML Basic 1.0//EN" "http://www.w3.org/TR/xhtml-basic/xhtml-basic10.dtd"> -->
<html>
<!--
     Depending on whether Tidy is treating this document as proprietary or
     strict, we should generate warnings or errors for the attributes and
     elements that are not allowed.
-->
<head>
<title>Test</title>
</head>
<body>
<marquee fake_attribute="I'm a proprietary attribute">Marquee is proprietary.</marquee>
<font size="+1">Font isn't allowed in strict documents.</font>
<acronym>Acronym is deprecated, too.</acronym>
<p jose="joseph">
This is a valid paragraph but contains a proprietary attribute.
</p>
</body>
</html>
~~~

The lexer's `versionEmitted` seems a bit wonky. I would expect it to tell me what the final output document type is. But between `versionEmitted` and `docType` I _think_ I'm detecting what's actually being output by Tidy -- this is _critical_ to knowing whether an element is permitted or not. The quick check above, though, gives me 100% success for each HTML version.

Now in order to get back to the initial issue, I will see if it's practical to check the attributes in post instead of during `ParseDocument`. What worries me about the check in `ParseDocument` is that we've not really settled into our final output HTML version, so it will be difficult to determine whether an attribute applies within the current document or not.

balthisar:
> I will see if it's practical to check the attributes in post instead of during ParseDocument. 

It was practical, and is in current commit. Next up is to implement an option to control this behavior.

@geoffmcl, as the real Tidy expert, a sanity check is appreciated!

geoffmcl:
@balthisar wow, I am trying to keep up... but with more than a dozen commits to this `attrdict_346` branch in that last few days, plus a proposed more than 6+ doctypes to test, and testing 2+ `--doctype` options, will take time ;=))

As stated off list, this seems to be generally headed in a new good direction for Tidy, which seems all **good**. Now no recent `crashes` which is great! And of course this is futher complicated by lots of `new` message text, your localization, which is also good, but complicates test comparisons, but not really a problem...

In general it seems, feels, much better that the analysis of the nodes collected during the parsing phase, and thus warning/error reporting, has been moved only to the important clean up phase... This facilitates the decision of what doctype is on the document, assume HTML5 if none; what doctype it desired by the user config, if any; and finally what doctype is tidy to finally output... and some previous hard coded decisions are moved into the tables... We are truely now enforcing some previous `faults` that current tidy overlooks...

As stated separately, given the right set of config option, allowing tidy to delete, modify, at will, I hope tidy can output say a `strict` document, that will pass W3C validation... sans fault...

And with all this reducing attrdict.c from its **massive** 1.5MB size, to a more `normal` 320KB file is fabulous... It is still the largest source file, but now only about twice the parser.c, and lexer.c, which are the true guts of Tidy... and the new language_en.h, which alfter all is just text strings...

But we really need that `yet-to-be-decided` option to turn this new behavior `on` and `off`, because that too needs to be tested! As stated separately, we need to be able to turn off this `new` behaviour to test that the old tidy is still there, without any problems, even if we later decide only the new behaviour should totally prevail...

This is a stupendous effort! I just hope your enthusiasm, and available time continues... I will certainly do my best to dig into testing, and give feedback... 

And really invite **all** others to check out this branch, and give it a whirl... a new tidy is born!

balthisar:
@geoffmcl, thanks for having a look.  I've added `strict-tags-attributes` to control the new behavior.

Thanks to the language changes and this change, I definitely want to go back and have a look at the testing suite! Now I'm off to start versioning the testing suite for output-based comparisons. I guess some of the tests will have a pre-localize, post-localize, and then whatever we end up pulling due to these changes.

geoffmcl:
@balthisar thanks for another 6 plus commits, including adding the `strict-tags-attributes` option... this is starting to come together ;=))

So my first test was to -

 - Create a new testbase using **5.1.36** master. This assumes all the changes there were purely text. That still needs verification!
 - Then re-ran the 228 testcases.txt adding `--strict-tags-attributes no` using version **5.1.36attr_dict** attrdict_346 branch.
 
Of course I expect **no change**. Well the simple testing of the tidy exit value yielded just 5 that were different 438954 661606 1183751 1207443 1282835 - just a 2% change. It seems simple that these warning outputs need to be 'protected' by say an `if (cfgBool( doc, TidyStrictTagsAttr ))`. That is only emit them if this is on, and tidy would pass this first stage...

The 'unexpected' warnings output were -

  - BAD_BODY_HTML5  - Warning: Found attribute on body that is obsolete in HTML5. Use CSS
  - BAD_ALIGN_HTML5 - Warning: The align attribute on the `<xxx>` element is obsolete. Use CSS

Now, I have not yet done the full compare of outputs. That may yield **more** cases of difference...

So how do you want to deal with this? I could find the above, add the above `if` and push, but if you get the time, you may find those and other 'new' messages that should **not** be output if the option is off... that is if I have the right idea here...

And as mentioned I am slowly migrating my testing to potentially a separate 'test' repo, and can now see multiple compares depending on versions, **AND** more importantly on the config options used. Many of these config items will now change the output expected, and each would need to be tested for `regression`... not an easy task... and I will continue to use batch processing, since that is what I know best...

As usual this is my short shopping and french day so do not have time to do more now...

balthisar:
Quick observance of the items that failed the exit status:

1183751: this happens because we're now doing `tidyDocCleanAndRepair()` in `tidylib.c:1810`, but having added `auto`; i.e., we're treating `auto` as HTML5 for **TidyDoctypeMode**. I'm a bit dubious about this treatment; what's the difference between `auto` and `html5`, then? In any case, this test result is _correct_ if we're looking at HTML5; but the output is correct transitional; _I think this reveals an additional bug_; we can't `TY_(CheckHTML5)()` based on the doctype mode; I think we have to tie this to `versionEmitted`, too. What are your thoughts?

- 1207443: Same as above.
- 1282835: Same as above.
- 438954: Same as above.
- 661606: Same as above.

If I revert `tidylib.c:1810` to ignore the `auto` mode, we get 100% pass rate on the exit status for all five of the reversions. There are still some markup and errors reversions (the Ruby script is convenient!), but at least the exit status is consistent.

I seem to recall your mentioning that `doctype-mode` set to auto equates to HTML5; but I'm thinking this only applies if no doctype is specified in the document; in these five cases a doctype is in the document, and so 1810 needs to consider not just the TidyDoctypeMode, but our versionEmitted, I think.

It's late and I've been force-fed too much red wine by my in-laws (XinNianKuaiLe, or Happy [Chinese] New Year) to do too much digging now, but at least there's an easy explanation for those five.

It's interesting how every step closer we get to solving the original issue, more and more complicated this becomes!

geoffmcl:
@balthisar did not get a big chance to look at this today... Hope you enjoyed the wine, and family and a **big** Kung Hei Fat Choi to you and all...

> what's the difference between auto and html5?

Well my reading of the [documentation](http://api.html-tidy.org/tidy/quickref_5.1.25.html#doctype) and understanding is that if the user adds html5, that is what tidy should output on the document, and if it is auto, then tidy should choose the best doctype to output after observing what is in the document... constrained versions, depending on what was found... and then the other choices - omit, strict, loose, or user...

Note this is output... so if it is set to html5, then for sure we should do CheckHTML5! However, if is is auto, then if and only if the document either does not have an existing doctype, or has a html5 doctype, only then does auto equal html5, and CheckHTML5 can be done...

These 5 tests 1183751 1207443 1282835 438954 661606 all have a doctype 4 transitional, so I think you are right! If auto, then should not use CheckHTML5! Thus [tidylib.c:1810](https://github.com/htacg/tidy-html5/blob/attrdict_346/src/tidylib.c#L1810) must be modified - **not** be a simple compare...

But what service gives us a clear answer to this? You suggest versionEmitted, but where, how is this set?

I can see it is set in FixDocType, but that is not called until line 1845 - that could be moved up - but FixDocType is some difficult reading!

Some things are clear... `if (xhtmlOut && !htmlOut)`, then SetXHMTDoctype is used... it will set versionEmitted to `ApparentVersion`, an X value, one of VERS_XHTML...  forget them for now...

If we have a document doctype node, and `(mode == auto)` and `(lexer->doctype == VERS_HTML5)` then set HT50, and return Yes. And later will switch case making a guess. Only case mode html5 set HT50, else otherwise, and return Yes.

Can it ever return no? It seems not checked... But it seems it could be called before line 1810... Maybe it is a mistake to be calling CheckHTML5 so early in `tidyDocCleanAndRepair`?

Quite frankly I can not remember the reason why I added it so early in the process... in fact maybe it should be nearly last? Let everyone else fix the document, warn about what they have to, fix, coerce, stylise, etc, only then check HTML5 compatibility, only if the output is to be html5

Then we could have a simple -

```
  if (lexer->versionEmitted == HT50)
  {
    TY_(CheckHTML5)( doc, &doc->root );
  }
  TY_(CheckHTMLTagsVersions)( doc, &doc->root );
  TY_(AttributeChecks)(doc, &doc->root);
```

What do you, others think?

I can agree that in getting close to closing this issue, the more complicated the solutions sometimes seem. But in this case also seems to simplify, and help demystify the code...

Out of time tonight to test this, and tomorrow having some raspberry pie ;=))))

But this feels good... Should remove the problem for those five tests, and maybe some others... maybe you will get a chance to try it before me ;=)) and I can pull a **fixed** tidy...

balthisar:
> What do you, others think?

That seemed to do the trick. I'd suggest that we'd also check for XH50 as well, but I'll dig into it to see what that might be wrong. I seem to recall that sometimes emittedVersion is set to VERS_HTML5 for some reason, which is a bit field for both HT50 & XH50.

balthisar:
Just an FYI, my compiled notes about the remaining test failures, where I'm comparing the error output and the generated HTML, too (my comments added to the test report):

~~~
HTML Tidy Regression Testing Exception Log
==========================================

Test conducted using /Users/jderry/Development/tidy-xcode/tidy-html5/build/cmake/tidy, version 5.1.36attr_dict
Results generated on 2016-02-11T09:56:44+08:00.

Summary:
========

     Number of case files: 228
Total test configurations: 228

        Case files tested: 228
    Configurations tested: 228
   Configurations aborted: 0
    Configurations passed: 198 of 228
    Configurations failed: 30 of 228


About this test suite:
======================

These test files represent that standard regression testing that must be
performed prior to committing changes to Tidy's code. In some circumstances
results are platform specific and these notices will be displayed in the
testing results.



Failed Tests:
=============

Case File            For Configuration     Markup   Errors   Exit Status   
---------            -----------------     ------   -------  -----------   
case-1004051.html    config_default.conf   PASSED   FAILED   PASSED     
case-1015959.html    config_default.conf   PASSED   FAILED   PASSED        
case-1241723.html    case-1241723.conf     PASSED   FAILED   PASSED        
case-1410061c.html   case-1410061c.conf    PASSED   FAILED   PASSED        
case-441568.html     config_default.conf   PASSED   FAILED   PASSED        
  New: Info: Document content looks like HTML 4.01 Transitional
  Old: Info: Document content looks like HTML 3.2

case-1014993.html    config_default.conf   PASSED   FAILED   PASSED    
  Actually this looks like an improvement, but is magic. The expected output
  was line 8 column 5 - Warning: <font> proprietary attribute "NULL", but this
  is confusing because there are no attributes on the <font> tag.
  
case-1053626.html    config_default.conf   PASSED   FAILED   PASSED    
case-1063256.html    config_default.conf   PASSED   FAILED   PASSED   
case-1316307b.html   config_default.conf   PASSED   FAILED   PASSED        
case-1326520.html    config_default.conf   PASSED   FAILED   PASSED        
case-426885.html     config_default.conf   PASSED   FAILED   PASSED        
case-504206.html     config_default.conf   PASSED   FAILED   PASSED        
  This looks okay. The order of the warning outputs is different, but output
  is the same.

case-1055398.html    case-1055398.conf     FAILED   FAILED   PASSED 
  This is an interesting behavioral thing that merits discussion. The <font>
  `onLoad` used to be categorized as proprietary, even though it wasn't, so
  now the output comparison fails because onLoad isn't proprietary; it's simply
  not supposed to be used pre-HTML5.
  More worrisome, though, is we're changing the doctype to transitional, whereas
  the test case maintains the original, document-declared strict doctype.

case-1062511.html    config_default.conf   PASSED   FAILED   PASSED        
  Missing line 3 column 1 - Warning: <marquee> is not approved by W3C
  For some reason it looks like these messages aren't getting out?

case-1098012.html    config_default.conf   PASSED   FAILED   PASSED        
  Missing line 6 column 1 - Warning: <table> lacks "summary" attribute

case-1198501.html    config_default.conf   PASSED   FAILED   PASSED    
  New: Info: Document content looks like HTML 4.01 Strict
  Old: Info: Document content looks like HTML 4.01 Transitional

case-1207443.html    case-1207443.conf     FAILED   PASSED   PASSED        
case-427826.html     case-427826.conf      FAILED   PASSED   PASSED        
  This is interesting; we're changing the doctype to strict!

case-1359292.html    case-1359292.conf     PASSED   FAILED   PASSED        
  Well, this looks like -quiet output, so I'll have to re-run this without
  -quiet. The expected number of warnings was three, versus one in the new
  behaviour.

case-431721.html     case-431721.conf      PASSED   FAILED   PASSED        
  Missing reports for proprietary and not approved by W3C items.

case-431964.html     config_default.conf   PASSED   FAILED   PASSED        
  New: Info: No system identifier in emitted doctype
  I've seen this here and now; need to investigate further.

case-433021.html     config_default.conf   PASSED   FAILED   PASSED        
  New:
    line 6 column 1 - Warning: The align attribute on the <p> element is obsolete. Use CSS
    line 9 column 1 - Warning: The align attribute on the <td> element is obsolete. Use CSS
    line 11 column 1 - Warning: The align attribute on the <td> element is obsolete. Use CSS
    line 13 column 1 - Warning: The align attribute on the <td> element is obsolete. Use CSS
  Actually, this looks like a correct warning. Unexpected, so need to figure out why!

case-433666.html     config_default.conf   PASSED   FAILED   PASSED        
  Added line 4 column 1 - Warning: The align attribute on the <table> element is obsolete. Use CSS

case-433856.html     case-433856.conf      PASSED   FAILED   PASSED        
  Added line 8 column 1 - Warning: Found attribute on body that is obsolete in HTML5. Use CSS
  I've never come across this before. It will be interesting to track down.

case-441508.html     config_default.conf   PASSED   FAILED   PASSED        
  Added line 4 column 1 - Warning: The align attribute on the <table> element is obsolete. Use CSS

case-443678.html     config_default.conf   FAILED   PASSED   PASSED        
  This is strange and makes no sense! The only difference is the case of the hex string!

case-470663.html     case-470663.conf      PASSED   FAILED   PASSED        
  Not getting the proprietary attribute warnings.

case-540571.html     case-540571.conf      PASSED   FAILED   PASSED        
  Missing line 10 column 1 - Warning: <span> proprietary attribute "color"
  This is interesting because we are getting the same warning for 10:23!

case-588061.html     config_default.conf   FAILED   FAILED   PASSED        
  A difference of four errors, and this is affecting generation of CSS replacements.

case-655338.html     config_default.conf   FAILED   FAILED   PASSED        
  Old: Info: Document content looks like XHTML 1.0 Strict
  New: Info: Document content looks like HTML 4.01 Strict
       Info: No system identifier in emitted doctype
  We're also changing the doctype and not using the old HTML attributes
       for an XML file.
~~~

I'm confident that _most_ of these are a result of not doing ConstrainVersion properly -- I'll have to dig back into that. Several of them can be attributed to changing the order of operations, i.e., certain checks are now not in the same order as they were originally.

geoffmcl:
@balthisar thanks for the quick fix ;=))

I think emittedVersion, assuming `lexer->versionEmitted`, should always be just one bit value, and would consider it a sort of **bug** if not - Tidy can emit only **one** version! - but in this case maybe it is also correct that the test be for VERS_HTML5, ie both bits... but have done very little XHTML testing... Does CheckHTML5 correctly deal with this different X doctype? Or should there be a similar CheckXHTML5 function? Questions to be explored...

Wow, thank you for the compiled, and commented, Exception Log... it goes a long way to convincing me that I should give Ruby a try ;=)) Nope, I take that back. Just tried to run `ruby regression_test.rb rtest -r=f:\path\to\tidy.exe` but it always exits with an error... it creates a `results-5.1.36attr_dict` folder, with maybe 3/4 cases done, but then bombs... 

```
D:\temp\tidy-testing\test\benchmarks>ruby regression_test.rb rtest -t=F:\Projects\tidy-testing\build\win64\Release\tidy.exe
.........D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:1454:in `rmdir': Directory not empty @ dir_s_rmdir - C:/Users/user/AppData/Local/Temp/d20160211-5476-1osf6yg (Errno::ENOTEMPTY)
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:1454:in `block in remove_dir1'
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:1468:in `platform_support'
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:1453:in `remove_dir1'
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:1446:in `remove'
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:776:in `block in remove_entry'
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:1502:in `postorder_traverse'
        from D:/Ruby21/lib/ruby/2.1.0/fileutils.rb:774:in `remove_entry'
        from D:/Ruby21/lib/ruby/2.1.0/tmpdir.rb:94:in `mktmpdir'
        from regression_test.rb:698:in `execute'
        from regression_test.rb:1085:in `block in process_case'
        from regression_test.rb:1014:in `each'
        from regression_test.rb:1014:in `process_case'
        from regression_test.rb:1156:in `block in process_all'
        from regression_test.rb:1156:in `each'
        from regression_test.rb:1156:in `process_all'
        from regression_test.rb:1290:in `rtest'
        from D:/Ruby21/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from D:/Ruby21/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from D:/Ruby21/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from D:/Ruby21/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from regression_test.rb:1363:in `<main>'
```

Oh, well, maybe next time ;=))

But returning to my batch testing I get the same/similar results as you. Now clean as far as exit code, as you indicate. Step one accomplished by this last change...

I got (aprx) 33 diferences, and on the 27 message compares, as you point out some are just the order of the warnings, which is understandable - and of the 6 html output differences, most seemed doctype, but still to look closely...

Concerning the `Info: doctype` diffs, like you my first thought was also we are not doing ConstrainVersion properly, but then noted commit 8bdaa2f763 added back this call in the one place you had removed it, so this needs more investigation...

So why do 5 message compare cases, 1004051 1015959 1241723 1410061-2 441568, have -

```
  New: Info: Document content looks like HTML 4.01 Transitional
  Old: Info: Document content looks like HTML 3.2
```

Tidy does add a `<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2//EN">`, so this just seems like the `Info:` output got it **wrong**! That is the `Info:` does **NOT** matching the doctype tidy added. That should **not** be possible... hopefully easily fixed...

And actually there are 8 outputs with `3.2//EN` - Why are 3 tests - 1445570 1986717-3 427818 - ok? That is very strange...

Do not know where the time went today? Well spent some time on the new 367 - line break on href, and some **more** time on the prepetual 348 - script parsing, and did not even get into eating my `Raspberry Pie` ;=))

balthisar:
> Just tried to run ruby regression_test.rb rtest -r=f:\path\to\tidy.exe but it always exits with an error...

Oh,I fixed that, but I've not been pushing the testing branch. I've been promised some time today, so I'll look at the issues above, and defer testing maintenance until later.

You've got a lot of willpower... with a new toy I would be tempted to disappear for a few days!

balthisar:
Actually, discount the test report above. It was based on an older binary. Actually things have regressed quite a bit more substantially than I imagined. Back to the drawing board...

balthisar:
Closing in favor of #370.

PR #358, opened 2016-01-26
Encode UTF-8 strings to hex escapes in header files

gagern:
In https://github.com/htacg/tidy-html5/pull/347#issuecomment-175000633 @balthisar wrote:
> * Using hexadecimal escapes doesn't work cross platform given the same source code. Clang on Mac OS X complains about invalid characters. GCC on Windows and Linux produce different results. It's as if each platform is interpreting the hex as its current locale, and that's a non-starter. There's no reason for Tidy to be converting character sets given a perfectly good UTF8 source.

I guess that was most likely because escape sequences of more than two hexadecimal digits were accidentially contained in the sources. This patch here avoids those by appending `""` if the next character could otherwise be misinterpreted as a hexadecimal character.

I must confess that I have been bitten by this as well. I had assumed that more than only up to two hex digits would be included in the codepoint, but found out in [this question on Stack Overflow](http://stackoverflow.com/q/5784969/1468366) that this was not the case, at least according to the standard.

### DIFF
--- a/localize/poconvert.rb
+++ b/localize/poconvert.rb
@@ -783,6 +783,13 @@ msgstr ""
       po_content.items.each do |key, value|
         longest_key = key.length if key.length > longest_key
         value.each_value do |value_inner|
+          value_inner[:string].gsub!(/[^\u0000-\u007e][0-9a-fA-F]?/) do |c|
+            esc = c[0].bytes.map{ |b| '\\x' + b.to_s(16) }.join('')
+            if c[1]
+              esc += '""' + c[1]
+            end
+            esc
+          end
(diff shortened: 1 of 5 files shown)


balthisar:
Very nice! I was continuing to do some experimenting on another branch, but yours is much more elegant. I may still bring over a command line switch.

balthisar:
I wish I'd found that Stack Overflow thread yesterday!

The msgunfmt tool is broken with this merged changed, but it should be simple to repair, just undoing your transformation. It's not an every day tool, but is useful when the main .h is updated with comments and such.

Issue #364, opened 2016-02-03
Can't install under Fedora 23


jsmad:
The rpm package provided by http://binaries.html-tidy.org/ does not install under Fedora 23.

The error provided by dnf is:

```
Error: Transaction check error:
  file /usr/share/man from install of tidy-5.1.25-1.x86_64 conflicts with file from package filesystem-3.2-35.fc23.x86_64
  file /usr/share/man/man1 from install of tidy-5.1.25-1.x86_64 conflicts with file from package filesystem-3.2-35.fc23.x86_64
```

geoffmcl:
@jsmad thanks for the report...

Sorry there is an install problem, and hope someone with Fedora 23 can shed some light on the problem...

hguemar:
It's a problem w/ CMake/CPack. I verified the package metadata, and it owns directories from base system package filesystem.
Fixed by the PR #366

```bash
rpm -qp --fileclass tidy-5.1.25-64bit.rpm  
/usr/bin/tidy	ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.24, BuildID[sha1]=a47d915eb211a8509ffefb1221d3425efc3db960, not stripped
/usr/include/tidy.h	C source, ASCII text
/usr/include/tidybuffio.h	C source, ASCII text
/usr/include/tidyenum.h	C source, ASCII text
/usr/include/tidyplatform.h	C source, ASCII text
/usr/lib/libtidy.so	symbolic link to `libtidy.so.5'
/usr/lib/libtidy.so.5	symbolic link to `libtidy.so.5.1.25'
/usr/lib/libtidy.so.5.1.25	ELF 64-bit LSB  shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=9703d64a92d489f39bb85e80a68eb5bb59098d43, not stripped
/usr/lib/libtidys.a	current ar archive
/usr/share/man	directory
/usr/share/man/man1	directory
/usr/share/man/man1/tidy.1.gz	troff or preprocessor input, ASCII text, with very long lines (gzip compressed data, from Unix, max compression)
``

hguemar:
Comparison with package generated using the fixed CMake file.
```bash
rpm -qp --fileclass tidy-5.1.*-64bit.rpm 
[...]

PR #366, opened 2016-02-04
Fix RPM generation

hguemar:
CPack generated RPM failed to install due to the RPM
owning directories owned by filesystem packages.
Exclude mandir directories from CPack.

Resolves https://github.com/htacg/tidy-html5/issues/364


balthisar:
It looks reasonable, but I don't have a Fedora system spun up to test this.

Does anyone else want to comment so that I can merge this?

PR #388, opened 2016-03-26
Merge fr.po to master

geoffmcl:
So far testing of this is looking good... will merge to master shortly...


geoffmcl:
After another round of french testing have merged this...

Issue #396, opened 2016-04-04
don't warn if ul tag without li as specified for html5


rtack:
https://www.w3.org/TR/html-markup/ul.html
says:
Permitted content:
Zero or more li elements

Following code get's a warning displayed even though it's valid:

```
<ul class="list-group">
    <a href="#submenu" class="list-group-item active" data-toggle="collapse" data-target="">someText</a>
...
```

line 2 column 1 - Warning: missing `<li>`

Edit: Added backcheck to element

geoffmcl:
@rtack thanks for the report...

I got quite a **surprise** when running your simple sample through a current tidy version 5.1.51, with the config `--show-body-only yes --show-info no`... 

It **converted** the `<ul class="list-group">` to a `<div...>`???

Output:

````
line 3 column 5 - Warning: missing <li>
line 2 column 1 - Warning: missing </ul>
Tidy found 2 warnings and 0 errors!

Info: Document content looks like HTML5
<div style="margin-left: 2em" class="list-group"><a href="#submenu"
class="list-group-item active" data-toggle="collapse" data-target=
"">someText</a></div>
````

So **yes** there is the warning about the missing `<li>`, and assume this would only apply to html5, and **not** to legacy html4-- documents? That is one thing to check and verify... Maybe the `Zero or more li elements` only applies to html5...

Then it also reports missing `</ul>`, but there is now no `<ul>`! It removed it... so why this report... maybe a minor niggle...

Really need help on understanding what is happening here! Under what circumstance will current tidy do this conversion? And why?

So at this point can do nothing until I understand more... comments, ideas, help **very** welcome... thanks...

**Comment by rtack** [comment]
yes I believe that in html4 it was 1 or more li.
in html5 it's 0 or more.

I plan looking into what's going sometime if nobody beats me to it.

converting ul to div is possibly (most likely?) a different issue.

**Comment by rtack** [comment]
It seems to be the case that with html4 at least one li element was required and with html5 it's 0 or more.

html4 spec for ul tag: https://www.w3.org/TR/html401/struct/lists.html#h-10.2
```
<!ELEMENT UL - - (LI)+                 -- unordered list -->
```
html5 spec for ul tag: https://www.w3.org/TR/html-markup/spec.html#ul
```
Permitted content:
Zero or more li elements
```


with tidy version 5.1.51
```
<!DOCTYPE html>
<html>
<head>
  <title>any</title>
</head>
<body>
  <ul>
    <a href="#submenu"></a>
[...]

PR #398, opened 2016-04-09
allow zero LI in UL when html5. fix for #396

rtack:


geoffmcl:
Now merged to `master`, and version bumped to 5.3.1... thanks...

PR #410, opened 2016-05-17
Pair va_copy calls with va_end

gagern:
According to the C99 spec, each [`va_copy`](http://en.cppreference.com/w/c/variadic/va_copy) call should be matched by a [`va_end`](http://en.cppreference.com/w/c/variadic/va_end) call to ensure proper cleanup.  Furthermore, since message filters might iterate over the list of arguments, we should hand a new copy to each filter, for the unlikely event that some user installed more than one filter.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -231,7 +231,7 @@ static void messagePos( TidyDocImpl* doc, TidyReportLevel level, uint code,
     {
         va_list args_copy;
         va_copy(args_copy, args);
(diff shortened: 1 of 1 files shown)


geoffmcl:
@gagern in a quick look I agree 100% with this **fix**...

All agreed?

geoffmcl:
@gagern oops nearly missed this good patch... thanks...

Issue #409, opened 2016-05-17
Easier handling of opaque types


gagern:
Tidy takes an imho pretty strange approach to opaque types. Publicly it declares them as pointers to some bogus structure, and internally it casts them between those and the actual types using a bunch of macros (and a bunch of functions in code conditioned by the `NEVER` preprocessor macro, so I'd consider this dead code).

Most other projects simply declare the structures globally but leave out their implementation. Applied to Tidy you'd get these in `tidy.h` (similar to what's currently in `forward.h`):

``` c
struct _TidyDocImpl;
typedef struct _TidyDocImpl *TidyDoc;
```

Then later in `tidy_int.h` you have the actual definition of `struct _TidyDocImpl { … };`, but that's invisible to library users. You could drop all those type converting calls, and operate on the non-opaque structures internally while keeping the internal layout out of the public API.

I'm willing to write a patch to this effect, but before I do so, I'd like to know whether there are any reasons against this approach.

geoffmcl:
@gagern you raise a very valid point, one I have had for a long time...

Why does tidy use an opaque structure? It does stop libtidy 
users messing with that structure... But is that needed?

And then you have to also ask about the TY(function)( params ) 
macro. Why 'hide' these functions? Some are static, but some 
are public...

So I guess part of it is being able to publish a very 
flexible API, and keep it documented, without the need 
to publish the meaning of every tidy doc member... every 
non-static internal C function...

I know tidy quite well, but that documentation would be the 
tip of an iceberg regarding some very neat C technologies 
in the code, which I hope one day **are** published...

Now, I am not in any way defending that as perhaps being a 
 "... pretty strange approach ..."? It has been well done, 
and reasonably documented... would always like to see more 
here, but where are the writers when you need them?

As you point out, a few minor changes could expose everything, 
but do we want that? At this time? To what gain?

And what additional doxygen needs to be added, to where, etc, 
such that that an extended API change can be `published`... 
and maybe install some additional 'tidy' headers, to expose 
more... add page to API documentation...

And there is the question of backward compatibility, if 
that is needed, or desired... Is this 5.6?

But I suppose, at present a documented API that gives you, 
to varying degrees, access to the structure, through API functions 
is what tidy presently does, in a C++ way, and that works too...

Anyway, @gagern this is my 2 bits on this ;=)) a good question!

gagern:
> Why does tidy use an opaque structure? It does stop libtidy users messing with that structure... But is that needed?

[...]

PR #411, opened 2016-05-18
Make internal and external types equivalent

gagern:
As discussed in #409, these changes turn the separate set of opaque types into equally opaque pointers to incomplete structures (i.e. declarations without definitions), with the complete definition added internally. In other words, the public API still doesn't know about the internal structure of any of these objects, but the tidy code can transparently operate them without any need to convert back and forth. As a consequence, the code allows more robust type-checking, is shorter and easier to read. Currently this diff removes 194 lines. There probably is some room for further improvements, since many of the public API functions are now just aliases for some private function, so that in a later commit we could eliminate the private names and instead implement the public functions directly. But the diff is long enough as it is, so I'll not do that just now.

`TidyIterator` doesn't actually name a single structure.  Often it's not a pointer at all, but an integer instead. So that will still be dealt with as before.

The `TidyOption` definition is somewhat special as well, since it contains the const qualifier, and the name of the structure doesn't follow the common schema. For that reason it doesn't make use of the `opaque_type` macro.

### DIFF
--- a/experimental/TidyNodeIter.c
+++ b/experimental/TidyNodeIter.c
@@ -3,7 +3,7 @@
 
 #include "TidyNodeIter.h"
 
-TidyNodeIter *newTidyNodeIter( Node *pStart )
+TidyNodeIter *newTidyNodeIter( TidyNode pStart )
 {
     TidyNodeIter *pThis = NULL;
     if (NULL != (pThis = MemAlloc( sizeof( TidyNodeIter ))))
@@ -14,7 +14,7 @@ TidyNodeIter *newTidyNodeIter( Node *pStart )
     return pThis;
 }
 
-Node *nextTidyNode( TidyNodeIter *pThis )
+TidyNode nextTidyNode( TidyNodeIter *pThis )
 {
     if (NULL == pThis->pCurrent)
     {
@@ -40,7 +40,7 @@ Node *nextTidyNode( TidyNodeIter *pThis )
     return pThis->pCurrent;
 }
 
-void setCurrentNode( TidyNodeIter *pThis, Node *newCurr )
+void setCurrentNode( TidyNodeIter *pThis, TidyNode newCurr )
 {
     if (NULL != newCurr)
         pThis->pCurrent = newCurr;
--- a/experimental/TidyNodeIter.h
+++ b/experimental/TidyNodeIter.h
@@ -8,7 +8,7 @@
 
   A traversal of the tree can be performed in a manner similar to the following:
 
-  Node *testNode;
+  TidyNode testNode;
   TidyNodeIter *iter = newTidyNodeIter( FindBody( tdoc ));
   for (testNode = nextTidyNode( &iter );
        NULL != testNode;
@@ -23,10 +23,10 @@
 
 typedef struct _TidyNodeIter
 {
-    Node *pTop, *pCurrent;
+    TidyNode pTop, *pCurrent;
 } TidyNodeIter;
 
-TidyNodeIter *newTidyNodeIter( Node *pStart );
+TidyNodeIter *newTidyNodeIter( TidyNode pStart );
 
 /* 
     nextTidyNode( TidyNodeIter *pIter )
@@ -36,10 +36,10 @@ TidyNodeIter *newTidyNodeIter( Node *pStart );
     and returns that value. When pTop == pCurrent, the function returns NULL
     to indicate that the entire tree has been visited.
 */
-Node *nextTidyNode( TidyNodeIter *pIter );
+TidyNode nextTidyNode( TidyNodeIter *pIter );
 
 /*
-    setCurrentNode( TidyNodeIter *pThis, Node *newCurr )
+    setCurrentNode( TidyNodeIter *pThis, TidyNode newCurr )
 
     Resets pCurrent to match the passed value; useful if you need to back up
     to an unaltered point in the tree, or to skip a section. The next call to 
@@ -48,4 +48,4 @@ Node *nextTidyNode( TidyNodeIter *pIter );
     Minimal error checking is performed; unexpected results _will_ occur if 
     newCurr is not a descendant node of pTop.
 */
-void setCurrentNode( TidyNodeIter *pThis, Node *newCurr );
+void setCurrentNode( TidyNodeIter *pThis, TidyNode newCurr );
--- a/include/tidy.h
+++ b/include/tidy.h
@@ -84,7 +84,8 @@ opaque_type( TidyDoc );
 /** @struct TidyOption
 **  Opaque option datatype
 */
-opaque_type( TidyOption );
+struct _tidy_option;
+typedef const struct _tidy_option * TidyOption;
 
 /** @struct TidyNode
 **  Opaque node datatype
--- a/include/tidyplatform.h
+++ b/include/tidyplatform.h
@@ -601,16 +601,13 @@ extern void* null;
 #endif
 
 /* Opaque data structure.
-*  Cast to implementation type struct within lib.
+*  The public header only declares these, but does not define them.
+*  The actual definition is restricted to the library code.
 *  This will reduce inter-dependencies/conflicts w/ application code.
 */
-#if 1
 #define opaque_type( typenam )\
-struct _##typenam { int _opaque; };\
-typedef struct _##typenam const * typenam
-#else
-#define opaque_type(typenam) typedef const void* typenam
-#endif
+struct _##typenam;\
+typedef struct _##typenam * typenam
 
 /* Opaque data structure used to pass back
 ** and forth to keep current position in a
--- a/src/access.c
+++ b/src/access.c
@@ -135,23 +135,23 @@ static const ctmbstr colorNames[] =
 
 
 /* function prototypes */
-static void InitAccessibilityChecks( TidyDocImpl* doc, int level123 );
-static void FreeAccessibilityChecks( TidyDocImpl* doc );
+static void InitAccessibilityChecks( TidyDoc doc, int level123 );
+static void FreeAccessibilityChecks( TidyDoc doc );
 
 static Bool GetRgb( ctmbstr color, int rgb[3] );
 static Bool CompareColors( const int rgbBG[3], const int rgbFG[3] );
 static int  ctox( tmbchar ch );
 
 /*
-static void CheckMapAccess( TidyDocImpl* doc, Node* node, Node* front);
-static void GetMapLinks( TidyDocImpl* doc, Node* node, Node* front);
-static void CompareAnchorLinks( TidyDocImpl* doc, Node* front, int counter);
-static void FindMissingLinks( TidyDocImpl* doc, Node* node, int counter);
+static void CheckMapAccess( TidyDoc doc, TidyNode node, TidyNode front);
+static void GetMapLinks( TidyDoc doc, TidyNode node, TidyNode front);
+static void CompareAnchorLinks( TidyDoc doc, TidyNode front, int counter);
+static void FindMissingLinks( TidyDoc doc, TidyNode node, int counter);
 */
-static void CheckFormControls( TidyDocImpl* doc, Node* node );
-static void MetaDataPresent( TidyDocImpl* doc, Node* node );
-static void CheckEmbed( TidyDocImpl* doc, Node* node );
-static void CheckListUsage( TidyDocImpl* doc, Node* node );
+static void CheckFormControls( TidyDoc doc, TidyNode node );
+static void MetaDataPresent( TidyDoc doc, TidyNode node );
+static void CheckEmbed( TidyDoc doc, TidyNode node );
+static void CheckListUsage( TidyDoc doc, TidyNode node );
 
 /*
     GetFileExtension takes a path and returns the extension
@@ -289,7 +289,7 @@ static Bool IsWhitespace( ctmbstr pString )
     return isWht;
 }
 
-static Bool hasValue( AttVal* av )
+static Bool hasValue( TidyAttr av )
 {
     return ( av && ! IsWhitespace(av->value) );
 }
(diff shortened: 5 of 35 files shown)


geoffmcl:
@gagern sorry, but do not understand? 

You say "As discussed in #409"? What was discussed? Anything agreed? Did I miss something?

As to the PR, so far all I can see is a lot of find and replace... ok, to what gain exactly? 

How will this improve tidy? What **bug** does it address? What is the desired `features` added? What do you find difficult to read in the code, where this will make it easier?, etc, etc, etc... ?

And just as an aside, please target such a large proposed code change in a branch, maybe `issue-409`, or something... but **not** the `master` branch... this has to be tested, agreed, discussed, understood, ... before committing it to the `master` branch...

But it looks like you have an **idea**, but can not yet quite see what it is ;=)) thanks, I think...

gagern:
In #409 I suggested what I did here, and your reply didn't indicate any pressing reason against this, so I went forward and coded the changes to make further discussion more specific.

This does not address an existing bug. It does not add a new feature. It is about maintainability. In my opinion, the change addresses the following issues:

1. Converting between external and internal types adds noise: additional function calls, additional variables, for what's essentially a no-op at the level of compiled code. 194 lines less code means 194 lines less to understand and maintain, so things become simpler.
2. Having different names for external and internal types can make you forget how they relate, while having the same names throughout helps users who know the public API to become contributors who enhance the internal API.
3. The type conversion macros did cast to the target type, without verifying the source type. Therefore, the compiler wouldn't issue a warning if you happened to pass in an inappropriate pointer into one of these conversion macros. (The conversion functions don't have this problem, but those were `NEVER` used.)
4. I'm no expert on this, but the fact that the opaque structures had some dummy definition might make them run afoul of some [strict pointer aliasing rules](https://gcc.gnu.org/onlinedocs/gcc-6.1.0/gcc/Optimize-Options.html#index-fstrict-aliasing-872). While I can't think of a piece of code where actual aliasing would occur, the compiler makes no guarantees so we might end up with undefined behavior. This would be a real bug, so perhaps there is a bug to be fixed after all.
5. Many other projects use pointers to incomplete structures in their public API. The fact that Tidy does something more elaborate made me (and may make others) wonder as to the reason for that added complexity, while apparently there is none.
6. Having equal types means that it's easier to promote a previously internal function to public API without having to change its signature.

I *did* write this in a separate branch. Namely [the one from which I'm proposing the merge](https://github.com/gagern/tidy-html5/commits/opaque). Feel free to test it from there. Feel free to discuss it here. If you need help setting up an additional git remote, I'll be glad to provide details. If you want things in the htacg repository, you'll have to duplicate the branch there, since I can only issue pull requests for existing branches.

I don't want you to merge this branch just because you trust me. Test and discuss as you see fit, ask questions, suggest changes. When you are satisfied, and convinced that this adds value, then merge this pull request. Since the change touches large portions of the code, I'll likely have to rebase it before you can merge it. But I'd like to do so only once, when you think it's ready for merging. I won't track every master commit here.

Most of the changes are indeed search and replace. And search and delete in the case of the conversion macro invocations. So let me direct your attention to the less trivial aspects.

1. The `opaque_type` macro in `tidyplatform.h` is now without definition, leading to an incomplete declaration. You can have pointers of those, but no other access to the structures.
2. The corresponding definitions in `lexer.h` and `tidy-int.h` now add a definition to the structure already declared in `tidy.h`.
3. In `forward.h` the declarations for the internal structures `TidyDocImpl`, `AttVal` and `Node` are gone.
4. The conversion macros in `tidy-int.h` and conversion functions in `tidylib.c` are gone.

That's the actual core of the change, the rest is just a consequence of that.

gagern:
I found [some useful information on the strict aliasing rules](http://stackoverflow.com/a/7291895/1468366). According to that, neither the old nor the new style of handling opaque types should be a problem, as long as noone attempts to access the (dummy) contents of the opaque type (i.e. dereferences the pointer). The only possible problem might come from alignment constraints.

On the other hand, some of the `TidyIterator` handling is likely relying on undefined behavior because conversion between integers and pointers isn't officially supported. But my change here doesn't improve that aspect; that would be a different issue.

geoffmcl:
@gagern, have downloaded, built, and done some testing of the result... which I versioned `5.3.1-exposed` :=))

While this PR has the diffs, a total diff I am reviewing is here - http:/geoffair.org/tmp/gagern.diff

And then paused a few weeks, waiting for comments from others... ok, none at this point... just you and me...

To be right up front, I do not like the merge yet! There are some **good** points, but to me it fails to reach the `50%` plus swaying point... really sorry...

Just to make sure we are on the same page, what I see is essentially what you have done is made a `TidyDoc doc` usable in `external` and `internal` code! That seems to defeat some of the idea of a tidy defined **opaque** type, but ok, I understand you are offering an alternative, that also works somewhat simularly? 

And done the same with some other pointers, like the `node *`, attr *... and I think added another... I do understand the different approaches to the same idea...

Just to put it in, hopefully, simple terms, currently an external app is passed a `TidyDoc tdoc` pointer. The fact that it is `tidy defined` opaque pointer means it can not address the structure, and equally important can **not** be use to call an internal library functions, not in the API, without most compilers complaining... aim achieved in current code...

When that `opaque` pointer is passed into the library it is **converted** to a `TidyDcoImp *`, which all/most internal libery services/functions use... Same with several other pointers...

After a fundimental header definition change, this leads to a mostly cut-and-paste, almost cosmetic, difference, like -

````
-void TY_(AccessibilityChecks)( TidyDocImpl* doc );
+void TY_(AccessibilityChecks)( TidyDoc doc );
````

So if we **agree** that is the discussion here, then to address your 6 points...

 1. Yes, if your calculations are correct we could save a few hundred instructions, which are just `stack a = b;`... But I do not think it really effects readbility to an exterienced programmer, or code maintainability overall... No.

 2. It could help new programmers, in that it is a tiny tidy code learning step. But once known, understood, is no problem... Yes.

 3. If a coder happened to pass in an inappropriate pointer into the API, they should suffer the consequence of that error, which will not always be flagged by every compiler/linker, **what ever we do**, so can not see how this improves that... No.

 4. I too know of no bug this fixes... sort of change with what gained? but open...

 5. How other projects do this is their business... the tidy creators constructed their own maco... why are we now choosing to abandon this? for simplicity? If it is wrong fix it, otherwise leave it alone... open

 6. ... easier to promote an internal service to the API? Yes, since with this change all internal library functions become very close to public, just not yet in `tidy.h`, ie not published... open...

Then you point us to the less trivial, fundimental changes in the important `definitions`

 1. `tidyplatform.h` yes, you changed the macro to a simpler macro, and changed the comment more to your choice!
 2. `lexer.h` and `tidy-int.h` yes, we understand, one change means more...
 3. `forward.h` as above, consequential to the first change...
 4. `tidy-int.h` more consequence to the above choice...

So the choice remains - should we stick with the current implemetation, or switch a few important definitions, and then accept 2. 3. 4.? A very difficult yes/no choice, at least for me... open...
 
So, in summary, while I agree there is some reduced verbosity and noise on pointers, I do not yet see a real gain at this time, and again wonder why we should change something that was put there by the tidy founders, that works... 

So to the central question here - Make internal and external types equivalent - I would suggest **no**!

But will certainly wait for other comments... thanks for the effort, and all the work done here... that is very great! Shows a fundimental understanding of tidy code... 

I hope you will apply the same enthusiam and efforts to fixing some other tidy **issues**...

gagern:
> […] which I versioned `5.3.1-exposed`

I'd consider that a misnomer, since it's not exposing any more than before.

> While this PR has the diffs, a total diff I am reviewing is here - [http:/geoffair.org/tmp/gagern.diff](http:/geoffair.org/tmp/gagern.diff)

That diff appears to be equivalent to what we have here in the “Files changed” tab, apart from end-of-line convention and path separator character following Windows. So I don't see the point of the distinction.

> Just to make sure we are on the same page, what I see is essentially what you have done is made a `TidyDoc doc` usable in `external` and `internal` code!

Yes.

> That seems to defeat some of the idea of a tidy defined opaque type, but ok, I understand you are offering an alternative, that also works somewhat simularly?

The data structures are opaque to the outside and transparent to the inside, both before and after my change. Previous code achieves the distinction by type casts, my code by incomplete types.

> Just to put it in, hopefully, simple terms […]. So if we **agree** that is the discussion here […]

I think we agree.


1. >> Converting between external and internal types adds noise: additional function calls, additional variables, for what's essentially a no-op at the level of compiled code. 194 lines less code means 194 lines less to understand and maintain, so things become simpler.

   > Yes, if your calculations are correct we could save a few hundred instructions, which are just `stack a = b;`... But I do not think it really effects readbility to an exterienced programmer, or code maintainability overall... No.

  I agree that performance is not an issue at this stage. Perhaps even the assignment you mention could be optimized away by compilers.

  Things might be a bit different if we use this simplification to remove redundant code in a future commit. For example, the public `tidySaveFile` is now (with the casts gone away) a simple forwarding to the private `tidyDocSaveFile`. We could just use the name of the former and the code of the latter to get rid of one function to maintain, and one function call on the call stack and performance bill. And there are several such forwarding functions. If you consider these savings relevant, I can add a commit to get rid of these as well.

2. >> Having different names for external and internal types can make you forget how they relate, while having the same names throughout helps users who know the public API to become contributors who enhance the internal API.

   > It could help new programmers, in that it is a tiny tidy code learning step. But once known, understood, is no problem... Yes.

3. >> The type conversion macros did cast to the target type, without verifying the source type. Therefore, the compiler wouldn't issue a warning if you happened to pass in an inappropriate pointer into one of these conversion macros. (The conversion functions don't have this problem, but those were `NEVER` used.)

   > If a coder happened to pass in an inappropriate pointer into the API, they should suffer the consequence of that error, which will not always be flagged by every compiler/linker, **what ever we do**, so can not see how this improves that... No.

 It's not an external user passing code to our API, but more about internal code. You are of course right that using wrong pointers should lead to broken runtime behavior, and that properly tested code should catch those in time. But personally I like the fact that a C compiler can detect many trivial mistakes before I even start testing code. Perhaps a matter of taste. Although seeing the state of affairs with respect to the tidyhtml test suite, I'd consider any help we can get pointing out mistakes to be a valuable one.

4. Strict aliasing should not be an issue, as I wrote.

5. >> Many other projects use pointers to incomplete structures in their public API. The fact that Tidy does something more elaborate made me (and may make others) wonder as to the reason for that added complexity, while apparently there is none.

   > How other projects do this is their business... the tidy creators constructed their own maco... why are we now choosing to abandon this? for simplicity? If it is wrong fix it, otherwise leave it alone... open

 For simplicity, yes. To lower the burden for potential contributors. With my strict aliasing concerns dispelled, the existing code is not wrong, just not very idiomatic either.

6. >> Having equal types means that it's easier to promote a previously internal function to public API without having to change its signature.

   > ... easier to promote an internal service to the API? Yes, since with this change all internal library functions become very close to public, just not yet in `tidy.h`, ie not published... open...

 Right. Which doesn't mean one should make functions public without due consideration, but after that consideration things become easier.

> I […] wonder why we should change something that was put there by the tidy founders, that works...

I wonder why the founders wrote it that way in the first place. It seems as though they didn't like to declare structures without defining them. Even the dummy structures used for the current opaque types come with a bogus definition. Perhaps there were some compilers not capable of properly handling such constructs back when the project started? I don't know, but I'm rather sure that these concepts are well-supported these days, and have been for quite some time.

> I hope you will apply the same enthusiam and efforts to fixing some other tidy **issues**...

I tend to address issues that affect me. I wanted to tidy HTML in Node, so I started the [node-libtidy project](https://github.com/gagern/node-libtidy). I see that it has some problems with how arguments are handled in error reporting, and would like to address those concerns by introducing a new opaque type. And I'd have preferred to do so my way, i.e. using an incomplete type in the public API. But I guess I'll rather stay with the style currently in use, and not wait on this issue here since I don't see this here getting accepted any time soon.

geoffmcl:
@gagern I do not think there would be any objection to a **new** opaque type being added, using the style currently in use...

And just a small negative to the suggested change here, when using the MSVC **Visual** C/C++ IDE, this opacity is very clear in the IDE local variable display, when debugging...

Presently the the tidy `opaque` pointer and the `impl` pointer show up **visually** different, making the situation very clean -

````
- tdoc	0x00000000004e64f0 {_opaque=0x00000000 }	const _TidyDoc *
	_opaque	0x00000000	int
- impl	0x00000000004e64f0 {root={...} lexer=0x0000000000000000 config={...} ...}	_TidyDocImpl *
 + 	root	{parent=0x0000000000000000 prev=0x0000000000000000 next=0x0000000000000000 ...}	_Node
 +	lexer	0x0000000000000000 {lines=??? columns=??? waswhite=??? ...}	_Lexer *
 + ... etc
````

That seems good, and very clear...

In the proposed `5.3.1-exposed` version there are not two pointers, they are the same, so the IDE will show the `doc` in full,  all the time - in and outside the library code - like -

````
- doc	0x00000000004e64f0 {root={...} lexer=0x0000000000000000 config={...} ...}	_TidyDocImpl *
 + 	root	{parent=0x0000000000000000 prev=0x0000000000000000 next=0x0000000000000000 ...}	_Node
 +	lexer	0x0000000000000000 {lines=??? columns=??? waswhite=??? ...}	_Lexer *
 + ... etc
````

So there is is a loss of a visual clue to the `opacity` of the first, versus the real `impl` structure pointer... making it **look** like you have access to say `doc->config.value[0].v`... of course the compiler will still belch an error if outside the library...

So of course the compile guards are still there. If you try to write code like this outside the library, the IDE/compiler will quickly stop you... but it does now **visually** imply you could...

Maybe not so important to coders that do not have or use a **visual** IDE to write code... but MSVC coders get to reply on it...

Anyway, I think you have correctly assessed the situation - without further support, reasons, ... - this merge will not be made, and I hope we can move on... thanks...

geoffmcl:
In the absence of further supporting comments, am closing this for the reasons given... but thanks for the work involved... it certainly shows another way to handle `transparent` types...

PR #418, opened 2016-06-08
Add support for the HTML5 "crossorigin" attribute

bdesham:
This attribute can only be used on "link" elements.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Attributes

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -553,6 +553,7 @@ typedef enum
   TidyAttr_COMPACT,           /**< COMPACT= */
   TidyAttr_CONTENT,           /**< CONTENT= */
   TidyAttr_COORDS,            /**< COORDS= */
+  TidyAttr_CROSSORIGIN,       /**< CROSSORIGIN= */
   TidyAttr_DATA,              /**< DATA= */
   TidyAttr_DATAFLD,           /**< DATAFLD= */
   TidyAttr_DATAFORMATAS,      /**< DATAFORMATAS= */
--- a/src/attrdict.c
+++ b/src/attrdict.c
@@ -395,6 +395,7 @@ const AttrVersion TY_(W3CAttrsFor_AUDIO)[] =
   INCLUDE_MICRODATA
   { TidyAttr_AUTOPLAY,              xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 },
   { TidyAttr_CONTROLS,              xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 },
+  { TidyAttr_CROSSORIGIN,           xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 },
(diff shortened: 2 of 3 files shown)


bdesham:
Actually, this can be used with the "img", "script", and "video" elements too. I'll modify my branch.

geoffmcl:
@bdesham thanks for this PR... in a quick review it looks **great**... I will certainly merge this in a day or so, unless there are any objections from others... and/or additions...

While we see `developer.mozilla.org` as an important informational source, we would prefer W3C references, and I have started to search for those... please advise any pointers you may have...

As stated, this looks good, and will probably merge directly to `master`... it certainly does not seem to pose any stability threat... but will pause a bit... thanks again...

bdesham:
Great! Here are links to the W3C’s latest documentation on HTML5, all of which mention the “crossorigin” attribute:

* [img](http://www.w3.org/TR/2014/REC-html5-20141028/embedded-content-0.html#the-img-element)
* [link](http://www.w3.org/TR/2014/REC-html5-20141028/document-metadata.html#the-link-element)
* [script](http://www.w3.org/TR/2014/REC-html5-20141028/scripting-1.html#the-script-element)
* [video](http://www.w3.org/TR/2014/REC-html5-20141028/embedded-content-0.html#the-video-element)

geoffmcl:
@bdesham, thanks for the great W3C link list...

Maybe I found one more?

* [audio](http://www.w3.org/TR/2014/REC-html5-20141028/embedded-content-0.html#the-audio-element)?

That is `W3CAttrsFor_AUDIO` in **`libtidy`**. Should this be added? Are there any others? thanks...

bdesham:
I added support for audio. This time I checked [the index](http://www.w3.org/TR/2014/REC-html5-20141028/index.html#attributes-1) and verified that "crossorigin" is only used on those five tags.

geoffmcl:
@bdesham thanks for the quick PR addition, and the [index](http://www.w3.org/TR/2014/REC-html5-20141028/index.html#attributes-1) link...

Just a matter of a day, or so... allow others to comment... thanks...

bdesham:
Any objection to getting this merged? 😃

geoffmcl:
@bdesham yes, sorry for the delay...

As noted bumped `version.txt` to `5.3.2 2016.06.18` for this change...

Thanks...

geoffmcl:
@bdesham  oops, we let the `TidyAttrId` enumeration get out of order with the dictionary table, `attribute_defs[]` - UGH! I should have checked!!!

### A big note to libtidy programmers ###

That includes me ;=))

So that we can easily address an attribute within the `attribute_defs[]` table, in `attrs.c`, using just its `ID`, rather than the usual searching, these **MUST be strictly kept** in the enum `TidyAttrId` order, in `tidyenum.h` - ie 1:1.

In the beginning this enumeration started life `alphabetically`, but that no longer applies. Then they were sort of added by groups... But that also makes a mess sometimes...

Now all **new** enumerations, and thus additions to the table should **always** just be added next, above the final `N_TIDY_ATTRIBS` entry, which **must** be last, in both the enumeration and table. This should avoid this problem in future...

And as a note to self, must check the API documentation to see if there is a clear message about this...

Have pushed a small fix, and bumped version to 5.3.5

This means we must watch out for versions 5.3.2, 5.3.3, and 5.3.4. These versions have this fault.

Anyway, we are back on track with this 5.3.5... thanks...

geoffmcl:
@bdesham can you look at #441 ... did we miss some cases, or is this different? 

Thanks...

geoffmcl:
@bdesham ok it seems we missed the `integrity` attribute... #441... hope you can help... thanks...

Issue #326, opened 2015-11-28
static linking by default? (on Linux)


vcunat:
(Suggested label: *Build/Install/Distribute*)

Currently when I build it, `bin/tidy` is statically linked, and both `lib/libtidy.so` and `lib/libtidys.a` are installed. I believe it would be better to default to dynamic linking and not installing `*.a`, at least on unix-like platforms.

Side note: I'm creating a package for [Nix / NixOS](http://nixos.org).

vcunat:
Ah, we already have a 4.9 package in there, separately from the old html-tidy. I assume this project is (also) meant as a drop-in replacement for the old html-tidy, including libtidy, right? From a quick skim I only see missing `bin/tab2space`, but perhaps that was rarely used.

geoffmcl:
@vcunat yes, at this time the tidy console binary is linked with the static library, mainly as a convenience to Windows users where we do not have a convenient DLL (shared library) install mechanism. Easier to have a stand-alone app...

And as far as I can see it poses no problem in linux. And in fact, as a developer, it is convenient for me having several stand-alone versions of tidy around for backward testing. But as things settle down we will consider changing this, for linux only, just to conform to the norm... but will continue to build and install both libraries, allowing 3rdParty libtidy or libtidys users a choice...

The `tab2space` is now an optional build, and even if built, is no longer included in the install or packages. This local only utility was just for developers that had not adjusted their favorite editor to conform to the tidy code style of using 4-space indents, and **no** tabs, so is seldom required, and there are other tools/utilities that can do the same job... probably better...

If you have a 4.9 package there already, I think this will be `tidy5`, and `libtidy5.so`, not sure? These should be removed... that suffix `5` was just used in development, when we were considering tidy in two forms - tidy for html4-- (old tidy), and tidy5 for html5++. We have since decided there should be only **one** tidy which handles all...

So, yes, 5.?.? is meant as a drop-in replacement for old tidy, removing that `5` suffix... it is good that you are updating the package for [Nix / NixOS](http://nixos.org). We hope other distribution will also update. We have some problem with distro with even older versions available... see #309 and others...

Will leave this open for a while for further comments... if any... but see nothing at present to do here...

vcunat:
Oh, is there no option to link `bin/tidy` dynamically? https://github.com/htacg/tidy-html5/blob/master/CMakeLists.txt#L189 We and most other Linux distributions tend to prefer dynamic linking. The old version is fine in this respect.

[...]

PR #421, opened 2016-06-13
Use dynamic linking as default and use standard define names

AdamMajer:
Most software distrubtions do not ship static libraries at all and
by policy, they refuse to ship software that is statically linked.
Hence default linking should be dynamic, with optional static for
niche cases.

Furthermore, standard %cmake macro in RPM package generation
relies on BUILD_STATIC_LIBS and BUILD_SHARED_LIBS - extra S. So change
the build package for least pain.

Reworked the build script to allow same name for static and dynamic
library. If this is a problem on some platforms, then that should be
done on case-by-case basis for those platforms. I think maybe only
Windows could affected here?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,17 +37,17 @@ else ()
 endif ()
 
 # By default, BOTH library types built, Allow turning OFF shared if not needed
-set( LIB_TYPE STATIC )  # set default message
-option( BUILD_SHARED_LIB "Set OFF to NOT build shared library"    ON  )
+option( BUILD_STATIC_LIBS "Set ON to build static library"         OFF )
+option( BUILD_SHARED_LIBS "Set OFF to NOT build shared library"    ON  )
 option( BUILD_TAB2SPACE  "Set ON to build utility app, tab2space" OFF )
 option( BUILD_SAMPLE_CODE "Set ON to build the sample code"       OFF )
 if (NOT MAN_INSTALL_DIR)
     set(MAN_INSTALL_DIR share/man/man1)
 endif ()
 # Issue #326 - Allow linkage choice of console app tidy
-option( TIDY_CONSOLE_SHARED "Set ON to link with shared(DLL) lib." OFF )
+option( TIDY_CONSOLE_SHARED "Set OFF to link with static lib." ON )
 if (TIDY_CONSOLE_SHARED)
-    if (NOT BUILD_SHARED_LIB)
+    if (NOT BUILD_SHARED_LIBS)
         message(FATAL_ERROR "Enable shared build for this tidy linkage!")
     endif ()
 endif ()
@@ -112,7 +112,7 @@ if (SUPPORT_GETPWNAM)
     add_definitions( -DSUPPORT_GETPWNAM=1 )
 endif ()
 
-if(BUILD_SHARED_LIB)
+if(BUILD_SHARED_LIBS)
    set(LIB_TYPE SHARED)
    message(STATUS "*** Also building DLL library ${LIB_TYPE}, version ${LIBTIDY_VERSION}, date ${LIBTIDY_DATE}")
 else()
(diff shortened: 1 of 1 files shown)


geoffmcl:
@AdamMajer thank you for the PR... but it covers separate things... not usually good for a PR...

**shared/static**

We have been over this topic a few times - see #326, and others - We have provided an option to link the console app with the shared library... Case solved?

I really do not know what is a `niche` case ;=)) but all cases are important for us... we want to run in every distribution imaginable... including my arm7 raspbian rpi machine...

BTW: Over the last few days I participated in the update of the `Tidy2.dll`, used by Notepad++ (NPP), to provide a HTML Tidy functions - ie pretty print a documents - Now, in generating that special DLL (shared) plugin library, using the latest HTML Tidy source, I chose to link `Tidy2.dll` with the static lib tidys. Why? Because then, in the install of NPP I can add a one file plugin, rather that a two file set `Tidy2.dll`, and `tidy.dll`, which in this case required a special WIN32 UNICODE build, so was already not the standard build...

Yes, the use of a shared library, ie the install 2 DLLS idea, gains by the fact that a user could update `tidy.dll` to the latest... But does he know how to do that? Where does he get such a DLL from? To install in an app specific location, in this case `C:\Program Files (x86)\Notepad++\plugins`, more or less a `private` location... requires `admin` priviledges! And that would miss the update of `C:\Program Files (x86)\Notepad++\plugins\doc\Tidy2\quickref.html`, update cfg files, ... all of which **is** handled well in unix... I know...

And this `tidy.dll` is installed in a `private` place, so can **never** actually be used as a `shared` library... ie never be used by any other processes in the system... that seems a waste, bloat, etc... When you link with a static library, only the modules your app uses are written into your exe, so, in a way, it can be less bloat... at least never more...

So yes, I guess there are `niche` out there where the **static** library is **important**... agreed, maybe much, much **less** in unix/mac/..., which have an easy, in place, documented, understood, shared library install, that all processes have access to... even deals with dependancies... **great**!

But tidy provides an option for this. It defaults to `OFF`, because that is `best` for windows users. But that can be set `ON` by unix users, much more savvy to command line switches... Best of both worlds? Shared culture...

**cmake macros**

Yes, we are aware of the cmake macros - with extra S - and chose to not use those... deliberately, not by accident... see no reason to change that yet...

But as I have gained more cmake experience, since that decision, and do now wonder `why not?`... was that the best choice? so this is for sure open... comments please...

**library names**

There are several ways do deal with the clash in library names in windows... CURL chose to rename the DLL library libcurl_imp.lib... others, like you suggest here, a big `-static.lib` suffix... etc, etc... 

We chose `tidys.lib`... which will be `libtidys.a` in unix... simple and elegent...

If this is **really** important to you, then maybe we can consider an option to set the static library name... I would certainly prefer the same naming convention be applied across unix/mac/windows... but could be persuaded otherwise... but it could be externally configurable... user selectable...

That would mess with some API documentation, ie the names used, and make the task of a cmake FindTidy.cmake script that much harder, and quite messy, but, just maybe... not sure... open

As usual, will leave some time for others to comment... but at this time, for the reasons given, and the previous discussions and decisions, suggest rejecting this PR as it presently stands...

geoffmcl:
In the absence of further supporting comments am closing this for now... for the reasons given...

AdamMajer:
I guess I missed your comments. My comments are that it is policy of every distribution (maybe except Gentoo that rebuilds everything anyway, not sure) that static libraries are not only discouraged and shared libraries must be provided and preferred (Debian), they are not to be distributed at all (SUSE). There are some exceptions with some software, but none of these apply here.

The reason has nothing to do with bloat of one executable. The reasons are,

 1. security
 2. bloat of archive
 3. security

And yes, security. If there is a bug that needs to be fixed in tidy, this only has to be done in one place not in 10 packages that use tidy.

Your comment about total executable size doesn't hold water when it comes to distributions. If I have 5 or 10 programs that link with tidy API, the size of one library is somewhat smaller than 5 or 10 copies of it inside executables. Furthermore, if these programs run concurrently, the size in the memory space is smaller too with dynamic linking (code segments are common to all programs).

Even when I wrote software for Windows, I always distributed DLLs for another reason - it's easier to track down bugs when your executable only contains your code, not 3rd party. It is much easier to follow executable code path in case of crash.

There are many more reasons for using dynamic libraries, but policies for all software distributions are quite clear that dynamic linking is to be used. The only case for static linking that I can think of you've brought out already - smaller total executable for one-of libraries.

As to cmake, using standard macros allows code to be build in a standard way, which is always nice for maintainers.

PR #422, opened 2016-06-14
prevent buffer overflow in debug output

sesom42:
This simple modification prevents a buffer overflow in debug output. Some debug output are larger than the defined `M_MAX_SPRTF`. `vsnprintf` checks the spezified size.


geoffmcl:
@sesom42 thanks for this PR...

It looks good, but need to experiment a little...

Will get back soonest... thanks...

geoffmcl:
@sesom42 ok, have now tested it for a while... it certainly seems a better option to output truncated debug data, than potentially overrunning the fixed buffer... thanks...

PR #423, opened 2016-06-24
Set HAS_FUTIME=0 on any GNU libc platform

pinotree:
GNU libc does not have `<sys/utime.h>` but `<utime.h>`, so set `HAS_FUTIME=0` for it; fixes build on non-Linux platforms using GNU libc.

### DIFF
--- a/include/tidyplatform.h
+++ b/include/tidyplatform.h
@@ -406,7 +406,7 @@ extern "C" {
 #if PRESERVE_FILE_TIMES
 
 #ifndef HAS_FUTIME
-#if defined(CYGWIN_OS) || defined(BE_OS) || defined(OS2_OS) || defined(HPUX_OS) || defined(SOLARIS_OS) || defined(LINUX_OS) || defined(BSD_BASED_OS) || defined(MAC_OS) || defined(__MSL__) || defined(IRIX_OS) || defined(AIX_OS) || defined(__BORLANDC__)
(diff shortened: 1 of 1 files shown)


danielhjames:
Hi @geoffmcl could you review this please? Seems like the best solution for fixing https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=828632 Thanks! Daniel

AdamMajer:
There should be a better way of doing this than what is currently done. 

https://cmake.org/Wiki/CMake:How_To_Write_Platform_Checks

geoffmcl:
@pinotree, @danielhjames thanks for this PR... adding `defined(__GLIBC__)` to an already long list looks ok...

Is there a better way to do this, as @AdamMajer suggests?

Well, yes in CMake terms there is, using say -

````
in CMakeLists.txt
check_include_file(utime.h     HAVE_UTIME_H)
check_include_file(sys/utime.h  HAVE_SYS_UTIME_H)
if (HAVE_SYS_UTIME_H)
add_definitions( -DHAS_FUTIME=1 )
else ()
add_definitions( -DHAS_FUTIME=0 )
endif ()
then maybe in a config.h.in
#cmakedefine HAVE_UTIME_H 1
#cmakedefine HAVE_SYS_UTIME_H 1
````

This could be add to the tidy cmake build, and thus define `HAS_FUTIME` before the header `tidyplatform.h` is used... that would hide any problem in the current `tidyplatform.h`, but...

But that does not help users of the binary tidy-dev install, while this PR addition does address that user's needs more... Of course they too could also add a `HAS_FUTIME` define in their build system, and again step over the problem...

Will leave a little more time for any others to comment, but at this stage this looks fine to be merged... thanks...

geoffmcl:
Now merged to master... thanks...

PR #424, opened 2016-06-27
Update LICENSE.md

danielhjames:
Change date to match LICENSE.txt


danielhjames:
Files src/access* and parts of src/message.c are Copyright University of Toronto so you may wish to update this file for that also.

geoffmcl:
@danielhjames thanks for this date update in `LICENSE.md`...

I am not sure why we have **both** a LICENCE.md **and** LICENCE.txt in the first place... but certainly they should agree...

IIRC it was something about when the documentation build was in the source... doxygen, or some other tool, preferred one and not the other, and a copy was made... but documentation has been moved to its own repo, so may be one, or the other, should be deletd here... or something...

Anyway, this looks good to upddate the `md` file...

And yes, it looks like this could/should be done in **many** places... just about every `src` file I looked at needs the year changed... but maybe that can be a separate issue... it should maybe be a once a year event... or maybe even the year should be removed from every src?

I also saw a query somewhere about the `Copyright University of Toronto` notice, but really do not know the answer to that... maybe some others can step in and comment on that...

As usual will leave a little time for others, but looks no problem... thanks...

balthisar:
Copyright is hard. Also, it's now 2016.

The license is the original W3C license (I believe), and any contributions _always_ continue to belong to the original contributor (barring a contributor agreement that Tidy has never had and that I would not support, e.g., an agreement assigning copyright to the W3C or HTACG).

Therefore the University of Toronto copyright doesn't poison anything per se; they contributed to an open source project under that open source license, and that open source licence still applies to us today. Not even we can change the licence, as all of the existing copyright holders (i.e., everyone who has ever contributed to Tidy) would have a valid legal complaint against us. (Well, we could change the license for _future_ code contributions, but that's not really useful, and would only apply to future code).

Given that the license applies to the original W3C code, I would suggest we leave the original W3C copyright intact, including the 2003 date (unless there have been official W3C contributions since 2003, and of course, HTACG is a W3C community group).

*However* there have been innumerable contributors since then. Many we know about, many have been lost in the mists of time. I would add another line to the licence file giving us something like this in the end:

~~~
Copyright (c) 1998-2003 World Wide Web Consortium (Massachusetts Institute of Technology, European Research Consortium for Informatics and Mathematics, Keio University). All Rights Reserved.
Additional contributions (c) 2003-2016 University of Toronto, @geoffmcl, HTACG, and others.
~~~

danielhjames:
Hi @geoffmcl, hi @balthisar,

I agree with Jim, as the HTACG is affiliated with the W3C there is no need to transfer any copyrights, and this transfer is not even possible in some countries. Therefore I see no need to put an 'end date' on the original copyright any earlier than 2016.

The reason that this issue came up was because in Debian, they noticed that there are copyrighted files (specifically the University of Toronto files) which are not mentioned in the main copyright notice.

Looking at the src/access* files, these say:

Copyright University of Toronto
  Portions (c) 1998-2006 (W3C) MIT, ERCIM, Keio University

It seems that W3C were involved until at least 2006, and http://tidy.sourceforge.net/ shows changes up to 2008. As these files say 'Modifications by : Terry Teague (TRT)' I think that dates them to between 2001 when Terry became a co-maintainer, and 2005, when Terry died. So it looks like the University of Toronto work was concurrent with the W3C work, but that this university was not part of the W3C.

Anyway, the two license files have some other discrepancies, like the Markdown version not crediting Dave Raggett, so I'll merge the two files and update the pull request.

geoffmcl:
Now merged to master... thanks...

PR #425, opened 2016-06-28
Restore balance in the force and the old header names

oerdnj:
Hi, renaming header files broke a lot of software that needs to be manually fixed and it was not necessary. Simply moving the header files into subdirectory will do the same job and it's much simpler to fix at the compile time because instead of editing sources you just add `-I/usr/include/tidy` (which is what f.e. Debian was doing already).

I understand that you wanted to fix the generic header names in the `/usr/include`, but the solution chosen is not optimal.

The PR contains two commits - one that restores the old header names and the second that setups compatibility symlinks with the new names. I don't think the second patch is needed and I think it should not be applied, but if there are any apps that are already using the new names, it might break those.

Please apply this as soon as possible before more harm is done to software relying on tidy library.


oerdnj:
I am fine with #420 as an alternative solution. Closing for now.

PR #426, opened 2016-06-29
Avoid crash in tidyCleanAndRepair if document was not loaded

nijel:
This can happen after attempt to load nonexisting file.

This used to work fine with old libtidy, but no longer works with html-tidy, so I think it is a regression. I'm not sure the fix is really correct, but it seems to do what is expected for me.

### DIFF
--- a/src/tidylib.c
+++ b/src/tidylib.c
@@ -1908,9 +1908,11 @@ int         tidyDocCleanAndRepair( TidyDocImpl* doc )
        it can ever be, so we can start detecting things that shouldn't
        be in this version of HTML
      */
-    if (doc->lexer->versionEmitted & VERS_HTML5)
-         TY_(CheckHTML5)( doc, &doc->root );
-    TY_(CheckHTMLTagsAttribsVersions)( doc, &doc->root );
+    if (doc->root.content) {
+        if (doc->lexer->versionEmitted & VERS_HTML5)
+             TY_(CheckHTML5)( doc, &doc->root );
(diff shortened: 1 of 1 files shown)


geoffmcl:
@nijel ok, what I understand is that your app calls `tidyDocCleanAndRepair(doc)` even when no document loaded! And this causes a crash in the current library, what version?, while it was harmless in an older library version, what version?

Well, that would certainly be a regression bug. Thanks...

But looking at your patch, I do not think this is quite right, even though as you say it works for you...

We need to look at why we need an extra guard for `CheckHTML5` **and** `CheckHTMLTagsAttribsVersions`... These calls should be harmless, like earlier `List2BQ`, and several others... with or without a loaded doc...

I think it might be the test `if (doc->lexer->versionEmitted & VERS_HTML5)` maybe the **crash**. Maybe there is no `lexer` if no document loaded...

It will take me a little time to setup a test for this... be back soonest... but look forward to any further information you can supply... thanks...

nijel:
It's [uTidyLib](https://pypi.python.org/pypi/uTidylib) which is doing this call even when file loading fails. It worked just fine with original Tidy and now leads to segfault.

To reproduce just invoke ``tidyParseFile`` on nonexisting file and then ``tidyCleanAndRepair``.

geoffmcl:
@nijel discussion and patch move to #429 ...

Meantime closing this as not being the correct patch... but thanks for pointing us to it...

Issue #429, opened 2016-07-01
Avoid crash in tidyCleanAndRepair if no doc, thus no lexer


geoffmcl:
@nijel I have moved this here, since investigation shows the PR #426 as **not** the correct fix for this...

As suspected `doc->lexer` is **null** if no document loaded... this causes the crash...

Of course I would remind you that what you are doing, namely calling `tidyCleanAndRepair` with no document loaded, is **not** correct libtidy API procedure... 

As can be seen in tidy.c, and the API docs, that clean and repair should only be called if there is a successful load, like -

````
    status = tidyParseFile( tdoc, htmlfil );
    if ( status >= 0 )
        status = tidyCleanAndRepair( tdoc );
````

But **OTOH**, it should not cause a **crash**, even if done...

The simple patch I see is as follows -

````diff
diff --git a/src/tidylib.c b/src/tidylib.c
index 28ebbb2..69ccf9b 100755
--- a/src/tidylib.c
+++ b/src/tidylib.c
@@ -1908,7 +1908,7 @@ int         tidyDocCleanAndRepair( TidyDocImpl* doc )
        it can ever be, so we can start detecting things that shouldn't
        be in this version of HTML
      */
-    if (doc->lexer->versionEmitted & VERS_HTML5)
+    if (doc->lexer && (doc->lexer->versionEmitted & VERS_HTML5))
          TY_(CheckHTML5)( doc, &doc->root );
     TY_(CheckHTMLTagsAttribsVersions)( doc, &doc->root );
 
````

Just check if there is a `lexer` before checking if `versionEmitted` still contains the `VER_HTML5` bits...

Hope you will get a chance to test this, after you rebase your fork to `master`, and remove your current patch from the log, and report... thanks...

**Comment by nijel** [comment]
With your patch I still get a crash:

```
#0  0x00007ffff507829c in prvTidyCheckHTMLTagsAttribsVersions (doc=0x273f560, node=0x273f560) at /home/nijel/work/tidy-html5/src/tidylib.c:1587
#1  0x00007ffff5078a37 in tidyDocCleanAndRepair (doc=0x273f560) at /home/nijel/work/tidy-html5/src/tidylib.c:1913
#2  0x00007ffff5077805 in tidyCleanAndRepair (tdoc=0x273f560) at /home/nijel/work/tidy-html5/src/tidylib.c:1179
```

geoffmcl:
[...]

PR #433, opened 2016-07-07
Avoid crash in tidyCleanAndRepair if document was not loaded

nijel:
These services can only be used when there is a document loaded, ie a
lexer created.  But really should not be calling a Clean and Repair
service with no doc!

See #429


geoffmcl:
@nijel wow, thanks for making this **easy**... now merged to master...

Issue #443, opened 2016-07-29
tidy ASAN build global-buffer-overflow bug


marcograss:
Hi,

I'm hitting this bug in the latest version 5.2.0 of tidy and also the master

> # ./tidy 
> 
> ==13070==ERROR: AddressSanitizer: global-buffer-overflow on address 0x082b12e6 at pc 0xf7177cf6 bp 0xffc16078 sp 0xffc15c48
> READ of size 7 at 0x082b12e6 thread T0
>     #0 0xf7177cf5  (/usr/lib/i386-linux-gnu/libasan.so.3+0x32cf5)
>     #1 0x80adab9 in tidySetLanguage /home/bob/VulnResearch/misc/xml/tidy-html5-5.2.0/src/language.c:705
>     #2 0x805339b in main /home/bob/VulnResearch/misc/xml/tidy-html5-5.2.0/console/tidy.c:1539
>     #3 0xf6fa7636 in __libc_start_main (/lib/i386-linux-gnu/libc.so.6+0x18636)
>     #4 0x8058553  (/home/bob/VulnResearch/misc/xml/tidy-html5-5.2.0/build/cmake/tidy+0x8058553)
> 
> 0x082b12e6 is located 58 bytes to the left of global variable 'tidyLanguages' defined in '/home/bob/VulnResearch/misc/xml/tidy-html5-5.2.0/src/language.c:38:26' (0x82b1320) of size 36
> 0x082b12e6 is located 0 bytes to the right of global variable 'result' defined in '/home/bob/VulnResearch/misc/xml/tidy-html5-5.2.0/src/language.c:603:17' (0x82b12e0) of size 6
> SUMMARY: AddressSanitizer: global-buffer-overflow (/usr/lib/i386-linux-gnu/libasan.so.3+0x32cf5) 
> Shadow bytes around the buggy address:
>   0x21056200: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
>   0x21056210: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
>   0x21056220: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
>   0x21056230: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
>   0x21056240: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
> =>0x21056250: 00 00 00 00 00 00 00 00 00 00 00 00[06]f9 f9 f9
[...]

PR #445, opened 2016-08-02
Fix static buffer overrrun (issue #443)

AdamMajer:
result[6] is a fixed array of size 6, but in the process
of copying data into it, we clobber the last allocated byte.

Simplify some of the code by not calling redundant functions.

### DIFF
--- a/src/language.c
+++ b/src/language.c
@@ -601,7 +601,6 @@ tmbstr tidyNormalizedLocaleName( ctmbstr locale )
     uint i;
     uint len;
     static char result[6] = "xx_yy";
-    char character[1];
     tmbstr search = strdup(locale);
     search = TY_(tmbstrtolower)(search);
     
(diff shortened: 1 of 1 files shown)


geoffmcl:
@AdamMajer thanks for the PR... eyeballing the code it certainly seems to avoid an overrun...

And the code feels cleaner, but wonder why `result[i] = tolower( search[i] );`... it seems it would already be in lower case from `search = TY_(tmbstrtolower)(search);`... but quite minor...

I have yet to test #443... but it certainly looks like this addresses that...

Will merge shortly, unless others see a problem... thanks...

AdamMajer:
maybe it is redundant but search can get overwritten by a lookup table and I didn't check if that lookup table is always lower case. I only removed crazy things like strncpy 1 char. :)

geoffmcl:
@AdamMajer well, it appear the table is in lowercase, but this protects against table changes...

Have now had a chance to add `-fsanitize=address`, test, and can repeat the bug, which seems fixed with your patch...

Accordingly, have merged it, and bumped to version **5.3.9** (2016.08.04)... thanks...

Issue #361, opened 2016-01-31
Add missing <meta charset="utf-8"> when html contains unicode characters


sria91:
The following command produces unreadable output.

```bash
$ echo '<!doctype html><html><head><title>ಕನ್ನಡ</title></head><body>ಸವಿಗನ್ನಡ</body></html>' | tidy.exe -i -utf8 -o output.html
```
Current `output.html`
```html
<!DOCTYPE html>
<html>
<head>
  <meta name="generator" content=
  "HTML Tidy for HTML5 for Windows version 5.1.34">
  <title>ಕನ್ನಡ</title>
</head>
<body>
  ಸವಿಗನ್ನಡ
</body>
</html>
```
Required `output.html`
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta name="generator" content=
  "HTML Tidy for HTML5 for Windows version 5.1.x">
  <title>ಕನ್ನಡ</title>
</head>
<body>
  ಸವಿಗನ್ನಡ
</body>
</html>
```

balthisar:
According to [The W3C](https://www.w3.org/TR/html5/document-metadata.html#character-encoding-declaration):

> If an HTML document does not start with a BOM, and its encoding is not explicitly given by Content-Type metadata, and the document is not an iframe srcdoc document, then the character encoding used must be an ASCII-compatible character encoding, and the encoding must be specified using a meta element with a charset attribute or a meta element with an http-equiv attribute in the encoding declaration state.

So first I don't see that this is a bug, but a feature request, because Tidy can't determine whether or not the page is delivered without the proper context type.

This leaves two points for discussion:

- Do we check for the existence of a UTF8 BOM before adding this metadata? BOMS are actually not recommended for UTF8, but if the BOM is present then the user agent should know that the file is UTF8, thus no metadata is required. What should Tidy do? (And what do we do if the options specifically state that a file is a different character encoding?)
- Next, do we use `meta charset` as requested, or use the also-allowed `<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>`?

That would address _just_ this request, but opens up a larger can of worms that we would have to address (I'm not saying "no", just pointing out that it's not as simple as limiting ourselves to the above). Right now Tidy doesn't try to interpret `http-equiv` or `meta charset`; instead we count on the `-encoding` options, both of which default to UTF8. This brings us the questions:

[...]

Issue #456, opened 2016-09-27
Feature request: add meta tag that matches actual output character encoding


AthanasiusOfAlex:
Tidy currently allows you to output a document in 14 character encodings. The problem is, unless you use ascii, the web browser generally gets confused by it. It would be nice if Tidy had the option of adding the "meta" tag that matches the encoding, e.g.,

`<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />`

for UTF-8 (which is the default). For the moment, I have to add this line manually.

geoffmcl:
@AthanasiusOfAlex thanks for the `Feature Request`

It seems a reasonable feature request, as a new option, to add this, styled depending on the document type detected, like `<meta charset="utf-8">` for html5, especially since it seems the W3C validator **barks** at you if it is missing...

Maybe the option could be say `--add-meta-charset`, or something... Boolean, default no...

There is an [OPTIONS.md](https://github.com/htacg/tidy-html5/blob/master/README/OPTIONS.md) which details the adding of an option... quite simple...

Would certainly consider this addition, if you or others presented code... thanks...

marcoscaceres:
I'm currently learning c, so I wouldn't mind taking this on (as it seems like a simple enough feature request).

@geoffmcl, could you mentor me on it? (i.e., give me some pointers as to where it might be good place to insert this in the code tree, etc.).

geoffmcl:
@marcoscaceres, thanks for your offer to take this on, and for sure I will help where I can...

As mentioned [OPTIONS.md](https://github.com/htacg/tidy-html5/blob/master/README/OPTIONS.md) gives the 4 steps to do this... and we should discuss and decide here the values for 1, 2, and 3... my first suggestion is as follows, but am open to other ideas -

 1. **ID** - `TidyMetaCharset`
 2. **table** - MS, "add-meta-charset", BL, no, ParseBool, boolPicks 
 3. **description** - "This option adds a charset meta tag, appropriate to the document type, in the document head."
 
As stated, these are just quick, first thoughts, suggestions, and look forward to them being massaged...

[...]

PR #458, opened 2016-10-04
WIP: Implement "add-meta-charset" option

marcoscaceres:
Covers initial implementation of "add-meta-charset".

@geoffmcl, this lays down the initial infrastructure. 

I have a couple of follow up questions: 
1.  Even though tidy picks up the new option, Tidy's command line output still shows: 

``` Bash
Config: unknown option: add-meta-charset=true
```

Anything obvious I've missed? The code is definitely working, as I can see the right element being added for HTML, and XHTML.
1. I'm getting null pointers when trying to match `*charsetAttr` and `*httpEquivAttr` in lexer.c, but can't work out why? Line 1691, 1692 below.  

I'm worried about what to do when we hit something like:

``` HTML
<meta http-equiv="content-type" content="text/html;charset=SOMETHING">
```

Where SOMETHING doesn't match the output encoding. I think all we can do there is warn the developer - as they have opted into this option, we can only assume they know what they are doing.

The warning would be something like:

``` BASH
line X column X - Warning: found redundant charset declaration. 
```

Question is, should I still just insert the correct `<meta charset>`? 

Lastly, if there is an valid declaration, that matches the output format, then it should be removed from the DOM. Then I can add the equivalent one to the document's head. 

WDYT?


marcoscaceres:
Actually, ignore this question:

> I'm getting null pointers when trying to match *charsetAttr and *httpEquivAttr in lexer.c, but can't work out why? Line 1691, 1692 below.

It started working. No idea 👍

marcoscaceres:
Ok, this is a rough version... let me know what you think about the general direction. 

I still need to look up how to emit the warnings properly.

marcoscaceres:
@garvankeeley, stupid noob question:... but debugging in lldb seems to only show me assembler :( I've tried to get this to work with XCode, so I can debug a little bit more sanely... I can get xcode to build (as an external source), but getting it to debug would be awesome. 

Any suggestions or will it just be a world of hurt?

marcoscaceres:
(I'm ok to just use lldb, but debugging assembler is no fun)

garvankeeley:
For debugging, i suspect the build does not have `-g -O0` (symbols on, zero optimization).

geoffmcl:
@marcoscaceres, @garvankeeley, sorry been busy on other things for a few days...

Have now had a chance to review the changes so far, and in general most looks good, but there are a few issues to be addressed...

 1. memory allocation
 2. C89 vs C99(11?)
 
The tidy library uses very indirect [Memory Allocation](http://api.html-tidy.org/tidy/tidylib_api_5.2.0/group__Memory.html). This allow **`libtidy`** users the ability to provide their own memory allocation and failure mechanisms. 

This is important when **`libtidy`** is embedded in bigger application like say perl, php, etc, where that application may want to deal with memory allocation/reallocation failure more calmly than just aborting...

So this means there will be no malloc/calloc/free in the library, except in the default alllocators... Please change to using this mechanism... or maybe use a stack buffer, if we are sure of the maximum needed...

While it maybe nice to apply the C99(11?) compiler standard to **`libtidy`**, this has not yet been done. I think you can add the switch `-std=c89` to gcc to set it back to that mode, and I think you will find the your lexer.c code will not compile.

In Windows I think any compiler before msvc140 (2015) will also bark at this... and in fact issue very difficult to understand error messages, like when using msvc100 -

````
  ..\..\src\lexer.c(1692): error C2143: syntax error : missing ';' before 'type' [F:\Projects\tidy-html5-charset\build\win64\tidy-static.vcxproj]
...
  ..\..\src\lexer.c(1694): error C2065: 'node' : undeclared identifier [F:\Projects\tidy-html5-charset\build\win64\tidy-static.vcxproj]
... and many more like this, ending with -
    42 Warning(s)
    79 Error(s)
````

In pure K&R `like` C, which **`libtidy`** is, you can **not** intermingle variable declarations, and code. All variable declaration must come before `any` code. And that includes `for` loops, like `for (Node *node = head->content;...`...

And while inline comments, ie `//`, are allowed (I think) in `c89` we do not use them in **`libtidy`** yet. Again I think you can add `-ansi -pedantic`? to gcc, to show these as an error. And am sure there is a similar switch for msvc in windows...

So while it may be time we discussed these changes for **`libtidy`**, that should be a separate discussion, and certainly not part of this PR. We do have a small [CODESTYLE](https://github.com/htacg/tidy-html5/blob/master/README/CODESTYLE.md) document, which certainly needs to be filled out more, but it is a start... As stated there, look at the current library coded in general for more...

And while this is still a WIP, which I understand, we have to eventually remove the `printf` statements, replacing them with the usual **`libtidy`** warning/error message system... which perhaps includes enumerating a new value, and adding the english output message, which can then be later translated...

Finally, when this is all sorted out, I would like it to be a clean PR, with very little internal history of change, but can help with that clean up when we get to that...

So as stated, really thanks for the effort and time. In general it all looks good. I may get some time this weekend to help further in this great effort... and hope you will keep going... you seem close... thanks...

**Comment by geoffmcl** [comment]
See issue #456 for initial discussion...

**Comment by geoffmcl** [comment]
Sorry for the extra noise, but just `experimenting` with the newly offered `review` process...

**Comment by garvankeeley** [comment]
thanks @geoffmcl! 

@marcoscaceres feel free to ping me (in or off thread) if you want me to look at anything, mainly have to make sure compiler is setup to catch the requirements (easy to miss all the needed changes otherwise).

**Comment by geoffmcl** [comment]
@marcoscaceres, continued on with my reviewing a little, using a C99 compatible compiler, and found a memory **bug** in these lines [1768-1770](https://github.com/htacg/tidy-html5/pull/458/files#diff-92587e94259bcadd0c463ae6237c5ecfR1768).

You are adding an attribute and **value** with `TY_(AddAttribute)( doc, node, "content", "text/html; charset=");`, but the `value` buffer now has the length `"text/html; charset="` + 1, and you can **not** do a following `TY_(tmbstrcat)(contentAttr->value, enc);`...

That would lead to a buffer overrun, which my msvc140 flags, when this buffer is later freed, at program exit...

You have to give the full `content` value to `AddAttribute`, so you need to prepare that **full** value, before that call... but I like your correct switch on `HTMLVersion` document type...

And in this case, where you need a buffer to prepare the full value string, and the earlier case where you use `calloc/free`, why not use a Tidy Buffer, where you have functions like `tidyBufAppend` which will automatically grow the buffer to contain the final full string...

Anyway running out of time today to continue with more now... So far have only tried samples with no current meta charset... Do want to build up some samples where there is already such a meta, and check the code that attempts to change this if not the correct type... be back soonest...

**Comment by geoffmcl** [comment]
@marcoscaceres,  in just a few more minutes, after declaring `TidyBuffer buf;` right at the top of the `TidyMetaCharset(doc)` function, c89 style ;=)), tried out some buffer code per the following patch -

````patch
diff --git a/src/lexer.c b/src/lexer.c
index 0a48e53..48005b4 100644
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -1682,6 +1682,7 @@ Node *TY_(FindBody)( TidyDocImpl* doc )
  */
 Bool TY_(TidyMetaCharset)(TidyDocImpl* doc)
 {
+    TidyBuffer buf;
     Node *head = TY_(FindHEAD)( doc );
     ctmbstr enc = TY_(GetEncodingNameFromTidyId)(cfg(doc, TidyOutCharEncoding));
     Bool charsetFound = no;
@@ -1765,9 +1766,11 @@ Bool TY_(TidyMetaCharset)(TidyDocImpl* doc)
             break;
         default:
             TY_(AddAttribute)( doc, node, "http-equiv", "content-type");
-            TY_(AddAttribute)( doc, node, "content", "text/html; charset=");
-            AttVal *contentAttr = TY_(AttrGetById)(node, TidyAttr_CONTENT);
-            TY_(tmbstrcat)(contentAttr->value, enc);
+            tidyBufInit(&buf);
+            tidyBufAppend(&buf, "text/html; charset=", 19);
+            tidyBufAppend(&buf, enc, TY_(tmbstrlen)(enc));
+            TY_(AddAttribute)( doc, node, "content", buf.bp);
+            tidyBufFree(&buf);
     }
     TY_(InsertNodeAtStart)( head, node );
     return yes;
````

And it seems to work fine... hope this helps...

**Comment by marcoscaceres** [comment]
This is great feedback, @geoffmcl! Thanks so much. I'll try to tackle everything over the weekend.

**Comment by marcoscaceres** [comment]
Just quick update: started making the requested changes. 

1. Hoisted variable declarations to top of function. 
1. now using TidyBuffer. 
1. move the using c89 style comments - will hopefully not need those in the end. Just for my own sanity right now. 
1. removed strcmp(), using tmbstrcmp

I have a question about this code, on line 1757... it is ok to do that dereference and cast?:

```c
            if(TY_(tmbsubstr)(lcontent, (ctmbstr) &charsetString)){
                printf("WARN ABOUT CLASH: %s \n", contentAttr->value);
            }
```

marcoscaceres:
@geoffmcl also, when I target c89, I get a linker error (but seems unrelated to my code). The compiler flags a bunch of issues in the existing code base about incorrect comments (again, not my code) and strings longer than 512bytes (also not my code).

marcoscaceres:
I'll hopefully get rid of the few remaining `printf`s soon too :)

marcoscaceres:
(doing a c++ course ATM... hoping to come back at this soon - excited to finish it)

geoffmcl:
@marcoscaceres how are things going with this?

No problem about the timing, but since we are headed for a `5.4` release soon, moving this out to `5.5`...

Can I help with anything more... thanks...

marcoscaceres:
@geoffmcl, this was getting close... just needed to replace the `printf()` with actual tidy warnings. I was unsure how to add test cases for this also. I could use some help with both of those things.

balthisar:
@marcoscaceres,

Check out the new `README/MESSAGES.md` in master to get started on where to add the strings.

Also take a look at `message.c/h` to try to use an existing output routine, which will ensure the message gets output to whereever CLI and LibTidy users expect them. If none of the existing output functions, e.g. `TY_(ReportNotice)` are suitable, feel free to write your own using one of the existing ones as a base. Don't be tempted to use `message` or `messageNode` or any of the non-namespaced functions directly.

You'll probably want to use `TY_(ReportWarning)`.

marcoscaceres:
Sadly, I have to concede defeat on this one :( I'm not finding enough time to work on it. 

Someone could take it over, as it was nearly there.

geoffmcl:
@marcoscaceres sorry to hear you are not finding the time to work on this ;=((

I have pulled your fork, did a **rebase** to current 5.5.20 tidy `next`, actually only 3 **conflicts**, easily solved... and now have a `meta-charset` branch that is 163 commits ahead of **your** fork origin... I could push these if I had commit rights on your repo...

**BUT** we probably need to close, remove the current PR #458, since this is a merge to `master`. Since then we have moved all development to the `next` branch and any/all pull requests now need to be against that `next` default branch...

I have built a tidy version **5.5.20.I458** using your now updated fork code...

Now I have started checking your **new** service, `TY_(TidyMetaCharset)(doc)`, using the 4 scenarios listed there -

   1. if there is no meta charset, it adds one.
   2. if there is a meta charset, it moves it to the top if HEAD.
   3. if it doesn't match the output encoding, warn about that.
   4. if there are duplicates, discard them.

In my first test html5 file [in_458-1.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_458-1.html), which has no `meta`, your 5.5.20.I458 correctly adds `<meta charset="utf-8">`, unless I add `--add-meta-charset no`, so it seems 1. is ok for HTML5++...

My next test file is `HTML 4.01` [in_458-2.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_458-2.html), which has no `meta`, your 5.5.20.I458 correctly adds `<meta content="text/html; charset=utf-8">`, unless I add `--add-meta-charset no`, so it seems 1. is ok for HTML4--...

**I have still to construct test files to test 2..4 combinations** ... but this is a **good** start...

But of course I note that service still has 2 **printf**, and tidy messages need to be added to replace these...

```
    if(TY_(tmbsubstr)(lcontent, (ctmbstr) &charsetString))
        printf("WARN ABOUT CLASH: %s \n", contentAttr->value);
and
    if(httpEquivAttr && charsetAttr)
        printf("WARN ABOUT HTTP EQUIV AND CHARSET ATTR! \n");
```

And then **fully** test all the other combinations...

Now yes, I could cut all these **changes** into current tidy `next`, or probably a branch, and **add** the two missing messages, and do some more testing, but then your name on the commits would be lost... 

Of course I would attribute the changes to you in the commit comments, but that is **not** the same thing... I would like to try to preserve your commits...

But, as stated, I would need commit access to your repo to do that, and to repeat, the first thing is to delete/close this current PR since that is against `master`... A new PR against `next` is needed...

So this seems **so close to completion**... what should we do next? how can I help? what little time can you offer? thanks...

PS: As I am suggesting we **must** close this, maybe we can move the discussion back to the **open** #456, which follows on from the closed #361...

balthisar:
Let's move the conversation back to the issue. If you'd like to pull it and push it as a branch here, we can continue working on it without requiring rights on @marcoscaceres repository.

geoffmcl:
**NOTE** - All discussion moved to #456 ... thanks...

geoffmcl:
**NOTE** created an `issue-456` branch, see #456, to continue work on this option... so closing this...

But @marcoscaceres again thanks for your efforts in this regard... and certainly hope you find the time to work on this or other parts of Tidy...

marcoscaceres:
@geoffmcl, thanks! I hope so too. We (folks working on web standards) are highly dependent on tidy and really appreciate the work you and @balthisar have been doing in maintaining tidy.

Issue #460, opened 2016-10-17
MAKE_C_FLAGS should probably have a default


marcoscaceres:
As the project uses c89, I'm wondering if the following should be the default to:

```
-DCMAKE_C_FLAGS:STRING="-std=c89 -pedantic"
```

benkasminbullock:
Those compiler options are specific to GCC (and Clang), so could cause errors when using Microsoft compilers.

marcoscaceres:
Is there equivalent for windows?

geoffmcl:
@marcoscaceres I agree 100% with this/these options for gcc/clang compilers... that support them...

The current CMakeLists.txt has two places where these options can be added... 

````
if(CMAKE_COMPILER_IS_GNUCXX)
    set( WARNING_FLAGS -Wall )
endif(CMAKE_COMPILER_IS_GNUCXX)

if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang") 
   set( WARNING_FLAGS "-Wall -Wno-overloaded-virtual" )
endif() 
````

Or a third, ...
```
if(WIN32 AND MSVC)
    # note dual, in Windows, and using MSVC...
else()
    # add any gcc flags
endif()
```

Really need help in setting the correct `gcc` flags,
as @arnaud always had in **his** build scripts... is it just the 2
flags - `-std=c89 -pedantic` - to be added? 

Welcome a patch, PR, ... on this gcc issue...

Concerning Windows/MSVC there are no `-std=...` flags, that
I know of... as the msvc advanced it supported, more to c11 
than before... seems msvc140 fully supports `c11`?!?!

But I remember, from somewhere, there may be a MSVC flag 
to treat the file as `pure` 'C'... at least as I remember,
in the early MSVC builds, it too would flag a warning about 
at least the use of `\\` inline comments...

Some help in finding this `flag` again would be appreciated...

So, yes, I agree `CMAKE_C_FLAGS` should probably have a default,
please help us in getting there... thanks...

marcoscaceres:
So, naive question: if c11 is backwards compatible with c89, and c99, and msvc140 fully supports c11, why not just target c11? Are there older systems that we need to support that can't handle c11 as a compile target?

geoffmcl:
[...]

PR #450, opened 2016-08-23
Add support for link 'as' attribute (closes #449)

marcoscaceres:

PR #463, opened 2016-10-25
style: ansi conforming comments

marcoscaceres:
Makes compiling a little bit less noisy when compiling on *nix with `-ansi` option is set. Also now adheres to projects commenting conventions.

### DIFF
--- a/console/tidy.c
+++ b/console/tidy.c
@@ -1105,18 +1105,18 @@ static tmbstr cleanup_description( ctmbstr description )
 
     /* Output Setup */
     tmbstr result = NULL;
-    int g_result = 100;  // minimum buffer grow size
-    int l_result = 0;    // buffer current size
-    int i_result = 0;    // current string position
-    int writer_len = 0;  // writer length
+    int g_result = 100;  /* minimum buffer grow size */
+    int l_result = 0;    /* buffer current size */
+    int i_result = 0;    /* current string position */
+    int writer_len = 0;  /* writer length */
 
     ctmbstr writer = NULL;
 
     /* Current tag name setup */
-    tmbstr name = NULL; // tag name
-    int g_name = 10;    // buffer grow size
-    int l_name = 0;     // buffer current size
-    int i_name = 0;     // current string position
+    tmbstr name = NULL; /* tag name */
+    int g_name = 10;    /* buffer grow size */
+    int l_name = 0;     /* buffer current size */
+    int i_name = 0;     /* current string position */
 
     /* Pump Setup */
     int i = 0;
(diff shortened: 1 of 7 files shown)


marcoscaceres:
Still getting noise from:
```bash
tidy-html5/src/message.c:233:9: warning: implicit declaration of function 'va_copy'
      [-Wimplicit-function-declaration]
        va_copy(args_copy, args);
        ^
1 warning generated.
```

A lot of noise from ones like this:

```bash
tidy-html5/src/language_en.h:109:9: warning: string literal of length 847 exceeds
      maximum length 509 that C90 compilers are required to support [-Woverlength-strings]
```

And this one is interesting too, as it uses a modern C99 feature?:

```
tidy-html5/src/language.c:30:25: warning: flexible array members are a C99 feature
      [-Wc99-extensions]
    languageDefinition *languages[];
```
(see also tidy-html5/src/language.c:30:25)

geoffmcl:
@marcoscaceres, wow, we have let things slip in this regard...

By the time I came back to tidy, I guess some of these `slips` had been done...

But is it the time to **reject** all the compiler advances since the `-ansi` option?

You have quite rightly highlighted many instances of where tidy has slipped a little... but are these wrong? Some maybe need to be looked at...

On the **single** issue of inline comments `//`, versus **only** `/* ... */`, I am very ambivalent... I do not see a problem accepting `//`... but that is just my view... 

Code I write at the moment exclusively uses `/*\ ... \*/`, but that is a **style** I have adopted for **tidy**...

Look forward to further comments... thanks...

marcoscaceres:
I don't mind so much, but the problem is when building the first time with "-ansi -pedantic" it generates a lot of noise because of the comments and string literals that exceed 509 bytes (related to #460). 
 
Also note that `languageDefinition *languages[];` is already in violation of c89 (and tidy has been shipping that for a while, I assume) - so trying to compile with c89 breaks the linker. 

So, this again raises the question about which c standard version to use (#460). Again, I'm really new to c, but It appears there is enough code currently in the project to say that this is a "c99"-conforming project. 

As someone who knows the code really well, could you maybe evaluate what is the right standard to use?

geoffmcl:
See #450 for additional comment...

Will try to pull and test this... thanks...

geoffmcl:
@marcoscaceres have now carefully reviewed your `ansi_compliance` branch, and this all looks good... 

Can you `rebase` your fork to master to ensure there are no conflicts, and will merge this... thanks...

marcoscaceres:
@geoffmcl rebased. Thank you for the review.

geoffmcl:
@marcoscaceres sorry for the brief delay... 

Are we still good to go on this?

marcoscaceres:
Should be all good to merge.

geoffmcl:
Done, and thanks...

Issue #469, opened 2016-12-04
Man page has repeated sections


jidanni:
```
$ man tidy|col -b|perl -we '
while (<>) {
    next unless /^[[:upper:]]/;
    $h{$_}++;
}

for ( keys %h ) { print "$h{$_} $_" if $h{$_} > 1 }
'
2 OPTIONS
2 DESCRIPTION
2 SYNOPSIS
```

geoffmcl:
@jidanni thanks for your issue...

Now I know very little about what is supposed to be in the tidy.1 man page, its format, but those headers also appear twice in the `man/tidy1.xsl.in` in the source, used to generate the man page, using `xsltproc` ... 

In other words, it does appear quite **deliberate** repetition! Is there something wrong with this?

Will leave it to others, who know about man page generation and format to comment... thanks...

jidanni:
Find me one other man page in the world with repeated sections.

And how, when they are converted to HTML, are they supposed to link
to a unique href="#synopsis" ?

And if your professor asks you for a synopsis, do you give him two?

He asked you for a synopsis, i.e., nutshell.

eric-brechemier:
For comparison, the issue is not present in the Mac version of tidy:

```
$ tidy --version
HTML Tidy for Mac OS X released on 31 October 2006 - Apple Inc. build 15.17

$ man tidy | grep '^[A-Z]'
TIDY(1)                                                                TIDY(1)
NAME
SYNOPSIS
DESCRIPTION
OPTIONS
USAGE
ENVIRONMENT
EXIT STATUS
SEE ALSO
AUTHORS
```

jidanni:
OK must be a Debian bug.
Created http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=847439 .

eric-brechemier:
@jidanni That's not what I meant. Apple has forked the [legacy version of Tidy](http://tidy.sourceforge.net/). And the issue is not present in [their version](https://opensource.apple.com/source/tidy/).

Reference:
https://www.apple.com/opensource/

Now, I wonder what causes this behavior.

Here is the `tidy1.xsl` file in the Apple version that I ran:
https://opensource.apple.com/source/tidy/tidy-15.17/tidy/htmldoc/tidy1.xsl

The version of `tidy1.xsl` in this project:
https://raw.githubusercontent.com/htacg/tidy-html5/master/man/tidy1.xsl.in

jidanni:
[...]

Issue #472, opened 2016-12-15
TidyDoctypeMode option documentation missing


ralfjunker:
The `TidyDoctypeMode` option is not documented. Documentation is missing both online at http://api.html-tidy.org/tidy/quickref_5.2.0.html#doctype as well as built into libtidy.

There is just a short comment in `tidyenum.h` which asks to "See doctype property". Unfortunately, `TidyDoctype` does not mention `TidyDoctypeMode`.

In particular,

* why are there two similarly named properties and 
* what is the difference between them?

Could you please add the missing `TidyDoctypeMode` documentation?

geoffmcl:
@ralfjunker wow, you raise an interesting question or 2, which leads to a potential **bug**... thanks...

As far as I can see, there is **no** actual option, `TidyDoctypeMode`, aka `--doctype-mode ????`, and will lead to a crash if used, so there should be no **documentation** of it! It is only used internally, in the library...

And there should be a **fix** to avoid that **crash**... In the `option_defs[]` config table, it will be noted that it is the option with a `NULL` parser, except of course for the first and last entries...

Perhaps the `tidyenum.h` comment should be something like - "Not a valid otpion! Only used by 'doctype' property" - or something...

To try to explain more. This particular `--doctype enum` option has two forms - 

 1. as an enumerated int value, and 
 2. in one case a user `string`

And as @jidanni points out in #435, all this is **not** well explained in - http://api.html-tidy.org/tidy/quickref_5.2.0.html#doctype! Nor, as also pointed out, completely implemented, in that a user string type seems ignored...

I think because of this **dual** option type the actual enumerated int is kept in the `TidyDoctypeMode` line in the config table, but if that is set to `TidyDoctypeUser`, then the user's string will be kept in the `TidyDoctype` line... certainly a bit confusing, for all...

So first, the simple **bug** I see. If a person enters an option `--doctype-mode <value>`, the first part of the config parser will find it, and see it as a **valid** option, but will blow up when the NULL `option->parser` is called. Note the following quick patch would solve that -

```patch
diff --git a/src/config.c b/src/config.c
index ddb677c..343b337 100644
--- a/src/config.c
+++ b/src/config.c
@@ -924,8 +924,12 @@ Bool TY_(ParseConfigValue)( TidyDocImpl* doc, TidyOptionId optId, ctmbstr optval
     const TidyOptionImpl* option = option_defs + optId;
[...]

PR #473, opened 2016-12-18
Merge documentation of "command-line" and "configuration" options (Issue #469)

eric-brechemier:
I have generalized the SYNOPSIS and the description of OPTIONS to cover both "standard" options and "extended" configuration options in the same sections of the documentation.

I have modified the [XSLT transformation](https://github.com/eric-brechemier/tidy-html5/blob/issue-469/man/tidy1.xsl.in) which generates the man page. I have also added two temporary files, the [generated man page](https://github.com/eric-brechemier/tidy-html5/blob/issue-469/tidy.1) (source) and the [formatted text](https://github.com/eric-brechemier/tidy-html5/blob/issue-469/tidy.1.txt) (for display in terminal) generated by `groff`.

The latter is best viewed in a terminal, e.g. using `less tidy.1.text` to view the file, or `git log -p tidy.1.txt` to review the effect of individual commits on the resulting document.

This is a work in progress submitted for preliminary review and discussion.

Before this branch is merged, I intend to simplify the description of each configuration option to avoid repeating information, make entries shorter and similar in format to the description of standard options. I will also delete the temporary files once the review has been completed.


geoffmcl:
@eric-brechemier have reviewed this WIP, and it is looking great... some little points...

#### SYNOPSIS

I like you have reduced this to one(1) single simple line, as it should be... but maybe it should read like -

```
SYNOPSIS
   tidy [options] [file[ ...]] [[options] [file[ ...] ...]
```

But maybe the additional square braces are **too** much, too `pedantic`... what do you think?

And in fact maybe it could be reduced to **just** `tidy [options] [file1[ file2[ ...]]]`, and the fact that this `[options] [file(s)]` can be repeated be only mentioned say in the description. Or not at all...

I have never really used, or tested, this multiple `[options file]`, since as far as I can see, if tidy exits `1`, you would not know, without actually reading the output, which file caused this warning... but assume it works... but what really is the use case?

In any batch, or scripted processing, I can understand having one set of options applied to a big list of files, although again, you do not know which actual file caused the exit value... but to then change the options for another file, or files, does not seem to make much sense...

But ok, maybe this is not the place or time to discuss this quite unique multiple `options file(s)` that tidy offers... but I would be ok with **not** suggesting, sort of promoting this...

As stated, I am happy with the single line, with, or without, the addition square braces, as it is...

#### OPTIONS

Again it is great that you have added a good description, clearly pointing out tidy has two types, `-`, and `--`. I suppose some might argue that in *nix systems two dashes, `--`, is more the 'regular', or 'standard', but we can wear that semantic difference... and your description makes the intention clear...

And I like that, for most single dash, `-`, you have added a double dash, `--`, form, which can then be put in a config file...

And maybe it would be too much if each double `--` option mentioned a `-` option, where it exists... like say `clean` with a `See also: -c`... that is each is fully crossed referenced... but as I say, maybe **too much**?

#### Temporary Files

Yes, thanks for making this effort, to be able to read them in a non-nix system... 

But in *nix I found I could just compile tidy in say build/cmake, and view the results with `man -l tidy.1`, **without** doing any **install**...

And I certainly hope others, espceially @jidanni, can do this, and offer his comments, since he opened #469. I do understand that he wants to use only released tidy in production, but as stated this great new man page change does **not** need to be installed - just in a few mintues -

```
$ cd some/tmp/work/dir
$ git clone git@github.com:htacg/tidy-html5.git
$ cd tidy-html5/build/cmake
$ cmake ../..
$ make
$ man -l tidy.1
```

So I think these 'temporary' files can be remove, but as stated, thanks for the thought...

#### Multiple, multiple commits

While I have no real problem with this as you were developing the change, but hopefully the final PR is essentially only one commit, since as far as I can see it only effects one file, man/tidy.xsl.in... 

You would probably need copy your final modified `tidy.xsl.in` to say `new-tidy.xsl.in`... back up to `master`... make sure it is rebased to the current `master`... create a new branch, say `man-final`... overwrite `tidy.xsl.in` with your new... maybe commit with a multiline commit message... and push... otherwise I understand github just adds each new commit to this existing WIP PR...

But, as stated, not really a problem...

#### Other possible changes

1. Yes, we are always looking for better, more concise, suscinct descriptions, but these each can be raised as separate issues...

2. I wish we could remove `internal` only option, like `gnu-emacs-file`. Maybe these should **not** appear in any docs, even with a comment `Used internally.`. See #472 for some discussion on others like this... But again these could be dealt with as separate issues. 

So, as stated, this is a sterling effort... look forward to the final... thanks...

eric-brechemier:
**SYNOPSIS**

> But maybe the additional square braces are too much, too pedantic... what do you think?

Maybe not pedantic, but definitely confusing in a lispy way. I'll think about it.

> And in fact maybe it could be reduced to just tidy [options] [file1[ file2[ ...]]], and the fact that this [options] [file(s)] can be repeated be only mentioned say in the description. Or not at all...

This is the option taken by `man curl`:

```
curl [options] [URL...]
```

but then they have a special parameter to separate options that are related to one URL from the next:

```
-:, --next
              Tells curl to use a separate operation for the following URL and
              associated  options.  This  allows  you  to  send  several   URL
              requests,  each  with  their  own specific options, for example,
              such as different user names or custom requests for each. (Added
              in 7.36.0)
```

while in the case of `tidy`, an option applies to all the following files.

> In any batch, or scripted processing, I can understand having one set of options applied to a big list of files, although again, you do not know which actual file caused the exit value... but to then change the options for another file, or files, does not seem to make much sense...

It may be useful when using a wildcard from the shell: `*.html` is expanded to the list of files that matches the pattern. Also, I remember a particular time when we had different files with different encodings, and used to include a hint for the encoding of the file in the file name, such as `.utf8.html` or `.latin1.html`. We could apply `tidy` to all these files in a single run with `-latin1 *.latin1.html -utf8 *.utf8.html`.

eric-brechemier:
**OPTIONS**

> And maybe it would be too much if each double -- option mentioned a - option, where it exists... like say clean with a See also: -c... that is each is fully crossed referenced... but as I say, maybe too much?

Ah, good point. Some options are available in both flavors. That deserves at least a mention when it is the case. I will look into it.

eric-brechemier:
**Temporary Files**

> Yes, thanks for making this effort, to be able to read them in a non-nix system...
>
> But in *nix I found I could just compile tidy in say build/cmake, and view the results with man -l tidy.1, without doing any install...

I looked long and large for this option, which is simply not available in macOS's man :( I am glad to know that it exists though.

**Multiple, multiple commits**

> While I have no real problem with this as you were developing the change, but hopefully the final PR is essentially only one commit, since as far as I can see it only effects one file, man/tidy.xsl.in...

Sure, I will [squash](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git#5201642) the commits into one, leaving no trace of the temporary files. Apparently, it is also possible to do it [directly from GitHub interface](https://github.com/blog/2141-squash-your-commits) when you merge the branch.

eric-brechemier:
**Other possible changes**

> 1. Yes, we are always looking for better, more concise, suscinct descriptions, but these each can be raised as separate issues...

It is not so much about the text that describes each option as the verbose format used to describe configuration options:

```
       add-xml-decl

              Type:    Boolean
              Default: no
              Example: y/n, yes/no, t/f, true/false, 1/0

              This option specifies if Tidy should add the XML declaration
              when outputting XML or XHTML.

              Note that if the input already includes an <?xml ... ?>
              declaration then this option will be ignored.

              If the encoding for the output is different from ascii, one of
              the utf* encodings, or raw, then the declaration is always added
              as required by the XML standard.

                                      See also: char-encoding, output-encoding
```

vs (for example)

```
       --add-xml-decl <boolean:no>
              This option specifies if Tidy should add the XML declaration
              when outputting XML or XHTML.

              Note that if the input already includes an <?xml ... ?>
              declaration then this option will be ignored.

              If the encoding for the output is different from ascii, one of
              the utf* encodings, or raw, then the declaration is always added
              as required by the XML standard.

              See also: char-encoding, output-encoding
```



> 2. I wish we could remove internal only option, like gnu-emacs-file. Maybe these should not appear in any docs, even with a comment Used internally.. See #472 for some discussion on others like this... But again these could be dealt with as separate issues.

I could test whether the `<description>` matches `Used internally. ` but a better way would be to remove the option from the output of `./tidy -xml-config` (which is used to generate the file `tidy-config.xml` that is fed to the XSLT transformation).

eric-brechemier:
@geoffmcl I will move forward with three points in mind:

1. further enhancing the SYNOPSIS if possible
2. looking into cross-referencing of options available in both single and double dash flavors
3. compacting the configuration options

geoffmcl:
@eric-brechemier, wow, thank you for the detailed, and very thoughtful responses...

#### SYNOPSIS

Really, my additions here were **only** a suggestion... I basically agree with going forward with the single line `SYNOPSIS` proposed, as in the current first `SYNOPSIS` used...

Forget my additional square brackets as **too** confusing, but not sure what you mean in a `lispy` way... and yes, like `curl`, tidy does offer this multiple `options file(s)` possibility, so lets go forward with the single line `synopsis` as offered... ok?

#### more concise, succinct descriptions...

And yes, I would support **not** repeating such mundane information, like I do prefer, say -

```
       --add-xml-decl <boolean:no>
            description
```

versus our currently very **repetitive**, space using form of -

```
       add-xml-decl

              Type:    Boolean
              Default: no
              Example: y/n, yes/no, t/f, true/false, 1/0
              description
```

The main `DESCRIPTION` should take care of this type of multiple acceptable inputs of a config option type, which it nearly does already...

#### showing non-options

No, the `tidy.1` build should **not** have to seek `Used internally.`! As you suggest, the `./tidy -xml-config` should be adjusted to just **not** output these as `options`, so forget this here... 

That is another issue about what the `-xml-config` output, that is fed to the XSLT transformation... agreed as a separate issue... the man page generation should absolutely use what it is fed...

#### compacting commits

Thank you for the links on this, and now I have seen that I can choose a `Squash and merge` commit that I did not know about before... look forward to trying this when you remove the `WIP` moniker from the PR... thanks...

Now I do think the cross referencing the `extended`, `--`, options back to those that have a single `-` option quite important, but not essential...

And now realize my suggested script to @jidanni, and others, was in error. They need to be cloning, and building **your** fork, and specific branch, not the htacg master, at this time... to test this change...

Again, thanks for your efforts on this... very much appreciated...

eric-brechemier:
@geoffmcl All clear. Thank you for the feedback.

> not sure what you mean in a lispy way...

I am currently reading [the original paper on Lisp](http://www-formal.stanford.edu/jmc/recursive.html) and the mental gymnastics involved in tracking opening and closing brackets in `tidy [options] [file[ ...]] [[options] [file[ ...] ...]` reminded me of the similar effort required to understand an expression such as `apply[(LAMBDA,(X,Y),(CONS,(CAR,X),Y)); ((A,B),(C,D))]`.

eric-brechemier:
> (...) I suppose some might argue that in *nix systems two dashes, --, is more the 'regular', or 'standard', but we can wear that semantic difference... and your description makes the intention clear...

I revisited this issue and enhanced the wording in 29215e3. The terms are now in line with the descriptions of `-xml-help` and `-xml-config`: (purely) *command-line* options (single dash) and *configuration* options (double dash).

eric-brechemier:
> And maybe it would be too much if each double -- option mentioned a - option, where it exists... like say clean with a See also: -c... that is each is fully crossed referenced... but as I say, maybe too much?

I gave more thought to this. I noticed that the documentation of some *command-line* options includes an equivalent *configuration* option **and value**, in the `<eqconfig>` element.

For example:

```
 <option class="process-directives">
  <name>-clean</name>
  <name>-c</name>
  <description>replace FONT, NOBR and CENTER tags by CSS</description>
  <eqconfig>clean: yes</eqconfig>
 </option>
```

Therefore, a *configuration* option alone is not equivalent to using a *command-line* option: it is a pair option+value that must be used.

This could justify having references only in one direction from *command-line* options to *configuration* options, and no references back.

If we do want to include the references from *configuration* options back to *command-line* options, mixing both kinds of options in the 'See also' section, we should use the same mechanism currently used to refer only to the double dash options, using `<seealso>` elements. Each `<seealso>` element includes the name of a single option, without the dashes:

```
  <seealso>input-encoding</seealso>
  <seealso>output-encoding</seealso>
```

We could change the format of `<seealso>` elements in the output of `tidy -xml-config`, adding the dashes to the option names:

```
  <seealso>--input-encoding</seealso>
  <seealso>--output-encoding</seealso>
```

This would allow to mix references to both kinds of options:

```
 <option class="print">
  <name>indent</name>
  <type>AutoBool</type>
  <default>no</default>
  <example>auto, y/n, yes/no, t/f, true/false, 1/0</example>
  <description>This option specifies if Tidy should indent block-level tags. If set to "auto", this option causes Tidy to decide whether or not to indent the content of tags such as TITLE, H1-H6, LI, TD, TD, or P depending on whether or not the content includes a block-level element. You are advised to avoid setting indent to yes as this can expose layout bugs in some browsers. </description>
   <seealso>-indent</seealso>
   <seealso>--indent-spaces</seealso>
 </option>
```

Which would result in:

**See also:** `-indent`, `--indent-spaces`

eric-brechemier:
**Note:** I still need to delete the temporary `tidy.1` and `tidy.1.txt` at the root before the merging.

geoffmcl:
@eric-brechemier ok, have pulled your fork, `issue-469` branch, built and reviewed this, in linux, and everything looks good, but really wish there were others mac/unix reviewers. Being a windows person, which does **not** have a `$ man page` command, do not feel totally comfortable with just my quick linux `$ man -l tidy.1` review...

As far as I can see, so far this is only modification of one file, `man/tidy1.xsl.in`. And it seems to me you have completely removed the **duplications**, and improved the wording here and there... that effectively solved the #469 issue raised by @jidanni... As I say this looks good to me... and once you remove the two temporary files, and hopefully a positive review by at least one other, would be prepared to merge this... many thanks...

And if I read, and understand things correctly, as a further enhancement, you would now like to modify the `xml` output of tidy.c. That is the modify the `xml` fed to the transformation. If I am understanding this correctly then I would like that as a separate issue. I need to explore how that is done presently...

Like I see in the code - `printf("  <seealso>%s</seealso>\n",tidyOptGetName(optLinked));`, but presently do not immediately see how to know to add a single `-` or a double `--` in front... The `tidyOptGetName(optLinked)` is a tidy API call, which probably can not be changed, but maybe another internal function could be added to do that... or something...

I certainly like the idea the we could have xml ouput of -

```
   ...
   <seealso>-i</seealso>
   <seealso>-indent</seealso>
   <seealso>--indent-spaces</seealso>
 </option>
```

which in turn would add **See also:** `-i`, `-indent`, `--indent-spaces`... to the tidy.1 page... That is fully cross-referencing the options... But as stated else where this is **not** essential that we have every cross-reference...

But is it convenient to separate this as a further, separate step?

Regards, geoff.

eric-brechemier:
@geoffmcl Agreed. It's a wrap!

I noticed a few other things that I would like to improve; let's open separate issues for further enhancements.

I have removed the temporary files from the branch.

geoffmcl:
@eric-brechemier yes, it's a wrap, even if we do not get another unix reviewer...

Just about out of time tonight, but tomorrow, or soonest, will merge this using the special `Squash and merge` option offered, which I think will pull in you some 47 commits as one... First time I have tried this, so hope it works out well...

Be back soonest... thanks

And look forward to your further enhancement as a separate issue... thanks...

geoffmcl:
@eric-brechemier seems the `SnM` worked fine... thanks...

eric-brechemier:
@geoffmcl 

I finished opening new issues for the improvements that I wanted to suggest:

* [Generation of the man page should be done in two phases #484](https://github.com/htacg/tidy-html5/issues/484)
* [Format <name> values of -xml-config in the same way as those of -xml-help #491](https://github.com/htacg/tidy-html5/issues/491)
* [Indicate type of value instead of <%s> in <name> values #492](https://github.com/htacg/tidy-html5/issues/492)
* [Format <seealso> values of -xml-config in the same way as <eqconfig> values of -xml-help #493](https://github.com/htacg/tidy-html5/issues/493)
* [Do not include options 'Used internally' in -xml-config output #494](https://github.com/htacg/tidy-html5/issues/494)

PR #480, opened 2017-01-08
Add optional xmlns:xlink attributes as valid to support inline svg

### LINKED ISSUE #478: <svg> proprietary attribute "xmlns:xlink"
Hey!

I'm getting this error  from inline svg elements in WordPress twentyseventeen theme:
```bash
Warning: <svg> proprietary attribute "xmlns:xlink"
```

The error happens because of this element:
```html
<svg style="position: absolute; width: 0; height: 0; overflow: hidden;"
  version="1.1"
  xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink">
... <!-- Insert the element svg nodes here -->
</svg>
```

I think this is false positive because https://validator.w3.org doesn't give any warnings from the same input.

I'm getting this error from quite new build from master commit: fd0ccb2bbf3f907e5425d5849409fbf1558197bc

onnimonni:
Add `xmlns:xlink` support as instructed by @geoffmcl.

I tested this build in debian docker container and the earlier false positive warning about`xmlns:xlink` being proprietary went away.

This fixes #478.

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -808,6 +808,8 @@ typedef enum
 
   TidyAttr_AS,               /**< AS= */
 
+  TidyAttr_XMLNSXLINK,        /**< svg xmls:xlink="url" */
+
   N_TIDY_ATTRIBS              /**< Must be last */
 } TidyAttrId;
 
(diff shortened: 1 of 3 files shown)


onnimonni:
@geoffmcl Are we still good to go with this?

geoffmcl:
Yes, done, thanks...

onnimonni:
Thanks a lot!

PR #488, opened 2017-02-10
Pretty up output of empty script tags.

balthisar:
- No longer break script tags up on two lines if there is content. However
    output is still subject to the `--wrap` behavior.
  - Previous behavior intact if there is content.

I've committed myself to tidying up the pretty-print engine a bit in order to address several feature requests, and this one addresses a piece of low-hanging fruit without being intrusive.

Specifically some users are tying to pass so-call AMP validation, something Google is promoting for fast websites. Current handling of empty `<script>` tags always break them into two lines, e.g.,

~~~
<script ...>
</script>
~~~

This patch ensures they will remain on a single line.

I've not bumped the version number, and despite the commit comment I won't assign a new configuration option for this purely visual change. 

Comments appreciated, else I'll merge it sometime soon (with a version bump).

### DIFF
--- a/src/pprint.c
+++ b/src/pprint.c
@@ -2000,7 +2000,11 @@ void PPrintScriptStyle( TidyDocImpl* doc, uint mode, uint indent, Node *node )
 
     PPrintTag( doc, mode, indent, node );
 
-    TY_(PFlushLineSmart)(doc, indent);
+    /* SCRIPT may have no content such as when loading code via its SRC attribute.
+       In this case we don't want to flush the line, preferring to keep the required
+       closing SCRIPT tag on the same line. */
+    if ( node->content != NULL )
+        TY_(PFlushLineSmart)(doc, indent);
 
     if ( xhtmlOut && node->content != NULL )
     {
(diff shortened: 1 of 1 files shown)


geoffmcl:
@balthisar have quickly tried and tested this, with my `input5\in_488.html` - no attributes, and `input5\in_488-1.html` - with attributes...

Of course, this only effects `<script ...>`, i.e. with attributes, otherwise such blank script tags are dropped anyway, unless the option `--drop-empty-elements no` is added. But your patch puts the `<script></script>` on one line, and looks **good** to me...

One small suggested addition, would be that I note if I have -

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<title>PR #488</title>
<script>
</script>
</head>
<body>
<p>Blank script tags</p>
<script>
</script>
</body>
</html>
```

and run `$ tidy --drop-empty-element no input5\in_488.html` I get a **nasty** newline after the `<title>`, before `<script></script>`. Maybe this should only happen if in tidy so called `classic` mode, i.e. `--vertical-space yes`, `TidyClassicVS`, and **stupidly** if I add this option this vertical space goes away, which seems **opposite** to the intent... but maybe this can be a separate issue...

And then thinking about this more, maybe your `if ( node->content != NULL )` could also include a test for this `classic` mode, which in general means people do **prefer** `more` vertical space, like `if ( ( node->content != NULL ) || TidyClassicVS )`, but this is just a thought... and again could be separate...

As stated, generally agree with it **as is**... please merge and bump version when you like... thanks...

balthisar:
I kept it as is; I plan to re-work the whole pretty printing soon.

PR #497, opened 2017-02-22
Fix leading white spaces trimming

seaburg:
Hi!

`start` - offset in `lexer->lexbuf`
`len` - length of name

As result`len` almost always less than `start`, and leading white spaces is't trimmed

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -4096,7 +4096,7 @@ static tmbstr ParseValue( TidyDocImpl* doc, ctmbstr name,
             while (TY_(IsWhite)(lexer->lexbuf[start+len-1]) && (len > 0))
                 --len;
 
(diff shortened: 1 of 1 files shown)


geoffmcl:
@seaburg thank you for the PR... we always like those...

This certainly looks like a good fix, in that, as you point out, `start` is an offset into the `lexer->lexbuf`, while `len` is the length of the last `name`, and thus `len` could/would almost always by less than `start`, and thus tidy would not enter that `while` leading white space trim ...

And thus while such leading white space would not be trimmed there, it none the less gets trimmed later in CheckAttributes, like CheckUrl, but this has **bad** consequences since there is then a warning - `escaping malformed URI reference` - which is **not** strictly correct... All that happened was the leading space was removed...

So I think this is a good move, and merged your PR... added a comment, and bumped version to 5.3.20... thanks for catching and fixing this...

PR #504, opened 2017-03-02
Clean deprecations

balthisar:
`next` should start to leave the deprecated stuff behind. This cleans things out quite nicely.


balthisar:
This will _break_ LibTidy users who have _ignored_ the deprecation notices for the last 10 years. If this affects you badly enough that patching your own code isn't feasible, please continue to use `master` until you can update your applications.

PR #507, opened 2017-03-05
Message system refactoring and new API

balthisar:
This addresses #409, and should be considered _after_ merging #504 first.

The **lengthy** commit notes aren't put here automatically by Github yet, but they can be read here: eaa9b33

This is a **big** positive change moving forward!


balthisar:
@geoffmcl, have you had a chance to test this? I'm making _great_ progress on our oldest issue, but unfortunately it's intertwined deeply with the changes this PR makes.

I've been running it stably, but mostly on macOS and Ubuntu... thanks!

geoffmcl:
@balthisar, sorry, was off dealing with some other projects that hold my interest...

While I am sure this is "_great_ progress", and really thank you for all the big **work** and **time** involved, but it represents so **many** changes, making a rational review almost **impossible**!

Please, please try to make incremental changes, split it into many commits, but ok, I do understand some changes **are** big, and requires lots and lots of changes, all at one time... 

But **again** I find myself reading through some changes, while it might start with an important change, like `+    TidyReportLevel_first = 350,`, and also adds some **new** enums, but why all the just space/indent changes? This just make reviewing very, **very** difficult... no, impossible...

But ok, in checking out this branch, `message_system`, is does **not** compile in Windows 10 64-bit, using msvc140.x64!

One of the immediate things is the use of `WCHAR`, if `WIN32`! Now Tidy has **never** supported `WCHAR`, utf-16 like. It seems the only other place it is used is in my SPRTF modules. but please note it is under an `#ifdef UNICODE` switch! And this was **never** intended to be turned **on**, and I suspect would **not** work anyway... 

I personally think `WCHAR` should be totally removed, even in my SPRTF modules... It simply has no place in a utf-8 world!

The next thing is the use of `int32_t`, `uint32_t`, `int64_t`, and `uint64_t` in the `PrintfArg` structure union, in messageobj.c. Yes, these can be supported, in Windows, with `#include <stdint.h>`, but are they really necessary?

Presently, message formats can include `%d`, or `%u`, but will it ever be necessary to support 64-bit values? That is output value over 2 million? Maybe... but **much** more work has to be done for this...

And that would also go for supporting a `double`, or pointers! Yes, maybe at some future time, perhaps, but that really gets into trying to support things format items like `%p`, or `%I64u`, ... in the format string... a giant **can of worms**...

Yes, there is nothing wrong with going for **everything** is supported, but could that not be left to some future date, when the need arises... Just an idea...

So even if I add the above `include`, I still see **several** warnings, of the type -

```
1>D:\Tidy\tidy-html5\src\messageobj.c(387): warning C4311: 'type cast': pointer truncation from 'TidyMessageArgument' to 'int'
```

Maybe these are benign, and could be suppressed, if they are not **really** a problem...

So, in the end I can get it to compile with the following patch -

```diff
diff --git a/src/messageobj.c b/src/messageobj.c
index 1c7b40f..f77006a 100644
--- a/src/messageobj.c
+++ b/src/messageobj.c
@@ -13,7 +13,9 @@
 #if !defined(NDEBUG) && defined(_MSC_VER)
 #include "sprtf.h"
 #endif
-
+#ifdef _WIN32
+#include <stdint.h>
+#endif
 
 /*********************************************************************
  * BuildArgArray Support - declarations and forward declarations
@@ -44,7 +46,7 @@ struct printfArg {
         double d;
         const char *s;
         size_t *ip;
-#ifdef WIN32
+#if (defined(_WIN32) && defined(UNICODE))
         const WCHAR *ws;
 #endif
     } u;
@@ -718,7 +720,7 @@ static struct printfArg* BuildArgArray( TidyDocImpl *doc, ctmbstr fmt, va_list a
                 break;
                 
             case 'S':
-#ifdef WIN32
+#if (defined(_WIN32) && defined(UNICODE))
                 nas[ cn ].type = TYPE_WSTRING;
                 break;
 #endif
@@ -815,7 +817,7 @@ static struct printfArg* BuildArgArray( TidyDocImpl *doc, ctmbstr fmt, va_list a
                 nas[cn].u.s = va_arg( ap, char* );
                 break;
                 
-#ifdef WIN32
+#if (defined(_WIN32) && defined(UNICODE))
             case tidyFormatType_WSTRING:
                 nas[cn].u.ws = va_arg( ap, WCHAR* );
                 break;
```

Have **not** yet had a chance to run, and test the results, but, as asked, really do we need all the additions that cause this?

Regards, Geoff.

OT: In other projects, some Mac, and linux/unix, developers have been able to install a Windows 10, maybe they had to buy a licence, with msvc140 (free) and at least test a compile before pushing forward changes... Just an idea!

balthisar:
Hi @geoffmcl,

Thanks for having a look. I completely understand "other projects," and I get your point about small commits, which just wasn't possible in this case. For white space, I configure diff to ignore it. In a lot of cases, the white space changes were simply made to match the style of the rest of Tidy's code.

I _really_ hated to include WCHAR support, and so this is one of the chief reasons I like to have a Windows developer take a look. I _do_ have gcc and MSVC on a Win10 installation at home (my preference, however, is influence by being forced to use Windows all day at work). Having those does _not_ make me a Windows developer with specialized knowledge about the Windows operating system, thus your comments on the Windows platform are **especially** important.

Likewise supporting larger ints and much of the rest of the printf formats was a big internal debate. Add them later? Keep them now? Personally I favor _dropping_ them, but I left them in because these decisions can't be mine alone. It sounds like you concur with the lack of need, and so it's absolutely no skin off my back to simplify these. By way of example, you can see that I didn't even bother to specify an API for doubles and pointers!

Rather than your patch above to accommodate WCHAR, I'd propose simply cutting it out. If you, as a Windows dev say that it's not needed in Windows, then by all means, I'm ready to yank out platform specifics in a heartbeat!

Thanks for the feedback.

geoffmcl:
@balthisar seems we agree ;=))

Yes, I do often configure diff to ignore white space changes, and in this case that removed about 11K from a 144K diff, but a 133K diff is still a lot to take in, but, as stated, also undertand sometimes it is **needed**... 

Ok, let's kick out the WCHAR from Tidy. While it can **not** be said that is is **not** needed in Windows - the whole OS **is** using WCHAR! - but thankfully the parts of the OS used by tidy are presently taken care of in the system API, so for the moment we can continue to use `char *` in a fully portable way...

In general, Tidy has avoided including `<windows.h>` in most modules, and would like to keep it that way for as long as possible. And even in the `mappedio.c` case, which has to use the native Windows API, for memory mapping, we chose to specifically use `CreateFileA`, and thus use the common `ctmbstr filnam`... and would have preferred it not be included in `tidy.c` either, for the console change, but that's another story...

In another big unix app, someone decided they just **had** to use `wchar_t` for Windows, and wow, you should see the squiggly, ugly, switched, code that resulted, and it was weeks, months to get it near right! So, agreed, **no** WCHAR for Tidy, for the moment...

Not sure where you had `a big internal debate` over printf formats. I would be 100% for Tidy only needs to support what we use **now**. And given that we deal with html, I can not see us even needing `int` **and** `uint`. Since we would not been dealing with over 2 million lines, they are the same... but, ok, no problem...

So, yes let's remove WCHAR and larger ints, doubles and pointers, and **none** of my patch would be required! 

That I would prefer...

balthisar:
@geoffmcl, thanks for the continued feedback. I am _very_ happy with the cleanup by eliminating all of the types. The internal debate was mostly centred about people adding strings and using new types for whatever reason. The current code is much cleaner.

Still...

> And given that we deal with html, I can not see us even needing `int` and `uint`.

There are exactly _four_ uses of `%d` in the strings, and I know for a fact that they predate my involvement! ;-)

I went ahead and left doubles in there, as they're not causing any harm. Maybe we'll add benchmarking or something in the future, and want to report that Tidy took 0.0025 seconds to complete (although I'd use milliseconds myself...).

Also I spent a couple of hours cleaning up my Windows environment, and things look good on Windows now, too.

geoffmcl:
Yes, your cleaned branch compiled in Windows, and I have had a chance to run some small tests, and all looked fine...

Can not see any reason why this is not merged to **development** `next`... thanks...

balthisar:
Hmmm... not sure why my command line merge didn't close this automatically, so I guess I'll close it myself.

Thanks for the check, @geoffmcl.

Issue #119, opened 2014-07-19
Support custom elements


Moult:
Spec is here: http://www.w3.org/TR/custom-elements/

Currently, custom elements throw errors.

geoffmcl:
Please note the current development version is in develop-500
$ git checkout develop-500
At this time issue must only be against this version...

But tidy has config items -
new-blocklevel-tags         Tag names  tagX, tagY, ...                         
new-empty-tags              Tag names  tagX, tagY, ...                         
new-inline-tags             Tag names  tagX, tagY, ...                         
new-pre-tags                Tag names  tagX, tagY, ...
where you can add custom elements. Is this what you mean?

If this does not work as expected in the develop-500 branch, then look forward to further comment, especially with a sample html input, the configuration used, the tidy output, html and error/warning text, and the output expected...

Maybe I misunderstand the feature requested, in which case again explain, preferably with sample html usage...

balthisar:
> Abstract
>This is a work in progress! This specification is for review and not for implementation! For the latest updates, including important bug fixes, please look at the draft on GitHub instead.
>This specification describes the method for enabling the author to define and use new types of DOM elements in a document.

As such, this should on the back burner for the time being.

yoshuawuyts:
I'm using `custom-element` now (easily enabled through [`webcomponents.js`](https://github.com/webcomponents/webcomponentsjs), and it throws a "proprietary attribute" error when I'm using custom elements. 

I understand that having full custom-element support might take quite some work, but maybe a good first step would be allowing `is=""` properties to be set on all elements? E.g. I'm trying to do this:
```html
<button is="custom-button"></button>
```
And the error is:
```txt
<button> proprietary attribute "is"
```

bendavis78:
Tidy still needs an option to disable checking for invalid tags, instead of having to pass every possible custom tag as an argument. When working on projects that involve a lot of custom elements (eg, Polymer), it becomes very difficult to construct the command line to check your file.

geoffmcl:
@bendavis78 feel free to code a new feature to be added to **libTidy** to fully disable invalid tag and attributes checking...

However, it does feels a little contrary to what tidy is about, but such a feature addition would be considered...

And remember such custom element additions can be added to a config file, thus removing the difficulty in constructing a command line...

Will close this for now, since there have been no further comments for a few months...

But as always, feel free to re-open, or open new issues...

geoffmcl:
[...]

PR #513, opened 2017-03-11
Custom Tags

balthisar:
Supporting #119, this PR addressed the issue of custom elements and the `is` attribute.

Don't try to read this PR until #507 is merged, as #507 is pretty massive. This PR _will not_ merge cleanly into next by itself.

Please squash before merging to avoid all of the WIP commits.


balthisar:
See #514 instead.

PR #516, opened 2017-03-23
Cpack test

geoffmcl:
Added ability to add system DLLs to packages, but it seems to miss some...

Add an option to build the MSVC static runtime, /MT, version.


balthisar:
It builds and runs, but all of my Windows installations have the runtimes installed, so I can't actually test that the libraries are truly statically linked, so I'm okay with this PR.

My only question is why the ugly license from the tidy.h header with the stupid Doxygen comments? The older .md file was a lot more readable.

geoffmcl:
@balthisar Thanks for testing... 

There are a few clues that the libraries **are** statically linked. The first is the size... I see at least 150K added. But the best is to run `dumpbin /ALL tidy.exe > temp.txt`, and search for **imports:**. There you will see a list for the `/MD` version, and just one for the `/MT` version... You need to be in a MSVC prompt to run `dumpbin`...

I have my own [dump4](https://github.com/geoffmcl/dump4), but it is broken at the moment for the 64-bit build... It also list the `imported` DLLs, when it is working ;=))

> why the ugly license from the tidy.h header with the stupid Doxygen comments? 

Initially just needed a file to test - one of the windows binary builders insists it be a `.txt` file - then forgot about it...

Maybe it could be a straight copy of `LICENSE.md`, or maybe delete what I hastily added, and rename that to `.txt`?

What do you think?

balthisar:
> Maybe it could be a straight copy of LICENSE.md, or maybe delete what I hastily added, and rename that to .txt?

Either or. I've been spending a _lot_ of time with Doxygen lately, and the ugly syntax is driving me crazy. I've got to say, though, that it generates pretty websites.

AlbertWiersch:
If you test a build with virustotal.com and go to the 'File detail' tab, it will show the 'PE imports'. The one that is statically linked only shows 'KERNEL32.dll' whereas the others show a lot of other required DLLs.

geoffmcl:
Yes, thanks for that... I just did a 5.5.12 `/MT` build of tidy.exe, and unfortunately it still show `Qihoo-360`... are you a member there... can you leave a comment that this appears to be a false positive?

AlbertWiersch:
Sorry, I'm not a member there. I would just ignore it... with all those virus checkers and constant changes and updates, false positives happen rather often it seems. If you still want to contact someone then I think you need to contact Qihoo directly. Most anti-malware companies should have a way to submit false positives.

PR #529, opened 2017-04-15
Fix behavior of <button> elements in HTML 5

### LINKED ISSUE #461: Feature request: add an option to allow inline elements inside a span
current behaviour is that:

```
<span><button>OK</button></span>
```

throws:

```
Warning: inserting implicit <span>
Warning: replacing unexpected button with </button>
Warning: missing </button>
```

and the final result is:

```
<span><button><span>OK</span></button></span>
```

It might be handy to have an option to prevent this behaviour and let the original markup unchanged as it is valid and it's a part of quite common pattern.

lhchavez:
HTML 5 restricts the children of <button> elements to only phrasing
content. HTML 4 also allowed block elements.

In particular, this change allows code such as

    <span><button>OK</button></span>

to be left as-is, since the <span> does not need duplication for
crossing a block boundary. Fixes #461.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -1442,6 +1442,22 @@ void TY_(ParseBlock)( TidyDocImpl* doc, Node *element, GetTokenMode mode)
 #endif
 }
 
+/*
+<button> changed in HTML 5. It now contains phrasing content only, rather than
+HTML4-era flow elements, which also allows for blocks.
+*/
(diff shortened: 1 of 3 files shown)


lhchavez:
After a bit more though, I think I'd rather fix #461 more generally. I still want to submit this, but I'll rename the branch to something more appropriate.

PR #530, opened 2017-04-15
Add a flag to avoid propagating inline elements across blocks

lhchavez:
This change adds the InsertInlineTags flag (default=on) that propagates
inline elements across block boundaries. This causes

    <span><div>OK</div></span>

to become

    <span><div><span>OK</span></div></span>

This fixes issue #461.

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -580,6 +580,7 @@ typedef enum
     TidyIndentContent,           /**< Indent content of appropriate tags */
     TidyIndentSpaces,            /**< Indentation n spaces/tabs */
     TidyInlineTags,              /**< Declared inline tags */
+    TidyInsertInlineTags,        /**< Insert injected inline tags into block tags */
     TidyJoinClasses,             /**< Join multiple class attributes */
     TidyJoinStyles,              /**< Join multiple style attributes */
     TidyKeepFileTimes,           /**< If yes last modied time is preserved */
--- a/src/config.c
+++ b/src/config.c
@@ -276,7 +276,7 @@ static const TidyOptionImpl option_defs[] =
     { TidyIndentContent,           PP, "indent",                      IN, TidyNoState,     ParseAutoBool,     autoBoolPicks   },
     { TidyIndentSpaces,            PP, "indent-spaces",               IN, 2,               ParseInt,          NULL            },
     { TidyInlineTags,              MU, "new-inline-tags",             ST, 0,               ParseTagNames,     NULL            },
-    { TidyInsertInlineTags,        MU, "insert-inline-tags",          BL, no,              ParseBool,         boolPicks       },
+    { TidyInsertInlineTags,        MU, "insert-inline-tags",          BL, yes,             ParseBool,         boolPicks       },
     { TidyJoinClasses,             MU, "join-classes",                BL, no,              ParseBool,         boolPicks       },
     { TidyJoinStyles,              MU, "join-styles",                 BL, yes,             ParseBool,         boolPicks       },
     { TidyKeepFileTimes,           MS, "keep-time",                   BL, no,              ParseBool,         boolPicks       },
--- a/src/config.c
+++ b/src/config.c
@@ -276,6 +276,7 @@ static const TidyOptionImpl option_defs[] =
     { TidyIndentContent,           PP, "indent",                      IN, TidyNoState,     ParseAutoBool,     autoBoolPicks   },
     { TidyIndentSpaces,            PP, "indent-spaces",               IN, 2,               ParseInt,          NULL            },
     { TidyInlineTags,              MU, "new-inline-tags",             ST, 0,               ParseTagNames,     NULL            },
+    { TidyInsertInlineTags,        MU, "insert-inline-tags",          BL, no,              ParseBool,         boolPicks       },
     { TidyJoinClasses,             MU, "join-classes",                BL, no,              ParseBool,         boolPicks       },
     { TidyJoinStyles,              MU, "join-styles",                 BL, yes,             ParseBool,         boolPicks       },
     { TidyKeepFileTimes,           MS, "keep-time",                   BL, no,              ParseBool,         boolPicks       },
--- a/src/istack.c
+++ b/src/istack.c
@@ -215,6 +215,11 @@ Bool TY_(IsPushedLast)( TidyDocImpl* doc, Node *element, Node *node )
 */
 int TY_(InlineDup)( TidyDocImpl* doc, Node* node )
 {
+    if (!cfgBool(doc, TidyInsertInlineTags))
+    {
+        return 0;
+    }
+
     Lexer* lexer = doc->lexer;
     int n;
 
--- a/src/language_en.h
+++ b/src/language_en.h
@@ -679,6 +679,21 @@ static languageDefinition language_en = { whichPluralForm_en, {
       - It's very important that <br/> be self-closing!
       - The strings "Tidy" and "HTML Tidy" are the program name and must not
       be translated. */
+        TidyInsertInlineTags,               0,
+        "This option enables inserting \"missing\" inline elements around the "
+        "contents of blocklevel elements such as <code>P</code>, "
+        "<code>DIV</code>, etc. For instance, this causes "
+        "<code><span><div>OK</div></span></code> to become "
+        "<code><span><div><span>OK</span></div></span></code>."
+    },
+    {/* Important notes for translators:
+      - Use only <code></code>, <var></var>, <em></em>, <strong></strong>, and
+      <br/>.
+      - Entities, tags, attributes, etc., should be enclosed in <code></code>.
+      - Option values should be enclosed in <var></var>.
+      - It's very important that <br/> be self-closing!
+      - The strings "Tidy" and "HTML Tidy" are the program name and must not
+      be translated. */
         TidyJoinClasses,              0,
         "This option specifies if Tidy should combine class names to generate "
         "a single, new class name if multiple class assignments are detected on "
--- a/src/parser.c
+++ b/src/parser.c
@@ -2157,7 +2157,8 @@ void TY_(ParseInline)( TidyDocImpl* doc, Node *element, GetTokenMode mode )
          *  big consequences.
          *  There may be other exceptions to be added...
         \*/
-        if (!(node->tag->model & CM_INLINE) &&
+        if (cfgBool(doc, TidyInsertInlineTags) &&
+            !(node->tag->model & CM_INLINE) &&
             !(element->tag->model & CM_MIXED) &&
             !(nodeIsSPAN(element) && nodeIsMETA(node)) )
         {


lhchavez:
PTAL! I tested this against https://github.com/geoffmcl/tidy-test/commit/4f6089bcb09c087b28a31b1e6353d2d6a9aa3ceb and passes, and against https://github.com/htacg/tidy-html5-tests , where it needs to add the `insert-inline-tags` config to case-1015959.conf in order to pass completely.

geoffmcl:
@lhchavez sorry for the brief delay in replying on this...

Although I have not yet tested it, it looks good eye-balling the changes...

Unless someone has a negative comment, I will merge this in a few days... thanks...

PS: The same for #531 and #532... thanks...

balthisar:
Test successful on macOS.

geoffmcl:
@lhchavez ok, have now had a chance to pull your fork, branch `add-insert-inline-tag`, build and test...

It certainly `passes` using my [in_461.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_461.html), in that, now by default, tidy will not propagate the `<span>`, unless I add `--insert-inline-tags yes` to the config... so this is **good**...

But I guess I am **confused** with the current state of the **regression** [tests](https://github.com/htacg/tidy-html5-tests)! Using the `next` branch, I get problems on much more than case 1015959! And there is no `case-1015959.conf`. That case uses `config_default.conf`, but even adding `insert-inline-tags: yes` to it does **not** clear up the several problems...

@balthisar, you say "Test successful on macOS", but is that the simple `in_461.html` test, or the full 232 regression tests... maybe I should be using some other tests branch? Or something...

Maybe we should, initially at least, default the `insert-inline-tags` option to **yes**, to keep the current behavior? Just for a certain period of time... or something...

Anyway, seek more comments, feedback before merging this... thanks...

lhchavez:
hmm interesting. let me try running that again AND sending the patch to the regression tests, which I should have done from the beginning instead of just mentioning it in the PR. Sorry about that.

lhchavez:
All right, checking out the `next` branch plus this PR, building it and running it against https://github.com/htacg/tidy-html5-tests/pull/13 passes everything on my machine. Let me know if there's anything else that's missing and thank you for your time :D

geoffmcl:
@lhchavez ok, I can see you have added a `case-1015959.conf`, which is just a copy of the `default` with `insert-inline-tags: yes` added... thanks, and that gets through the **first** part of the tests, which just checks tidy's exit value...

But I am still getting differences when doing the `diff -u ../cases/testbase-expects ../cases/testbase--results`, which is the **second** part of the tests, but maybe I am doing something really wrong... 

There is a diff on cases 1050673, 1062511, 1068087, 1410061b, 1410061c, 1423252, 1638062, 1773932, 427838, and quite a few more... 

Naturally, I do **not** get these diffs when using `next` Tidy 5.5.14... Of course, leaving the new addition `conf` file there, I do get an `expected` diff, since 5.5.14 does not know about this option... of -

```diff
diff -u ..\cases\testbase-expects\case-1015959.txt ..\cases\testbase-results\case-1015959.txt
--- ..\cases\testbase-expects\case-1015959.txt	Sat Apr 29 18:19:59 2017
+++ ..\cases\testbase-results\case-1015959.txt	Sat Apr 29 19:17:25 2017
@@ -1,3 +1,4 @@
+Config: unknown option: insert-inline-tags
 line 7 column 1 - Warning: missing </font> before <p>
 line 8 column 4 - Warning: inserting implicit <font>
 line 9 column 4 - Warning: inserting implicit <font>
```

Taking just case **1050673**... it is a simple html **mess** - 

```html
<HEAD>
<TITLE>
<MARQUEE>
</BASEFONT>
</BODY>
<B>
<BASE>
<HR>
```

Using 5.5.14 I get the output - and ignoring the different warnings issued for now -

```html
<!DOCTYPE html>
<html>
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 for Windows version 5.5.14">
<title></title>
<base>
</head>
<body>
<hr>
</body>
</html>
```

When I use 5.5.14.I461 I get a completely different output -

```html
<!DOCTYPE html>
<html>
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 for Windows version 5.5.14.I461">
<title></title>
</head>
<body>
<b>
<base>
<hr></b>
</body>
</html>
```

Quite a lot of **striking** differences... `<base>` not in `<head>`, `<b>` retained...

Note, to get that **additional** version numbering, you can add `-DTIDY_RC_NUMBER=I461` to the cmake command line... this is how I try to make sure I **am** running the modified version of tidy... 

This was all using a **Windows** build... maybe I have done something really **crazy**... tomorrow, or soonest, I will re-run the build and test in linux... sorry for the delay...

lhchavez:
ah! i thought `./runalltests.sh` ran everything :P okay, let me go back and investigate the failures.

geoffmcl:
@lhchavez yes, sorry, maybe we do **not** make it clear enough that the tests are at **two** levels... will try to look at the docs on this... 

Using the Windows tools-cmd bat files, I do do this diff, and report a problem if diff does not exit zero... maybe this 2nd diff part needs to be added to runalltests.sh also?

PS: Of course in testing 5.5.14.I461 using the option `--insert-inline-tags yes`, I do get back close to what is expected -

```html
<!DOCTYPE html>
<html>
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 for Windows version 5.5.14.I461">
<title></title>
<base>
</head>
<body>
<hr>
</body>
</html>
```

Still to **check** this, and the warnings closely for an **exact** match...

As mentioned earlier, maybe the default should be `yes` at this time... but maybe there is also more to understand here... maybe need to add more checking to the PR... unfortunately out of time tonight... be back soonest...

lhchavez:
All right, inspected a few of the failures and it turns out that having this flag off by default produces HTML that is incorrect in edge cases :( I agree that the default should be `yes` because it's probably more correct to deal with the edge cases on a case-by-case basis like I did for https://github.com/htacg/tidy-html5/pull/531

This now passes with tip-of-tree https://github.com/htacg/tidy-html5-tests (and I made sure to inspect the diffs this time), so PTAAL!

geoffmcl:
@lhchavez back briefly after dinner...

Wow, really sorry you are feeling **frustrated** with trying to add a **new** option to tidy... I can say I also feel that many, **many** times...

Not sure what you describe as **edge** cases... Tidy tries very hard to deal with whatever it is given... whatever the html **mess** is... **edge** or not... and tries to produce **valid** html output, regardless...

Ok, you have flipped the default to **yes**... but have you fully described to the **user** what is likely to happen if they set this option to **no**? What are the **full** consequences... I agree your `help` text describes one case, but it seems there are others... quite difficult to describe...

Please do **not** get me wrong... I seem to support, encourage, such an option... it does do some things I may want, as in my simple `in_461.html` case... I only suggest maybe more could be done to protect against what you call **edge** cases...

Like in `src/istack.c`, should we block **every** `InlineDup` just because this option is `no`? Or in `src/parser.c`, where there is already a **big** comment about `block level tags end this element... big consequences...`, should it also be only on this option is `no`?

Now, I do **not** know yet the full answers to these questions! But am willing to **explore**!

Although I have still to **test** #531, there seems a good case for having a separate parser function for `<button>`... and will get around to testing this...

And again, thanks for the PR 14 on [tidy-tests](https://github.com/htacg/tidy-html5-tests/pull/14) documentation... this looks good... 

I have still to run the `tests` in linux, to see how we can clearly describe what the **regression** tests do... 

 - stage 1 - just tidy exit value... 
 - stage 2 - diff of the result to expected... 
 
but on reading this feel sure it helps... and assume this negates the previous [PR 13](https://github.com/htacg/tidy-html5-tests/pull/13)...

So again **very sorry** this feels like such a **frustrating** process... 

While in some ways HTML Tidy is **simple**, in other ways it can be quite **complex**... developed over a **long** period of time, by many participants... 

**Please continue to try to help**... thanks...

geoffmcl:
@lhchavez now that we are specifically addressing the `<button>` issue #461, with your, or my alternative, branch `issue-461` patch, is this **new** option **really** required?

**Are there other cases, like your `<div>` sample, where this is required, desired, beneficial?**

Yes, in a way, we understand browsers, in general, accumulate, set, display type attributes, until the next DOM change, and do not close, revert, such attributes, on the next `block` element, but should Tidy also do this?

And as demonstrated, this can have **dire** consequences... at least for **legacy** doctypes, and certain html snippets...

You have now defaulted this option to `yes`, so it only influences those that add a **no** to the config...

Even if this option is **accepted**, then at the very least, I would like the **help** text expanded to -

```
        TidyInsertInlineTags,               0,
        "This option enables inserting \"missing\" inline elements around the\n"
        "contents of blocklevel elements such as <code>P</code>,\n"
        "<code>DIV</code>, etc. For instance, this causes\n"
        "<code><span><div>OK</div></span></code> to become\n"
        "<code><span><div><span>OK</span></div></span></code>.\n"
        "Experimental: Warning - Setting this option to <code>no</code>\n"
        "can have \"dire\" consequences on some html documents!\n"
        "It can result in Tidy outputting \"invalid\" html!\n"
        "Use at your own risk!\n"
```

Or something like that! It explains the **risk**...

Simply, abandoning the current default behavior of Tidy **can** produce **invalid** html... like case **1050673**...

And now that we have identified, and have fixes for the `<button>` case, should **not** each such `block` element be dealt with individually, rather than a **gross** abandonment of this behavior?

Maybe I am really **out of order** here...

As always, just seek comments, feedback, ideas... thanks...

lhchavez:
Ah sorry that I hadn't replied. I agree with you in general, but there is one use case I think this still *might* be useful: templating languages. If we declare a custom tag it has to be either blocklevel or inline, and whichever one you choose will cause Tidy to insert extraneous closing tags in some places.

The one I'm interested in is the one from [Vue](https://vuejs.org/v2/guide/single-file-components.html), but since they use `<template>...</template>`, and (after just going through the spec one more time) IIUC that one's content model includes almost every other tag and can be added mostly anywhere, I might have run into another bug. I'll probably re-try whatever I was doing and document the exact failures instead of adding workarounds.

Having said that, I'm fine with either adding this flag with the warnings you suggested, or adding a new kind of custom tag that can be placed *anywhere* and supports adding *any* element without Tidy trying to guess whether tags should be closed when the boundary is straddled (which is mostly the same as adding this flag, but containing its effect within a small scope of the file). Regardless, I'll try to send patches to fix any issues I see specifically with Vue :)

geoffmcl:
@lhchavez given that you agree with the additional **warning**, if you can provide some **use** cases, samples, and show how it can be helpful, maybe there is an argument for this option... just concerned that it can also mangle a document... but not entirely against it...

But it seems your initial example is **not** exactly correct. The current, normal propagation of Tidy will **close** the `<span>` on reaching the `<div>`, and then reopen it, or inline-duplicate it, after the `<div>`, and then close it before the end `</div>`. And, unless you add `--drop-empty-elements no`, will then drop the leading empty `<span>`...

In other words, `<span><div>OK</div></span>` will become `<div><span>OK</span></div>`, and `<span><p>OK</p></span>` will become `<p><span>OK</span></p>`, etc...

Accordingly, below is the corrected text, plus the suggested warning, now conforming to the text template -

```
        TidyInsertInlineTags,               0,
        "This option enables inserting \"missing\" inline elements around the "
        "contents of blocklevel elements such as <code>P</code>, "
        "<code>DIV</code>, etc. For instance, this causes "
        "<code><span><div>OK</div></span></code> to become "
        "<code><div><span>OK</span></div></code>."
        "<br/>"
        "Experimental: Warning - Setting this option to <var>no</var> "
        "can have \"dire\" consequences on some html documents! "
        "It can result in Tidy outputting \"invalid\" html! "
        "Use at your own risk!"
```

But now we agree it **must** default to `yes`, maybe this could be completely reworded, in light of the above - sort of making it clearer that this has always been Tidy's **normal** behavior, and this option allows that to be **reversed**...

Maybe you could add that text, or a modified version, to the PR... if you agree...

What do others think? Would certainly appreciate some **more** feedback on this option. For sure, I am still slightly on the negative side, but can be swayed by **use** cases, where it helps, with samples, and helpful comments... thanks...

PS: A `+1` will not cut it here ;=))

PR #531, opened 2017-04-15
Fix behavior of <button> elements in HTML 5

lhchavez:
HTML 5 restricts the children of <button> elements to only phrasing
content. HTML 4 also allowed block elements.

In particular, this change allows code such as

    <span><button>OK</button></span>

to be left as-is, since the <span> does not need duplication for
crossing a block boundary. Mitigates #461.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -1442,6 +1442,22 @@ void TY_(ParseBlock)( TidyDocImpl* doc, Node *element, GetTokenMode mode)
 #endif
 }
 
+/*
+<button> changed in HTML 5. It now contains phrasing content only, rather than
+HTML4-era flow elements, which also allows for blocks.
+*/
+void TY_(ParseButton)( TidyDocImpl* doc, Node *element, GetTokenMode mode)
+{
+    if (TY_(HTMLVersion)(doc) == HT50 || TY_(HTMLVersion)(doc) == XH50)
+    {
+        TY_(ParseInline)(doc, element, mode);
+    }
+    else
+    {
+        TY_(ParseBlock)(doc, element, mode);
+    }
+}
+
 /* [i_a] svg / math */
 
 struct MatchingDescendantData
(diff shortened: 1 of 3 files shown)


lhchavez:
PTAL! This was tested against https://github.com/geoffmcl/tidy-test/commit/4f6089bcb09c087b28a31b1e6353d2d6a9aa3ceb and passes.

geoffmcl:
As replied to #530...

balthisar:
Confirmed as working on macOS.

geoffmcl:
@lhchavez sorry for the delay on this...

I was working on an alternate patch, to put this change of the parser for `<buttom>` in `TY_(ResetTags)` - reset to html5, and `TY_(AdjustTags)` - set to html4 mode, if a `legacy` doctype is found. As you will see this is where **several** other such mode swithes are handled...

And have pushed my patch to the `issue-461` branch for testing...

This would keep **all** the `switches` between html5 and htm4 mode in **one** place, rather than creating a **new** parser just for `<button>`, which does the same thing - switches between `ParseInline` and `ParseBlock`...

But ran into `regression` testing problems, in that my **fix** caused quite a number of **diffs**, and some exit value changes... Arrgh!!!

On investigating this, it seems some **errors** have crept into the `regression` tests!

Now most of the `regression` tests are still **legacy** documents. Take say 1086083. It is `HTML 4.01`, and testcase contains the code `<li>Huh?</li>`, and tidy is supposed to **fix** this by adding a `<ul>`, which, if we back up the tidy-tests repo is what we will find in the original `out_1086083.html`, but **not** in the current `expects/case-1086083.html`!

In other words, this `expects/case-1086083.html` now **fails** validation with the error `Line 16, Column 6: document type does not allow element "LI" here; missing one of "UL", "OL" start-tag`! While the original `out_1086083.html` passes!!

That is we now expect, allow some html5 tolerance in a legacy document, and this is an error! And there seems to be several other similar cases... yet to be explored fully...

Now this has nothing to do with `<button>`, or #461, and may in fact merge this, as is, **until** I can find time to investigate and correct the current `expects`, in a separate issue... please bear with me... thanks...

geoffmcl:
PS: Now opened [Issue 15](https://github.com/htacg/tidy-html5-tests/issues/15)... 

Any help appreciated... thanks...

geoffmcl:
@lhchavez as agreed have merged #540 and closed #461, so closing this...

Issue #354, opened 2016-01-25
Publish tidy-html5 to a Debian repo


mcandre:

geoffmcl:
@mcandre as some background I would add -

Both I and @balthisar have added to an old bug report - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=655200#70

Maybe a current Debian user could file a **new** bug? Add their voice to ours...

For Ubuntu I am holding open a question - https://answers.launchpad.net/ubuntu/+source/tidy/+question/276623 - Maybe other Ubuntu users could add to this, or add something new?

As first I thought it may be the holiday season that delayed things, but we are now nearly at the end of January 2016 and still no results!!!

It is really ugly that the tidy offered is a 2009 cvs source release... or have I missed something here?

**What else can we do?**

geoffmcl:
Hi Manfred,

Re: Update HTML Tidy package

Thank you for help searching around for a solution to this... you
are a great help...

In fact your search link leads to an email from Jason, clearly
saying he no longer has time to maintain HTML Tidy, dated 2011-06-01
01:05:31 GMT
http://article.gmane.org/gmane.linux.debian.devel.general/162695/match=

So am also sending this to the MIA team -

Dear MIA Team,

Hi, I hope you can help HTML Tidy.

Back in 2011-06-01 Jason advised he no longer has time to maintain
HTML Tidy. Manfred's search reveals his last action was in 2013 -
http://search.gmane.org/?query=&author=Jason+Thomas&group=gmane.linux.debian.*&sort=date

I do not know your process in choosing a new maintainer, but note
for example that Daniel offers his help, twice -
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=655200#65
and
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=770129#20

Hope you can help get HTML Tidy updated, when you get a chance.

Last release: https://github.com/htacg/tidy-html5/releases/tag/5.1.25

Regards,
Geoff.

geoffmcl:
On Mon, Jan 25, 2016 at 05:38:37PM +0100, Geoff McLane wrote:
> Hi Manfred,
> 
> Re: Update HTML Tidy package
> 
> Thank you for help searching around for a solution to this... you
> are a great help...
> 
> In fact your search link leads to an email from Jason, clearly
> saying he no longer has time to maintain HTML Tidy, dated 2011-06-01
> 01:05:31 GMT
[...]

PR #534, opened 2017-04-17
Generate pkgconfig file for tidy

spk:
This allow to generate pkgconfig file from cmake for tidy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@
 cmake_minimum_required (VERSION 2.8.7)
 
 set(LIB_NAME tidy)
+set(LIBTIDY_DESCRIPTION "${LIB_NAME} - HTML syntax checker")
+set(LIBTIDY_URL "http://www.html-tidy.org")
 
 project (${LIB_NAME})
 
@@ -371,7 +373,7 @@ else ()
 endif ()
 
 set(CPACK_PACKAGE_NAME "${LIB_NAME}")
-set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "${LIB_NAME} - HTML syntax checker")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "${LIBTIDY_DESCRIPTION}")
 set(CPACK_PACKAGE_VENDOR "HTML Tidy Advocacy Community Group")
 set(CPACK_PACKAGE_CONTACT "maintainer@htacg.org")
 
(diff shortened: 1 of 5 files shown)


geoffmcl:
@spk thanks for this PR... we like those...

But please remind me why we **need** a pkgconfig file... Working mainly in Windows, it seems this is not required, or used there... but...

And it seems you have only added it to the CPack package config... that is if a DEB/RPM/... is generated. What about a normal build and install from source?

However, it also seems no harm, so if no negative comments, will merge this in a few days... thanks...

spk:
Hi @geoffmcl thanks for the feedback
The pkgconfig file is useful for people or tools that need to find information about the libs or headers; like the path of the `tidy.h` on the system with for example `pkg-config --cflags tidy` give for me `-I/usr/local/include`
I found an issue with the Debian package with your information; currently going to `/usr/local` where for Debian its a violation of the policy, I will make a fix soon

balthisar:
Seems okay by me.

spk:
@geoffmcl @balthisar I found a fix with pkgconfig with packages in 3333ca7 let me know what you think?

geoffmcl:
@spk thank you for the update...

However, as a mainly Windows person, where this `pkgconfig` change is **never** used, I think?, then I am **not** the person to decide on this...

One comment against `currently going to "/usr/local" where for Debian its a violation of the policy`, when working with the **last** Debian update - a very **long** and **arduous** process - I got the **opposite** impression... but what do I know...

See https://answers.launchpad.net/ubuntu/+source/tidy/+question/276623, and https://launchpad.net/ubuntu/+source/tidy-html5 - added Tidy 5.2.0...

**Especially** see Issue #354, and maybe others, for more on this **Debian** process... with @AdamMajer, @danielhjames, @jasonethomas, @mcandre, and we urge them to update to release Tidy 5.4.0... when they get the chance...

And maybe during the **Debian** package creation, this was already done... I do not know...

For sure, CMake defaults to `/usr/local` if **no** `-DCMAKE_INSTALL_PREFIX=<path>` is given... that is **known**...

Simply, if this is ok by @balthisar, in OS X, a *nix type OS, then for sure I have no problem with this... thanks...

balthisar:
@geoffmcl, @spk, from the Mac perspective but _not_ as a pkg-config user, I'm okay with this. The default `make install` isn't broken, and the only difference is the include headers are now in a subdirectory (but still within the default header search path).

However, when attempting to `make package` I encounter this error on macOS:

`CPack Error: Cannot initialize CPack generator: PackageMaker`.

This makes sense. PackageMaker on macOS is long-deprecated. Let me see if I can find out how to tell CMake to use the current tools, or at least put in a guard for macOS prior to merging this.

balthisar:
Okay, I found the macOS issue, so I'll merge this and then apply the macOS fix.

AdamMajer:
There should be no problem with this change for any Linux distribution. If they don't need it, they will just not distribute the .pc file.

spk:
Great, thanks !

PR #537, opened 2017-04-20
Correctly process 'bookmarks' in html exported from Google Doc.

deathbaba:
HTML exported from Google Docs contains empty anchors like ```<a id="random id"></a>```. They are used to "bookmark" any part of the document and make a link to it.

Looks like in the past (or for some specific documents) Google has used anchors like ```<a name="random id"></a>```, but now I don't see them any more.

Previous code moved "name" attribute to the parent "id" tag and deleted empty anchors. Proposed fix also supports "id" attribute in addition to "name".

### DIFF
--- a/src/gdoc.c
+++ b/src/gdoc.c
@@ -115,6 +115,11 @@ static void CleanNode( TidyDocImpl* doc, Node *node )
                 else if (nodeIsA(child) && !child->content)
                  {
                     AttVal *id = TY_(GetAttrByName)( child, "name" );
+                    /* Recent Google Docs is using "id" instead of "name" in
+                    ** the exported html.
+                    */
+                    if (!id)
+                        id = TY_(GetAttrByName)( child, "id" );
 
                     if (id)
                         TY_(RepairAttrValue)( doc, child->parent, "id", id->value );


geoffmcl:
@deathbaba thanks for the PR... we like those...

Have not yet tested, but this looks like a simple, effective change...

Have you run the [Regression Tests](https://github.com/htacg/tidy-html5-tests) with this change? While this seems to comply with all html5 spec read, I have a small concern about legacy html4 docs... Maybe the change needs to also check the active doctype... but maybe not...

Given a suitable consideration of the above, and time for other comments, and feedback, will consider merging this in a few days... thanks...

balthisar:
@geoffmcl, there are no regressions in the `next` testing branch, but there are no test cases using `gdoc`. However when used with `gdoc`, it works as advertised.

I suggest that we can accept this PR as is, however, I wonder if we should also check HTML version level, @deathbaba. For example, in HTML5 documents, `name` [is conforming but deprecated](https://www.w3.org/TR/html5/obsolete.html#obsolete-but-conforming-features):

> Authors should not specify the name attribute on a elements.

This PR correctly allows `id` to be used, as is current standard for HTML5, but I wonder if we should flag a warning if `name` is found (in HTML5).

For HTML<5, I don't see an issue.

We could accept this as it fixes `gdoc`, then open another issue to consider the deprecation of `name`.

Thoughts?

deathbaba:
Regression tests are all passing.

I suppose the warning about deprecated `name` is not related to `-gdoc` option at all. `-gdoc` should only clean html exported from Google Docs and it does it well with a fix. Interestingly enough, html exported from gdoc does not have DOCTYPE at all.

balthisar:
Fair enough. If no one objects to merging this, let's plan on rolling it at the same time as 461 and 338.

geoffmcl:
@deathbaba as stated thanks for this... while I share the idea with @balthisar that the `name` has been depreciated in html5, but this can be raised as a separate issue if needed...

geoffmcl:
@deathbaba I just created a simple gdoc online, and downloaded it, as a html zip... my last creation there was 2012!

I only had a header, `<h1>`, and two `<p`>... 

**Wow** what a giant `<style>` **mess** they create ;=)) Creating some 25+ style elements, and only use abt 5 in the document...

And yes no `<!DOCTYLE ...>`, which would default to html5, but then uses legacy **meta** - `<meta content="text/html; charset=utf-8" http-equiv="content-type">`, and **style** - `<style type="text/css">`... 

It immediately reminded me of the **old** MS Word 2000 export to html, which added lots of proprietary tags and attributes, but maybe not quite as bad... but close ;=))

Default tidy barked about the some of these, and immediately deleted 4 empty spans...

But **zut**, adding the `-gdoc` option certainly massively **simplified** the content...

Then I got adventurous, and added a couple of internal links - `bookmarks` - to the page and saw how tidy-5.5.18 correctly adds the `id` to the paragraphs, while tidy-5.5.14 **failed**...

Good work, and now see the `name` attribute has **nothing** to do with this issue...

**Seems a very good fix**... thanks...

PS: Add test [in_537.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_537.html) to my tidy-test repo, so other can try and verify this fix...

Issue #539, opened 2017-05-03
Some html5 coding has drifted into legacy documents


geoffmcl:
While working on an alternate patch for #461, now in the `issue-461` branch, found a **bug** in Tidy.

This is in code I added quite some time back, in `TY_(AdjustTags)` and `TY_(ResetTags)`... The latter correctly sets `doc->HTML5Mode = yes;   /* set HTML5 mode */`, but the former, used when a doctype indicates a legacy document type does **not** do `doc->HTML5Mode = no;`...

That means the service `TY_(IsHTML5Mode)`, used in a few places to control code, flow always returns **yes**. That is html5 mode! UGH!

The effects of this can best be seen in the `regression` tests, and have opened as [Issue 15](https://github.com/htacg/tidy-html5-tests/issues/15) there to deal with this, once the above correction to Tidy is made...

balthisar:
#461 merged, so closing this.

PR #540, opened 2017-05-03
Issue #461 - alternative patch for this issue

geoffmcl:
@lhchavez while your PR #531 does the same thing, this **fix** includes a fix for Issue #539, and some other code tidying... and as mentioned collects all the html5/html4 mode switching into one place... and avoids creating a special parser just for `<button>`...

Would be appreciated if you could checkout the `issue-461` branch, and make sure it does the same job... thanks...

Of course, at this time it will **fail** the compare on some 9 `regression` tests, which I will address through [Issue 15](https://github.com/htacg/tidy-html5-tests/issues/15) once this is merged...


balthisar:
@geoff, confirmed working in both HTML4 mode (i.e., current behavior) and HTML5 mode (desired behavior) on macOS.

Nice fix on the HTML detection; this is definitely preferable to adding a new option.

Issue #545, opened 2017-05-06
Support for Continuous Integration


lhchavez:
It is super useful to have an automated way of running the whole suite of tests automatically on every pull request and commit, which reduces a bit of verification burden off maintainers and ensures that no change ever introduces regressions. This can be trivially achieved by using a service such as [Travis CI](https://travis-ci.org/), which is free for open source projects.

balthisar:
Not sure why this was closed. The Bash scripts did _not_ include this topic. Re-opening. Sorry for any confusion delivered via email.

lhchavez:
That was probably my fault. I included the string "fix ${URL}" in the commit message, and GitHub automatically closed the issue based on that.

balthisar:
This will be a longer, more in depth discussion. Here are some back references: #330, #269, #261, and #175.

So far, there's not been any real interest. Complicating things, in some cases we expect (or get!) different results on different platforms, meaning Travis good 99% of the time, but not 100% of the time.

I'm open to it, but there's generally never been much of a case for it.

lhchavez:
Ah I did search the issues for "Travis", "CI", "Continuous Integration", but failed to also look at the closed ones.

I completely agree that developers should test their changes before sending them over for review and have been heavily pushing pre-upload tests on all the projects I've worked on. They are amazing and do reduce the amount of noise that even gets to the PR stage. Unfortunately, git does not support automatically enforcing these :(

Then again, as you mention, different results on different platforms is a real issue. Not all developers have access to all combinations of OSs, and I believe CI can help here. Throwing a Windows-specific CI like Appveyor into the mix, and you should be able to get 99.5+% since you'll be able to cover the three most popular OSs.

In any case, it seems like my case is not any stronger than the previous ones that have already been rejected, so I'm happy that at least the improvement to the scripts was accepted.

balthisar:
I'll leave it open for discussion. Like I said, I'm not against it. It would save me from spinning up a Linux VM for testing, for example. Or a Windows VM for that matter.

The Holy Grail would be something that can test Raspberry Pi, Windows, some variation of Linux, and macOS. As you can see, we often delay commits because we like to have confirmation from different environments. Sometimes, though, it's obvious that a change is not platform specific, and so those are usually merged without too much formality.

[...]

PR #546, opened 2017-05-06
Add Travis-CI support

lhchavez:
This change adds the tidy-html5-tests repository as a submodule and
makes Travis-CI run the full suite of tests on every commit and PR.

This change is intended to fix https://github.com/htacg/tidy-html5/issues/545,
but since this checks out the test repository as a submodule, there are two
downsides:

* Any commit that requires modifications to the tests would either appear
  as broken, or not receive automated coverage (in the case new tests are
  to be added).
* Every time the test repository is updated, the submodule must be uprevved.

This change also only runs the `testbase` set, because I could not find how to
run the others correctly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "tests"]
 	path = tests
-	url = https://github.com/lhchavez/tidy-html5-tests.git
+	url = https://github.com/htacg/tidy-html5-tests
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "tests"]
 	path = tests
-	url = https://github.com/htacg/tidy-html5-tests
+	url = https://github.com/htacg/tidy-html5-tests.git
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tests"]
+	path = tests
+	url = https://github.com/lhchavez/tidy-html5-tests.git
```
lhchavez: Note that this needs to be changed to https://github.com/htacg/tidy-html5-tests , but since I needed to test this out with my own repository I had to hardcode my own fork here.

If we ever decide to go forward, I'll change this with the correct URL.

lhchavez:
PTAL!

This is how it would look like: https://travis-ci.org/lhchavez/tidy-html5/builds/229552310 . Every pull request would also display a small green / red badge with the result of the tests, for immediate feedback to any PR author.

Note that this requires https://github.com/htacg/tidy-html5-tests/pull/18 in order to work.

Issue #352, opened 2016-01-23
tidy-html5 generates invalid HTML5 when href elements contain illegal characters


ncouture:
When asking tidy-html5 to fix URIs using the fix-uri option (that is active by default), invalid href element characters are present in the output of tidy-htm5.
#### case in point (fix-uri documentation):

> fix-uri
> 
> Type:    Boolean
> Default: yes
> Example: y/n, yes/no, t/f, true/false, 1/0
> 
> This option specifies if Tidy should check attribute values that carry URIs for illegal characters and if such are found, escape them as
> HTML4 recommends.
#### example and reproduction steps

``` bash
 $ echo '<meta charset="utf-8"><title>xyz<a href=":|">invalid HTML5</a>' | tidy --doct --tidy-mark no --fix-uri yes - 2> /dev/null 
```

output

``` html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<title>xyz</title>
</head>
<body>
<a href=":|">invalid HTML5</a>
</body>
</html>
```

``` bash
$ echo $(!!) | html5check -h
```

output:

``` ascii
Error: Bad value “:|” for attribute “href” on element “a”: Illegal character in path segment: not a URL code point.
From line 1, column 88; to line 1, column 100
There were errors. (Tried in the text/html mode.)
```

html5check: https://about.validator.nu/html5check.py

geoffmcl:
@ncouture thanks for reporting...

Only had time for brief test tonight, but can confirm current tidy seems to ignore those **bad** chars... and indeed so does tidy circa 2009, even back to tidy 2000... 

Are we missing something in configuration here? Do we need to `enable` something else, aside from `fix-uri` which is on by default?

Tidy has a `CheckUrl` (CH_URI) service, but is does not yet seem called on your sample, but even if it was, I can not see where it would flag those characters as bad?

I am tentatively marking it as a **bug**, and will explore more...

Maybe others have comments, ideas, patches, PR...  and maybe some W3C references on valid URI charcters... thanks...

ncouture:
To answer your question I have two comments and one question:

- URLs look better without urlencoding special characters
- AFAIK there is no other impact to it than having "invalid" HTML

Why are characters like "pipe" classified as invalid for use in HREF elements???

geoffmcl:
@ncouture thanks for the further comment...

[...]

Issue #338, opened 2015-12-22
Maybe tidy access Warnings need some TLC?


geoffmcl:
Going back to a sf [bug](https://sourceforge.net/p/tidy/bugs/932/) I was able to create a sample html that continues to erroniously issue this audio access warning - using config `--accessibility-check 3` - on -

Input:

``` html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>#sf932 - href with .au</title>
<style>
body { background-color: #d0e4fe; }
</style>
</head>
<body>
<p><a href="mailto:something@other.com.au">Mail Me</a></p>
</body>
</html>
```

Ouput:

```
Accessibility Checks: Version 0.1

line 1 column 1 - Access: [3.2.1.1]: <doctype> missing.
line 6 column 1 - Access: [6.1.1.2]: style sheets require testing (style element).
line 13 column 4 - Access: [1.1.6.2]: audio missing text transcript (au).
No warnings or errors were found.

Info: Document content looks like HTML5
```

Expected:

Now, when I look at it, even the other two warnings are suspect, but for sure the au is nothing to do with `audio`!

Any help in addressing this would be much appreciated... thanks...

geoffmcl:
Since no one has stepped forward to work on this, moving the milestone to 5.3

We are considering a 5.2 release shortly, and improvements in these access warning is unlikely to happen before then...

geoffmcl:
In combing through the **old** SF bugs, and closing them as `out-of-date` ran across more access test problems in [Bug #797](https://sourceforge.net/p/tidy/bugs/797/) I added 2006-10-28, listing and documenting other problems with these tests...

It would be really appreciated if someone would take an interest in this issue, and hopefully commence to sort things out better... thanks...

geoffmcl:
Since we are to shortly release 5.4, and no work has been done on this, am moving it to `next` 5.5... thanks...

geoffmcl:
While the merge #541 might not exactly close this, in that there are probably other access tests that need TLC, they can be opened as new issues...

PR #547, opened 2017-05-06
Attempt to address issue #352. 

balthisar:
This patch correctly address the specific issues in #352, but I'm worried that there's some over-reach here.

Currently only implemented as a warning, with no switch to turn it off, which
maintains current behavior other than the warning.

In general, we're treating any string as a complete URL, rather than breaking
URL's into component parts. Thus the `IsURLCodePoint()` check includes a few
other generic characters that strictly speaking aren't valid codepoints, but
are valid as escape characters and delimiters.

When addressing #338, I ran into a similar situation in not having a built-in
method to separate path components (although a simple generalized solution was
good enough in that case).

Thus without introducing a new structure and functions to deconstruct a URL
into scheme, authority, path, parameters, etc., some variation of this patch
will have to be used to address #352.

Current no regressions with the `next` testing branch. I've used this as a proposed new test case:

~~~
<!doctype html>
<html>
<head>
<meta charset="utf-8" />
<title>issue 352, illegal character in path segment</title>
</head>
<body>
<a href="€">technically valid HTML5</a>
<br>
<a href="http://€.com">technically valid HTML5</a>
<br>
<a href=":|">invalid HTML5</a>
<br>
<a href="http://[">also invalid</a>
<br>
<a href="http://www.example.com">valid</a>
<br>
<IMG SRC="circleincircle.gif"
   HEIGHT=158 WIDTH=158 ALT="Click the O!" BORDER=0
   USEMAP="#map1"><BR>
<br>
<a href="%E2%82%AC">Same €, already escaped</a><br>
</body>
</html>
~~~

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -171,6 +171,7 @@ extern "C" {
     FN(ESCAPED_ILLEGAL_URI)           \
     FN(FIXED_BACKSLASH)               \
     FN(ID_NAME_MISMATCH)              \
+    FN(ILLEGAL_URI_CODEPOINT)         \
     FN(ILLEGAL_URI_REFERENCE)         \
     FN(INSERTING_AUTO_ATTRIBUTE)      \
     FN(INVALID_ATTRIBUTE)             \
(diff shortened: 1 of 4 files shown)


geoffmcl:
@balthisar thanks for this, but in testing your `issue_352` branch, have some questions which I will post on #352... that is in the originating issue... if that's ok with you?

Issue #391, opened 2016-04-02
In unix given a directory, tidy will generate a document!


geoffmcl:
In combing through, and closing all the **old** SF bugs as `out-of-date`, came across [Bug 744](https://sourceforge.net/p/tidy/bugs/744/), created: 2005-12-06 by Jeff Pohlmeyer, and he even offered a simple solution...

Problem: only in linux

```
$ tidy /usr/bin
line 1 column 1 - Warning: inserting missing 'title' element
Info: Document content looks like HTML5
Tidy found 1 warning and 0 errors!

<!DOCTYPE html>
<html>
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 for Linux version 5.1.50">
<title></title>
</head>
<body>
</body>
</html>
...
```

While that might be `funny`! Ha, ha, it is a bug in my book. 

And also do not like the simple `Error: Can't open "dir-name"` in Windows. It is not just `can not open`, it is a **directory**, and should say so!

Patch or PR welcome... thanks...

geoffmcl:
Since we about to release a 5.4, and no work has been done on this, am moving it to `next` 5.5... thanks...

balthisar:
@geoffmcl, I'd suggest the expedient fix would be the console application... should be an easy check. A more robust solution would be to implement this is the library and add a new message, because `TY_(FileError)` is used all over the place.

balthisar:
I ended up with a simple fix in the library, including a new message. Tested on macOS and Windows.

balthisar:
Thanks for the fix, @geoffmcl. I've merged it, so will close this.

PR #549, opened 2017-05-07
Address #391. Tested on macOS and Win10.

balthisar:
- Add a check upon opening a file for validity of the file.
- Add a new message to indicate that the path is not a file.

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -101,6 +101,7 @@ extern "C" {
  */
 #define FOREACH_MSG_MISC(FN) \
 /** File can't be opened */                        FN(FILE_CANT_OPEN)              \
+/** Not a file */                                  FN(FILE_NOT_FILE)               \
 /** line %d column %d */                           FN(LINE_COLUMN_STRING)          \
 /** Document content looks like %s */              FN(STRING_CONTENT_LOOKS)        \
 /** discarding */                                  FN(STRING_DISCARDING)           \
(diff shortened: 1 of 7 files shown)


geoffmcl:
@balthisar thank you for looking into this #391 issue...

I have only tested in Win10, and it seems to work... **except** you need to move the `fclose(fn)` up to before the `#ifdef _WIN32` switch... windows will **not** allow that file to be opened again... now pushed - `Close the file before the _WIN32 switch f7e7554`...

I did think of using the existing libtidy service `tidyFileExists`, but this uses `access`, which, for some reason, I never liked, **and** at this time we have a `TidyDocImpl *doc`, **not** a `TidyDoc tdoc` required for that service - we would need to move it back to `tidyParseFile(tdoc,...)` - **and** maybe a `directory` would pass the `access` test... not tested, tried... forget it... too troublesome...

I also thought of using `stat` instead... this **saves** opening it as "r+" only to close it, also something I do not like, **and** could replace the later `fstat` done... but when I tried to do this, this added even **more** squiggly code... it would work... 

But involves yet another macro for `_WIN32` to test for a `directory`... just the usual **with** or **without** a leading underscore, like -

```
#ifdef _WIN32
#define M_IS_DIR _S_IFDIR
#else /* !_WIN32 */
#define M_IS_DIR S_IFDIR
#endif /* _WIN32 y/n */
```

So on balance, **if** this works across platforms, let's go for what you have... and like the new `descriptive` message...

Will get around to testing linux, but if it works on the macOS, no reason to think not... merge it... thanks...

Issue #378, opened 2016-02-20
Accents in URIs cause incorrect warning "escaping malformed URI reference"


a3nm:
Consider the following document (in UTF-8):

```
<!DOCTYPE html>
<html dir="ltr" xml:lang="en" lang="en">
<head>
  <title>a</title>
  <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
</head>
<body>
  <p><a href="http://example.com/é">foo</a></p>
</body>
</html>
```

When running `/usr/local/bin/tidy -utf8 -errors -quiet` on this document, I get the following warning:

```
line 8 column 6 - Warning: <a> escaping malformed URI reference
```

The warning only shows up when the href contains an accented character. This seems bogus to me, I don't think non-ASCII characters in href are disallowed. Further, validator.w3.org accepts the document. I think the warning shouldn't be emitted.

This may be related to #352, but maybe not.

geoffmcl:
@a3nm I think this is **very** related to #352 - sort of `what are valid characters in a URL`?

As then, I again searched **hard** for answers here. All I can read according to [rfc 1738](http://www.rfc-editor.org/rfc/rfc1738.txt), and many other references, all characters out of the US-ASCII range must be escaped, and even quite a few within that range!

Even trying an example in the [W3CSchools](http://www.w3schools.com/tags/ref_urlencode.asp), admittedly not to be taken as W3C approved, escapes such `accented` characters...

As indicated in #352, the existing internal tidy service [CheckUrl](https://github.com/htacg/tidy-html5/blob/master/src/attrs.c#L1469) has a simple test of what characters tidy will escape, namely `if ((c > 0x7e) || (c <= 0x20) || (strchr("<>", c)))`, and even this range check should maybe be extended to more invalid characters...

I would ask you, and others, to provide further W3C references on this issue... and the fact that the W3C validator does not flag a problem is definitely **not** the sole criteria to be considered... thanks...

a3nm:
[...]

Issue #365, opened 2016-02-04
SVG elements and attributes are being lowercased


mojavelinux:
SVG elements such as "clipPath" and attributes such as "viewBox" (when included in the document via the `<svg>` element) are being lowercased. This could cause SVGs to stop working (though in my tests it doesn't seem to break inside the browser). At the very least, it's annoying because it causes unnecessary diffs.

Input:

```
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 16 9"></svg>
```

Output

```
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewbox="0 0 16 9"></svg>
```

There are many such elements and attributes, as you can see here: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute

Is there any way we could disable the automatic lowercasing within SVG regions?

balthisar:
Will `input-xml` or `output-xml` work for your use case? They both preserve attribute case.

mojavelinux:
It's true that there are certain HTML documents that are able to masquerade as XML going in, which does result in the element/attribute case being preserved (regardless of output setting). However, the first time the parser his a perfectly valid short element (e.g., `<img src="..." alt="...">`), it blows up. I also notice that by enabling `input-xml`, there are other side effects. For instance, the newlines around content in a `<pre>` element show up again (fixed earlier in 5.1 for HTML).

But I think we're missing the point. An inline SVG is an XML document inside of an HTML document. Therefore, it should be handled as XML, and that, in part, means preserving the case of elements and attributes.

The way I've been able to workaround this is to wrap the SVG in CDATA before tidy touches it, then unwrap it afterwards. That seems to keep tidy from touching the SVG data. Not pretty, but it gets the job done.

mojavelinux:
Interesting to note that there is a related discussion in Jsoup.

https://github.com/jhy/jsoup/issues/272

mojavelinux:
It seems like this behavior might actually be desired. In a reply on Twitter to $subject, Sara Soueidan said:

> I always lowercase [because] some attributes had problems w/ uppercase in IE a while back.

The concern I had, as well as others, was that by changing the case, it would break SVG rendering. However, it seems like that might not be true (and in the case of older IEs, it might actually help the situation).

I'm curious to hear what others think.

balthisar:
I see your points. Due to the followup questions you raised, I won't tag this as a bug or feature request, but I'll leave it open for discussion.

[...]

PR #553, opened 2017-05-11
New picklists and parsers

balthisar:
See the commit notes, but essentially this refactors the current picklists and option parsers to remove a bit of code redundancy, but more importantly, to make it **vastly** easier to implement new potential option values without having to write new parsers or awkwardly use existing parsers.

There are no testbase regressions on macOS; I've not built and tested on Windows as of yet.

### DIFF
--- a/README/OPTIONS.md
+++ b/README/OPTIONS.md
@@ -38,7 +38,7 @@ struct _tidy_option
     TidyOptionType      type;       /* string, int or bool */
     ulong               dflt;       /* default for TidyInteger and TidyBoolean */
     ParseProperty*      parser;     /* parsing method, read-only if NULL */
-    const ctmbstr*      pickList;   /* pick list */
+    PickListItems*      pickList;   /* pick list */
     ctmbstr             pdflt;      /* default for TidyString */
 };
 ~~~
@@ -78,7 +78,7 @@ Care, each of these enumeration strings have been equated to two uppercase lette
 
 The next item is the `default` value for a boolean, tristate or integer. Note tidy set `no=0` and `yes=1` as its own `Bool` enumeration.
 
-There are a number of `parser` for the options. Likewise a number of `pickList`. Find another option similar to your new option and use the same values.
+There are a number of `parser` for the options. Likewise a number of `pickList`. Find another option similar to your new option and use the same values. The `parser` is the function that parses config file or command line text input, and the `picklist` constitutes the canonical values for the option. Some types of values logically don't have picklists, such as strings or pure integers.
 
 Presently no options have the final `default` string, and it is left out of the table. The compiler will add a NULL.
 
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -724,6 +724,36 @@ typedef enum
     TidyCR          /**< Use Macintosh style: CR */
 } TidyLineEnding;
 
+    
+/** TidyEncodingOptions option values specify the input and/or output encoding.
+ ** @remark This enum's starting value is guaranteed to remain stable.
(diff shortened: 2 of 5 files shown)


balthisar:
Builds and tests cleanly on Windows 10 and Ubuntu 16.04.

geoffmcl:
@balthisar just checked out `new_picklists` branch to test, in Windows 10, MSVC14, **but** the build **FAILS**???

```
  config.c
F:\Projects\tidy-html5\src\config.c(596): error C2065: 'ParseAutoBool': undeclared identifier [F:\Projects\tidy-html5\build\temp-553\tidy-static.vcxproj]
F:\Projects\tidy-html5\src\config.c(596): warning C4047: '==': 'ParseProperty (__cdecl *const )' differs in levels of indirection from 'int' [F:\Projects\tidy-html5\build\temp-553\tidy-static.vcxproj]
````

Have I missed something?

Have not had a chance to look into yet, but maybe you will see the reason immediately... thanks...

**Comment by balthisar** [comment]
Yup, I see it. I'm using MSVC 17, and for some reason the _DEBUG macro isn't enabled for me, thus the function to the non-existent ParseAutoBool isn't being flagged for me.

I've just pushed a fix. I'm not sure how to enable _DEBUG when building for Windows. For that matter, I should figure out why Clang isn't defining that macro, too.

**Comment by geoffmcl** [comment]
@balthisar starting to look into it...

One thing I remember you saying is that I think you **only** build the `Release` version... In the `Release` version an `assert` compiles to nothing...

Try -
```
build\Win64> cmake --build . --config Debug
```

Only then will an `assert` exist...  and this error is in such an `assert`... thanks...

PS: Looks like you found it... but for the future I always build **both** Debug **and** Release in Windows... Do you use the above `cmake` to do the building... I do not think _DEBUG is the same thing...

balthisar:
I usually build on Windows using `--config Release` because that's in the build instructions. I'll just remember to use `Debug` in the future.

geoffmcl:
Yes, it now builds, and will get on with testing...

In fact the controlling MSVC macro is `/D NDEBUG`... if this is defined then there is **no** `assert`!

And I think this is the **same** in gcc, so would assume same in `Clang`... it is defined that way in some standard header...

In fact, never sure why tidy uses `assert` at all, since only if there is **no** `-DNDEBUG`, does `assert` do anything, thus is really only for the **Debug** build... 

So we are sort of asking **all** developers to first build and test in the **debug** version... But that is perhaps another topic...

geoffmcl:
@balthisar have now compiled and tested this branch, in Windows 10, and see no immediate problem in this merge...

A **small** niggle! 

It would be **really** appreciated if you would consider opening a `Feature Request` prior to **refactoring**, and presenting a `fait accompli` PR, saying one can read the commit notes...

This would give us a chance to present objections, ideas, feedback, ... or nothing... **before** all the effort is expended... not that there would be any objection to such a `change` and `cleanup`, but would also allow for suggested improvements, enhancements, etc... that could be considered...

Just a quiet **suggestion**... it is already noted in the `CONTRIBUTING.md`, which you wrote - "it is best to create an issue before putting the effort into a pull request". Why **not** follow this?

Anyway, for what, at this time, after such a brief, pushed, review, looks like a **good** move... thanks...

balthisar:
Thanks for having a look, @geoffmcl. This was a pre-requisite for #365 and #378 -- infrastructure -- so a feature request really wasn't appropriate. A PR gives everyone a chance to say yay or nay, otherwise it would have been included in one of the bug fixes anyway.

PR #554, opened 2017-05-12
Issue 365

balthisar:
Address #365 by introducing a new option to uppercase-attributes.

Requires #553 to be merged, first, since this takes advantage of the updated picklist/parser system. Actually, #553 doesn't have to be merged first, but it would be better style than incorporating all of this at once.


geoffmcl:
LGTM... see comments on #365...

Issue #597, opened 2017-09-14
Memory leak parsing "<a><div><a href=1>"


ralfjunker:
Parsing this HTML with default options:

```html
<a><div><a href=1>
```

results in a memory leak.

The problem happens because https://github.com/htacg/tidy-html5/blob/a79458a0ef24b0082603caf4b2e56b542cd6a31c/src/parser.c#L1362-L1365 prevents the tag from being unset / freed.

Similar code is in https://github.com/htacg/tidy-html5/blob/a79458a0ef24b0082603caf4b2e56b542cd6a31c/src/parser.c#L1922-L1933 takes care to free the tag if it is not unset.

A similar else clause should probably be added above.

geoffmcl:
@ralfjunker thank you for the issue... it caused me to turn **on** some memory diagnostics that I have not run for a **long** time, and indeed do find **several** `memory leaks`...

I had to further enhance my memory diagnostics to track them down... that I am in the process of doing...

But the particular case you point to does **not** seem to be one of them? 

First, the `parser.c[1362-1365]` lines are correct... the process of `TY_(UngetToken)( doc );` is **not** trying to discard and free a node. It pushed the `node` onto the lexer stack, by `doc->lexer->pushed = yes;`, and allows tidy to continue processing the current `token`, and that pushed node will be recovered, and dealt with on the next `GetToken(doc)`... Read the top of `lexer.h` for a bit of an explanation of this...

As you further point to `parser.c[1922-1933]`, that shows a case of a switch between using `UngetToken`, which will be later used, and `FreeNode` which completely destroys/frees the node...

So while I am not sure the code you point to is the problem, I am **sure** there appears to be some 16 memory leaks - from my additional memory diagnostic output, massaged with perl script, `shwtidymem.pl`, to give a **summary**, which is -

```
Remains 16 in HASH
# :cnt: address          size
 1: 74: 0000020AF9707860 136
 2:122: 0000020AF97087A0 136
 3: 36: 0000020AF9707FD0 2
 4: 13: 0000020AF970C750 136
 5: 97: 0000020AF9706DC0 136
 6: 38: 0000020AF9708050 2
 7:105: 0000020AF9706E90 136
 8: 49: 0000020AF9707540 136
 9: 41: 0000020AF9707470 136
10:115: 0000020AF97086D0 136
11: 39: 0000020AF97073F0 56
12: 24: 0000020AF97071C0 136
13: 37: 0000020AF9709AA0 5
[...]

PR #599, opened 2017-09-17
Issue #597 - Memory tests/diagnostics

geoffmcl:
Some slightly enhance **old** WIN32/MSVC memory diagnostic code, which could probably be ported to `unix`, the addition of a WIN32 `<crtdbg.h>` module, and added cmake options to enable...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,27 @@ else ()
 add_definitions ( -DSUPPORT_CONSOLE_APP=0 )
 endif ()
 
+# Enable building with some memory diagnostics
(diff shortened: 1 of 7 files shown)


balthisar:
Nice fix. Works on macOS; I need to get SPRTF working on macOS and Linux so that we have the same tools available.

Feel free to merge.

You don't mind if I start poking around SPRTF to make it cross-platform? Then I'd probably take a look at all of the macros that invoke it, too.

geoffmcl:
@balthisar now merged, and bumped to 5.5.40...

Re SPRTF, as replied on #604 look forward to this... advise if any probs... thanks...

PR #605, opened 2017-09-19
Add Brazilian Portuguese translation

rffontenelle:


rffontenelle:
Looks like I messed a bit this PR. The localize/README.md says I should make a PR against "master" branch, but I seems I based my translation in the default branch "next". Which branch should I use?

balthisar:
> Looks like I messed a bit this PR. The localize/README.md says I should make a PR against "master" branch, but I seems I based my translation in the default branch "next". Which branch should I use?

@rffontenelle, thanks. Yup, good catch. It should be based on `next`. Sorry about that. Can you rebase based on `next`? That should clean up this PR greatly, and a Portuguese translation would be simply awesome!

rffontenelle:
Actually, I had it based in `next` thinking it was `master` and then I created this Pull Request against `master`. I changed the target branch to `next`, so it should be good to go. Any problems, please let me know.

geoffmcl:
@rffontenelle **AWESOME** - thank you...

And sorry about the misleading PR against `master`, in `localize/README.md`. 

We missed that when we switched to using `next` as the default development branch... and maybe this also needs to be fixed in `README/CONTRIBUTING.md`...

The file `README/BRANCHES.md` does already make it clear that a PR should be against the **development** `next` branch... but for sure all files should convey the **same** idea...

From my perspective this can be merged, but @balthisar is our language guru... thanks again...

balthisar:
@rffontenelle, thanks, it's merged. You might have a new look at it with the current PO or POT, just to make sure a translation didn't go missing. I had to rebase it because there was a previous merge that should have only affected string order, but better safe than sorry.

balthisar:
@rffontenelle, also it just occurred to me (and I should comment the POT) that `TEXT_GENERAL_INFO_PLEA` would probably read better if it referred to Brazilian Portuguese instead of English, because presumably someone in Brazil would be reading the message in Brazilian Portuguese.

>"Você fala um idioma diferente do inglês, ou uma variante diferente do\n"
>"inglês? Considere nos ajudar a traduzir HTML Tidy. Para detalhes, consulte\n"
>"https://github.com/htacg/tidy-html5/blob/master/README/LOCALIZE.md"

Obviously this isn't a demand, but if you should happen to issue another PR, it would be accepted!

By the way, I can't understand spoken Brazilian, but can somewhat understand Portuguese people when I run into them. How different is the _written_ language? I can understand most of what you wrote... would this be appropriate to use as the Portuguese base language, too? Or is it really, really specific to Brazil?

rffontenelle:
I don't mind translating `TEXT_GENERAL_INFO_PLEA` back to English, but would it be possible to mark this string as not translatable? This would be safer and easier for both translation and maintenance of the translations.

> By the way, I can't understand spoken Brazilian, but can somewhat understand Portuguese people when I run into them. How different is the written language? I can understand most of what you wrote... would this be appropriate to use as the Portuguese base language, too? Or is it really, really specific to Brazil?

Indeed, the accent and slangs are very different between European Portuguese (pt) and Brazilian Portuguese (pt_BR), but the written language is not so different. It has its differences, like "file" in pt_br is "arquivo" and in pt is "ficheiro", but I understand their writing, so they should understand mine.

So, I don't see as a problem using my translation as "pt" for now. When someone who speaks European Portuguese provide translation to pt, just move mine back to pt_br.  How does it sound?

If decide to rename pt_br to pt, let me know and I'll wait before fixing `TEXT_GENERAL_INFO_PLEA` and fuzzy translations, if any.

balthisar:
> I don't mind translating `TEXT_GENERAL_INFO_PLEA` back to English, but would it be possible to mark this string as not translatable? This would be safer and easier for both translation and maintenance of the translations.

Oh, my meaning was, leave the string in Portuguese, but change to it to something like:

>"Você fala um idioma diferente do _Portugues Brasileño_, ou uma variante diferente do\n"
>"_Portugues Brasileño_? Considere nos ajudar a traduzir HTML Tidy. Para detalhes, consulte\n"
>"https://github.com/htacg/tidy-html5/blob/master/README/LOCALIZE.md"

My substitution is Spanish because I can't write Portuguese!

This way, Tidy users in Brazil who speak another language might want to offer another translation, too.

rffontenelle:
Oh, sorry for misreading of your previous comment. Consider it done.

PR #606, opened 2017-09-19
Dialogue cleanup

balthisar:
Internal refactoring of non-report type messages emitted by Tidy. Each message has been categorized by purpose, and these are kept in a master array for each message type. Like the report messages, a dialogue message is now formatted by a "formatter," giving us a single dialogue message output routine.

No diffs generated!

This is a nice preparation for some additional work I'd like to complete, for which I'll add feature requests.

Also added the doxygen documentation to the header file.


geoffmcl:
@balthisar checked out this `dialogue_cleanup` branch, and it built fine in Windows...

And a run of the regression tests using this 5.5.37-exp-jsd, 2017.08.31 version showed no problems...

From my perspective, this can be **merged**, with the obvious `version.txt` change... thanks...

Issue #412, opened 2016-05-26
Add warning for unquoted attribute containing `=`


zanona:
Apparently tidy does not complain about markup like 

``` html
<meta name=viewport content=width=device-width,initial-scale=1>
```

On W3C validator [this throws an error](https://validator.w3.org/nu/?doc=data%3Atext%2Fhtml%2C%253C%21doctype%2520html%253E%253Cmeta%2520name%3Dviewport%2520content%3Dwidth%3Ddevice-width%2Cinitial-scale%3D1%253E%253Ctitle%253ETitle%253C%2Ftitle%253E):

> Error: = in an unquoted attribute value. Probable causes: Attributes running together or a URL query string in an unquoted attribute value.

Currently tidy's output is already great where it successful wraps the content within quotes properly.
This is just a Warning/Error request issue.

Thanks in advance

geoffmcl:
@zanona thanks for the issue...

Sorry for the delay... I thought I had replied, but it was just in my imagination ;=))

First, yes tidy could probably warn about this **silent** fix. But it does quite a lot of `quiet` fixes. Should we warn about these too?

And it follows, should tidy follow the validator type path... make a big noise about everything... 

Or be a relatively quiet, mainly fixing tool... just get on with the job of tidying and pretty printing... leave validation to nu?

Hmmm, will have to think about that? Sometimes I think tidy is already too noisy about somethings...

But what do other think? Add one, or more, warnings? Remember, this also effects the exit value of tidy.

Look forward to more comments on this... thanks...

zanona:
Thanks @geoffmcl.
Well, I believe you guys can be very proud nowadays because tidy is the most (if not the only) reliable HTML linter out there. This is my main usage of it actually.

I do really like the pro-activeness of tidy on making things right outputting documents as they should be. It already does a great job on this.

In regards the warnings, I believe this can simply be put behind a flag what do you think? I believe it may be something that would benefit many that want to know about their document status before any processing is done. So if we could add `--verbose` for example, Tidy could simply warn about all the issues _(good for lint in and improving best practices)_. without this flag, things are just done silently as you mentioned.

It would be interesting to know what others think as well.

All the best

geoffmcl:
@zanona, thanks for the kind words... tidy does try to be a good `linter` ;=))

[...]

PR #612, opened 2017-09-22
Address #412

balthisar:
Add a TidyInfo message each time an unquoted attribute is found. However,
refer to #412 for discussion before merging this.


balthisar:
Refer to #412 for additional discussion...

Issue #616, opened 2017-09-26
bdi tag should be marked as inline, not block


behnam:
Current Config: https://github.com/htacg/tidy-html5/blob/0016e0083505ccee25a5c76dcf64cfe336765128/src/tags.c#L300

From [the spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-bdi-element):
> The `bdi` element represents a span of text that is to be isolated from its surroundings for the purposes of bidirectional text formatting.

The language and example clearly show that `bdi` is an inline element and not a block element.

Is there any reason this tag is being treated as block element in tidy?

balthisar:
Confirmed.

balthisar:
Added candidate #620. This is a simple change, and so will merge in a day or so unless objections are raised.

balthisar:
@behnam, confirming that this is merged and live now, so closing this issue.

PR #620, opened 2017-09-29
Address #616

balthisar:
- Fix the declaration and parser for the bdi tag.


behnam:
I'm wondering what's the best way to double-check all the items in the list here. Many of the element display type lists that I check miss many of the tags, like `bdi`, and there's no clear property defined for this in newer specs.

Am I missing something in the specs? Or maybe there's a list maintained somewhere out of the HTML spec?

balthisar:
@behnam, there's not really a master list; it's all based on convention, and with HTML5 in particular it's expressly _not_ defined in the specification. Content versus appearance, and all of that. In general, we try to keep consistent with browsers' default stylesheets' `display` property for each element. These days, all of the browsers tend to agree. Of course you'll notice that a lot of our elements are mixed, and so it becomes a judgment call.

Issue #602, opened 2017-09-19
Eliminate legacy conditionals


balthisar:
At some point, experimental features were added into Tidy that have been stable for years and years and years, and so I see no need to continue to support compiler macros capable of disabling them.

I propose that we remove `SUPPORT_UTF16_ENCODINGS`, `SUPPORT_ASIAN_ENCODINGS`, and `SUPPORT_ACCESSIBILITY_CHECKS`.

I can't fathom a use case that requires the ability to turn these off any more.

geoffmcl:
> I can't fathom a use case that requires the ability to turn these off any more.

@balthisar me neither... they should be removed...

You might also consider `TIDY_STORE_ORIGINAL_TEXT`. 

Although I used it for some diangostic reasons years and years ago, it needed quite a number of patches to make it work, which never made it into the base code, and would be difficult for me to recover now... Even Björn, who I understand introduced the compile option, thought it a bad idea that I try to use it... 

I think it too can go...

There may be others... will look around... thanks...

balthisar:
Proposal #624 available.

PR #624, opened 2017-10-03
Legacy conditionals

balthisar:
Per #602, eliminate a lot of things that have become a permanent part of tidy now, as well as dead code (which is still available in git history, of course).


balthisar:
Passing all tests in Windows, Ubuntu, and macOS, so I will merge this. Binary output should be nearly identical, but will bump version anyway.

PR #626, opened 2017-10-04
Sprtf

balthisar:
This is based off of #624, due to the amount of merge conflicts otherwise. Builds and acts as expected on macOS, Windows 10, Ubuntu.

@geoffmcl, playing with this is a _lot_ nicer than stepping through the code in a painstaking matter! I'm glad that it's basically already compatible with Unix.


balthisar:
Doesn't break Windows, and works a treat on macOS, so in it goes.

PR #627, opened 2017-10-04
Options categories

balthisar:
Assign all existing options to new categories based on how they work with the document. Update the man page xsl in order to use them. This is purely a documentation change.

There is a corresponding PR in the api.html-tidy.org repository to update the quickref documents.


balthisar:
Here's an example of the new, proposed quickref with the new categories.

[quickref_next.html.zip](https://github.com/htacg/tidy-html5/files/1357506/quickref_next.html.zip)

PR #632, opened 2017-10-14
For Issue #604

geoffmcl:
For Issue #604 - Tested in `Windows 10`, `Ubuntu` and `Raspbian`, and all appears ok...

If others can review/test would **merge** this... thanks...

### DIFF
--- a/console/tidy.c
+++ b/console/tidy.c
@@ -22,13 +22,10 @@
 #include "tidy.h"
 #include "tidybuffio.h"
 #include "locale.h"
+#include "sprtf.h"
 
 #if defined(_WIN32)
 #  include <windows.h>    /* Force console to UTF8. */
-#endif
-
-#ifdef ENABLE_DEBUG_LOG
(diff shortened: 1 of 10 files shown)


balthisar:
@geoffmcl, it looks good by me. I did _not_ test in macOS (I'm on the road and only have my Windows laptop). Given that we're not checking anything platform related, and it works for me in MSVC 2010, I'm good with this. Thanks for fixing it up.

If you have a chance to verify with Ubuntu, feel free to merge; otherwise I'll provide macOS and Ubuntu verification on Thursday most likely.

geoffmcl:
@balthisar sorry for the brief delay on this. Got sucked into another project which should end shortly and can then cycle back to `tidy`...

Thanks for the addition testing, but want to add a MSVC only cmake option, `DISABLE_DEBUG_LOG`, to enable cutting out the extra debug code in the MSVC Debug build, if need be...

Be back soonest... thanks...

PR #634, opened 2017-10-20
Ensure that `TidyQuiet` silences everything that it currently quiets,…

balthisar:
… plus

`TidyConfig` and `TidyInfo` messages. This causes a single regression by hiding
`TidyInfo`; will publish PR in testing repo.


geoffmcl:
@balthisar have **tested** this `quiet_fix` branch, and together with [PR 24](https://github.com/htacg/tidy-html5-tests/pull/24) `quiet_fix` branch in tests and it all works as expected...

Given that the one test case changed, 1067112, originally [SF Issue 638](https://sourceforge.net/p/tidy/bugs/638/) was about **fixing** an empty `<o:p></o:p>` tag, and nothing about message output, and this remains **fixed** in the html output, reducing the `expects` message text to a null file seems fully correct... with a config containing `quiet: yes`, and `show-warnings: no`, there should be **no** output...

Also thanks for undating the `.po`, and `.pot` files to better reflect the meaning of `--quiet yes`...

Accordingly, think this and PR 24 in tests should be merged, and will do so shortly, unless you beat me to it... thanks...

balthisar:
Looks I beat you to it! Thanks for checking it out.

Issue #636, opened 2017-10-21
strdup may cause segfault when std=c11 is used


jokester:
When `-std=c11` is used, call to `strdup` may return incorrect result and cause segfault. (It returned a 32bit pointer in my gcc/linux64 build, but works fine in clang/freebsd).

This can be avoided by using internal `tmbstrdup` instead, see [this (conceptual) diff](https://github.com/htacg/tidy-html5/compare/release/5.4...jokester:build-c11-5.4). Do you think this should be fixed?

balthisar:
Can you do it without changing the compiler to c11, though? Tidy is meant to be buildable with c89, although I personally seem to require c89gnu.

jokester:
Sure, I only mean to make it a little more standard complaint.

The fix I proposed should not depend on specific version of C, but I may require some help in not breaking things (not much experience in c/portable here).

BTW tidy may have some unrevealed requirement on C version? I could build last 5.4.0 release with `-std=gnu89` but not `-std=c89`

geoffmcl:
@jokester thanks for the issue...

Seems to be at least two questions here -
 
 1. Use of `strdup` versus `tmbstrdup` in library
 2. GCC C standards, and portability
 
First, concerning the use of `strdup` in the **library**, I feel this is wrong. **ALL** memory allocations in the library should use **TidyAlloc**, ie `tmbstrdup`...

This is so that other apps using `libtidy` can supply their own memory allocator, especially including what to do in case of a `failure`, ie **panic**... When the library is part of a bigger app, it may not be appropriate to just `exit` that app, so they can add their own memory handler...

So my suggested patch for that, which includes replacing the `free` also, is -

```diff
diff --git a/src/language.c b/src/language.c
index 6ef97f5..4f758ea 100644
--- a/src/language.c
+++ b/src/language.c
@@ -335,7 +335,7 @@ tmbstr TY_(tidyNormalizedLocaleName)( ctmbstr locale )
     uint i;
     uint len;
     static char result[6] = "xx_yy";
-    tmbstr search = strdup(locale);
+    tmbstr search = TY_(tmbstrdup)(&TY_(g_default_allocator), locale);
     search = TY_(tmbstrtolower)(search);
     
     /* See if our string matches a Windows name. */
@@ -343,8 +343,8 @@ tmbstr TY_(tidyNormalizedLocaleName)( ctmbstr locale )
     {
         if ( strcmp( localeMappings[i].winName, search ) == 0 )
         {
[...]

PR #637, opened 2017-10-26
use default allocator in place of strdup/free

jokester:
This removes fully-unmanaged memory allocation, and happens to fix problem of non-standard `strdup` call (see #636 )

### DIFF
--- a/src/language.c
+++ b/src/language.c
@@ -299,6 +299,7 @@ ctmbstr TY_(tidyLocalizedString)( uint messageType )
  **  Determines the current locale without affecting the C locale.
  **  Tidy has always used the default C locale, and at this point
  **  in its development we're not going to tamper with that.
+ **  @note this routine uses default allocator, see tidySetMallocCall.
  **  @param  result The buffer to use to return the string.
  **          Returns NULL on failure.
  **  @return The same buffer for convenience.
@@ -306,6 +307,7 @@ ctmbstr TY_(tidyLocalizedString)( uint messageType )
 tmbstr TY_(tidySystemLocale)(tmbstr result)
 {
     ctmbstr temp;
+    TidyAllocator* allocator = &TY_(g_default_allocator);
     
     /* This should set the OS locale. */
     setlocale( LC_ALL, "" );
@@ -315,7 +317,7 @@ tmbstr TY_(tidySystemLocale)(tmbstr result)
     
     /* Make a new copy of the string, because temp
      always points to the current locale. */
-    if (( result = malloc( strlen( temp ) + 1 ) ))
+    if (( result = TidyAlloc( allocator, strlen( temp ) + 1 ) ))
         strcpy(result, temp);
     
     /* This should restore the C locale. */
@@ -329,13 +331,16 @@ tmbstr TY_(tidySystemLocale)(tmbstr result)
  *  Retrieves the POSIX name for a string. Result is a static char so please
  *  don't try to free it. If the name looks like a cc_ll identifier, we will
  *  return it if there's no other match.
+ *  @note this routine uses default allocator, see tidySetMallocCall.
  */
 tmbstr TY_(tidyNormalizedLocaleName)( ctmbstr locale )
 {
     uint i;
     uint len;
     static char result[6] = "xx_yy";
-    tmbstr search = strdup(locale);
+    TidyAllocator * allocator = &TY_(g_default_allocator);
+
+    tmbstr search = TY_(tmbstrdup)( allocator, locale );
     search = TY_(tmbstrtolower)(search);
     
     /* See if our string matches a Windows name. */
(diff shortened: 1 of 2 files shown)


balthisar:
@jokester, it looks good and builds fine for me on macOS.

In the future, you might want to consider the macros `TidyAlloc` and `TidyFree`, defined in `forward.h`; they'll make your life a little easier:

~~~
#define TidyAlloc(allocator, size) ((allocator)->vtbl->alloc((allocator), (size)))
~~~

geoffmcl:
@jokester thanks for this PR replacing `strdup/free` with `TY_(tmbstrdup)/TidyFree`... this is great...

Not so sure about the need for the comment change in `tmbstr.c` but minor...

But your discovery of one case of a `malloc`, and changing it to `TidyAlloc`, certainly raises some questions! 

As far as I can see it is **not** used in the **library**, and is passing the allocated memory out of the library for the user app to free. It guards, and warns in its comments, that it can be `null`, while `TidyAlloc` will **abort** if it fails... quite different behaviour...

This raises the question what is this API doing in a HTML parsing library at all... but will raise that as a separate issue soonest... if/when I get the chance...

Still to get around to a test in an app that **replaces** tidy's default allocator with its own... but expect no problems... Will do that shortly... 

Meantime, no objection to a merge, as is... thanks...

jokester:
Oops, I misunderstood what `tidySystemLocale` really does. Its comment says `result: The buffer to use to return the string.` while in the code `result` passed in is ignored. Also I don't get the latter half: `Returns NULL on failure.` seems more likely to be part of `@return`.

If we consider the first half comment to be truth, not allocating memory here seems more consistent to me (`tidy.c` need to allocate it instead).

geoffmcl:
@jokester,

> Oops, I misunderstood what `tidySystemLocale` really does.

Me **too!**, given its `confusing` comments, and that it only does `setlocal(...)`! Why is it in **libtidy** at all? If `tidy.c` needs to do this, and I am **not** sure it does, then it should do it **all**... It certainly does **not** seem required in Windows...

I would think that a user of a particular language would already have the console setup as desired, i.e. `stdout/stderr`, or could set it up differently, outside `tidy`, if need be... so maybe this API can be removed... 

This is the jist of the separate issue I was going to raise, but maybe can be dealt with here...

Maybe @balthisar will comment when he gets a chance, since I think he introduced this code as part of localization... part of commit 53f2a2da2, and maybe others...

Still to get around to a test the `strdup` replacement using an `external` **allocator**, but as stated, do not expect any problems... thanks...

balthisar:
@jokester, @geoffmcl, that function signature is a monstrosity of bad design decisions, and perhaps its functionality should be moved to the console application.

As for the signature, we should supply a buffer and size from the console:

`void TY_(tidySystemLocale)(tmbstr buf, uint bufsize)`

This breaks current API, though, but since no one uses it except the console application, I'd say bite the bullet, pull the trigger, and make the change.

Now the question as to _why_ this exists:

- Tidy does some sorting internally, e.g., eventually using `TY_(tmbstrcmp)`. This is probably a good thing, because it avoids any locale-related sorting that a standard library might introduce.
- When I added localization, I did _not_ want to affect how Tidy sorts things (i.e., there was no audit on standard library sorting versus the use of Tidy's own sorting). The "safe" thing was to ensure that Tidy internally, always used the "C" locale despite the user locale, in order to work predictably.
- TY_(tidySystemLocale) therefore checks the user's locale without affecting the executable's locale, i.e., the console (and therefore the library) will always run with the C locale. This doesn't apply to LibTidy users, though, because LibTidy will run in its executable's locale.

Of course, Tidy needs to know the user's locale in order to set the language automatically.

As to whether or not we need to preserve the C locale in console Tidy... maybe not! But it was the "conservative" (safe) choice at the time.

geoffmcl:
@jokester now merged... and bumped to version 5.5.74... thanks...

Have done some initial tests with my [tidy-opts](https://github.com/geoffmcl/tidy-test/blob/master/src/tidy-opts.c), which has a `test 2` which replaces the default tidy `allocator`  and all is ok, but need to do more...

Will try to open a separate issue to address the other API questions raised here...

geoffmcl:
@jokester, @balthisar, **OOPS**, all is **not** right, or at least **not** what I `expected`!!!

It seems using `doc = tidyCreateWithAllocator(&allocator.base);` does **not** set the `g_default_allocator/g_malloc/...` as I expected!

It does set the `allocator` used in a `TidyDoc`, so most allocations are done using my replaced `allocator`, but **not** anything using the `g_default_allocator`...

That does **not** necessarily negate this `merge`, but it seems to point out **more** needs to be done...

If I create a `doc` giving my `allocator` I guess I expect that to be used in **all** allocations... seems to do that I maybe need to specifically do `tidySetMallocCall` with my `malloc`, etc, etc...

Will revert after **much** more testing, and understanding... further feedback appreciated... thanks...

Issue #643, opened 2017-11-06
LibTidy does not support all entities! It should!!


geoffmcl:
At present `libtidy`, in `entities.c`, supports only about some 253, **case sensitive**, entities... it uses `TY_(tmbstrcmp)` in the `entitiesLookup` static function...

Yet W3C pages like these - https://dev.w3.org/html5/html-author/charref - or - https://www.w3.org/TR/xml-entity-names/byalpha.html - and probably others, indicate in excess of **2,000** entities, some like `&amp;`, `&AMP;`, ... in multiple case forms... not yet exactly counted, and there may be **more**, yet to be fully explored...

Here is a live web page - https://support.apple.com/en-us/HT202021 - containing the likes of `&Tab;`, `&NewLine;`, pointed out to me by Karl Dahlke, of the [edbrowse](http://edbrowse.org/) project, which uses `libtidy` to load and `view` web pages - thanks Karl...

So at present, `edbrowse` sees warnings about **unknown** entities, and unless `preserve-entities: yes` is added to the config, they will be rendered/returned as text `&amp;Tab`, `&amp;NewLine`, along with the warnings... **not good**!

I propose `libtidy` support **all** known entities, and seek feedback, and more, other W3C references, on this... thanks...

geoffmcl:
Have push a new branch `issue-643` which contains an `entities.c`, with some 2031 entities, scraped from the [charref](https://dev.w3.org/html5/html-author/charref) page with some quick perl...

Now processing that downloaded web page [HT202021](https://support.apple.com/en-us/HT202021) using my new tidy 5.5.76.I643, and adding `--drop-empty-elements no`, the warnings fell from 104 to just 18, and of course none of the warnings were about **unknown** entities.

And the test file pushed to my repo, [in_643.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_643.html), goes from 16 warnings, including 2 spurious `missing <li>`, to clean... 

As noted in the source, I have removed the column alignment to keep the total size down to 80K, rather than the nearly 150K for adding sort of meaningless alignment. This is of course much larger than the original 13K.

I hope this is kept sane... it does break a little from tidy's usual **good** code, especially table alignment...

And as also noted in the source, I have added some more small optimization steps to the `entitiesLookup` service which depend upon the entities table being kept in alphabetic order, and think this should be fast enough for now, rather than adding to the lib setup time by creating some type of hash. 

[...]

PR #644, opened 2017-11-07
Issue #643 - Greatly expand entities list

geoffmcl:
Expand known entity list to 2031...


geoffmcl:
Branch `issue-643` now rebases to `next` to add the **fixed** header install location, see issue #646, PR #647 ... 

Testers **must** do a pull to update their local repo. `issue-643` branch, for further testing, **and** should make sure any installs to `include/tidy` are deleted... thanks...

balthisar:
Beautiful! There are a couple of obvious regressions in 412 and 437468, but regenerating the testbase-expects seems acceptable in this case.

For `VERS_CHECK`, though, I would assume that anything that's not part of the original set is HTML5 _only_ ([citation](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references)).

If I used (this document)[https://dev.w3.org/html5/html-author/charref] and change the doctype to an HTML4 doctype, they still render in my modern browsers, but is this behavior we can count on across the board?

geoffmcl:
This merge should be simultaneous with [PR 25](https://github.com/htacg/tidy-html5-tests/pull/25) in the tests repo...

Issue #439, opened 2016-07-15
Prevent tidy from quoting attributes


marcoscaceres:
There are cases in HTML5 where one wants to simply put the name of an attribute in a tag, and for it not to be set to anything (specially for proprietary attributes). It would be useful if there was an option to honor this behavior so that, for instance:

```HTML
<pre>
  <a idl>x</a>
</pre>
```

Does not become:

```HTML
<pre>
  <a idl="">x</a>
</pre>
```

If there is already some way to do that, apologies. I was not able to find an option for it.

geoffmcl:
@marcoscaceres, thanks for the comment... I searched, but did not find any `option` for this...

If seems if the attribute is **not** marked as a `boolean` type, ie has `CH_BOOL` bit, like say `nowrap`, `checked`, ... etc, that `=""` will unconditionally be added, if no value given...

So it seems a **new** `option` would be the only way to go to inhibit this behavior...

There is a README [OPTIONS.md](https://github.com/htacg/tidy-html5/blob/master/README/OPTIONS.md), which enumerates the basic four steps of a **new** option, the last just being its use in the code... 

But would ask you, or others, to comment on the use case for this... thanks

marcoscaceres:
> But would ask you, or others, to comment on the use case for this...

I'm trying to use tidy with a HTML-like markup language that we use to produce specs at the W3C (using a tool called [BikeShed](https://github.com/tabatkins/bikeshed/)). 

BikeShed allows unquoted proprietary attributes to appear in markup (e.g., `<pre idl>`, where idl means the contents of the `pre` is to be treated as "WebIDL" - which is a W3C thing for describing JavaScript bindings). There are several of these little attributes, and more could be added to BikeShed over time. 

Anyway, our primary use case is for tidy to leave those attributes alone and not quote them - so that when we `git diff` after Tidy, those attributes won't have changed. We are primarily interested in tidying the content of elements. 

Use case/problem is described in a bit more detail here:
https://github.com/tabatkins/bikeshed/issues/662

geoffmcl:
Since we are about to `release` **5.4**, and nothing has been done about this `Feature Request`, can only move the milestone to `next` **5.5**.

As always, further comments, patches, PR **very** welcome... thanks...

fyiman:
**marcoscaceres** - It sounds like you want a way to classify your attribute such as "**idl**" as a `boolean` type. I don't know if that's already possible, but it seems to me to be a way to do what you want.

[...]

PR #654, opened 2017-11-20
Added a new boolean-attributes configuration option.

balthisar:
_Don't merge this_ without a discussion at #439...


geoffmcl:
@balthisar see **more** discussion in #439... not sure this is needed...

geoffmcl:
@balthisar years later, and after re-reading all #439, still feel this is **not** needed, so closing it... thanks...

PR #684, opened 2018-03-03
Update language_en.h

jidanni:
Default: no means YES DO print out comments.
If set to YES, then DON'T print out comments.


geoffmcl:
@jidanni thanks for this PR... we like those... ;=))

This **reversal** of the text sense of this option looks **correct** to me...

Subject to any further comments, will merge this shortly... thanks...

Issue #704, opened 2018-03-27
Unescaped `&` emitted despite using **output-xhtml** key bindings in 5.6.0 in PHP bindings


TysonAndre:
This occurred when I upgraded from tidy-html5 5.4.0 to 5.6.0. I also upgraded php at the same time to 7.1.14, but php-src/ext/tidy has no recent changes

https://github.com/htacg/tidy-html5/issues/207#issuecomment-114830901 seems related. I don't see any discussion about how xhtml would be affected in #207 

http://validator.w3.org/check would warn about `hello & bye` as an XHTML fragment. If XHTML5 is HTML5 represented as XML, shouldn't that be automatically fixed by tidy-html5?

- If I'm missing any necessary options, please let me know

The below script reproduces this bug with PHP 7.1.9 (Running tidy-html5 5.4.0) and 7.1.14 (Built with tidy-html5 5.6.0)

```php
<?php
// Related to https://github.com/htacg/tidy-html5/issues/526 ?
const CONFIG = [
  'output-xhtml' => true,
  'show-errors' => false,
];
 
function print_tidy(string $html, array $config) {
	$tidy = new tidy();
	$tidy->parseString($html, $config, 'utf8');
	$tidy->cleanRepair();
	// errors are in $tidy->errorBuffer)
	$errors = [];
	if (!empty($tidy->errorBuffer)) {
	    $errors = array_merge($errors, explode("\n",$tidy->errorBuffer));
	}
	$clean = (string)$tidy;
	printf("In %s: %s (errors %s)\nconfig: %s\n", PHP_VERSION, $clean, json_encode($errors), json_encode($config));
}
print_tidy('hello & bye', CONFIG);
// EDIT: I've also tried print_tidy('<body>hello & bye</body>', CONFIG + ['preserve-entities' => true, 'quote-ampersand' => true, 'doctype' => 'strict', 'output-encoding' => 'ascii', 'input-encoding' => 'utf8']);
// but none of those additional options resulted in something different with the tidy-html5 bindings
/*
In 7.1.14: <!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<title></title>
</head>
<body>
hello & bye
</body>
</html> (errors [])
config: {"output-xhtml":true,"show-errors":false}
*/
/*
In 7.1.9: <!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<title></title>
</head>
<body>
hello &amp; bye
[...]

Issue #673, opened 2018-02-09
Unexpected behavior of 'add-xml-space' setting when used with 'wrap' => 0 and saveBuffer is called twice in tidy-html5 5.6.0


TysonAndre:
I'm using the PHP tidy wrapper(PHP 7.1.14) and the tidy HTML5 5.6.0 release

The behavior is different from 5.4.0. Tidy is inserting an excessive amount of newlines when those two options are used together.

Additionally, the behavior seems to contradict the documentation for `'tidy' => 0`

> This option specifies the right margin Tidy uses for line wrapping. Tidy tries to wrap lines so that they do not exceed this length. **Set wrap to zero if you want to disable line wrapping.**

It seems like https://github.com/htacg/tidy-html5/pull/645/files#diff-4a4e354609aef4a160784a5caa18e868R1710 might be a cause, but that's just an uninformed guess. (It looks like code that adds a newline).

```
php > $x = new Tidy();$x->parseString('<p>this is a string</p>', ['add-xml-space' => true, 'show-body-only' => true, 'wrap' => 50], 'utf8'); echo (string)$x;
<p>this is a string</p>
php > $x = new Tidy();$x->parseString('<p>this is a string</p>', ['add-xml-space' => true, 'show-body-only' => true, 'wrap' => 0], 'utf8'); echo (string)$x;
<p>
this
is
a
string</p>
```

---

A workaround is to set 'wrap' to an extremely large integer. This will continue to preserve newlines without breaking up long lines

```
$x = new Tidy();$x->parseString('<p>this is a string</p>', ['add-xml-space' => true, 'show-body-only' => true, 'wrap' => (1 << 30)], 'utf8'); echo (string)$x;
<p>this is a string</p>
```

geoffmcl:
@TysonAndre unfortunately I do **not** have PHP tidy wrapper (PHP 7.1.14) installed so I can not specifically test there...

But testing with command line tidy I can get the same `wrapped-to-multiple-lines`, as you show by using `-w 1` or `-w 2`... 

But when I use `-w 0` I get **no** wrapping as expected, and this is true for any version of tidy I tried, `5.4`, `5.6`, or current `5.7.3`...

So it seems the PHP Wrapper code that handles `'wrap' >= 0` is **wrong**. Rather than setting the wrap to `zero` it is setting it to `1`... or something...

To check the `wrap` option is otherwise ok, you could try `'wrap' >= 14`, that is `-w 14` and you should get two lines, like -

```html
[...]

PR #705, opened 2018-03-29
Is #673 - Revert 350f7b4 and 86e62db AdjustConfig logic

geoffmcl:
While these look like a code cleanup, they appear to have an adverse consequence in the use of libtidy by PHP 7+, so have been reverted. Changes in -

	modified:   src/config.c
	modified:   src/config.h
	modified:   src/tidylib.c

As indicated while these look like seemingly innocuous changes in AdjustConfig logic, it has caused  some very undesirable changes... see #704 for more details...

As stated in #673, unless a better fix is found will consider merging the `issue-673` to next...

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -303,6 +303,7 @@ static const struct {
 
 
 /* forward declarations */
+static void AdjustConfig( TidyDocImpl* doc );
 static Bool GetPickListValue( ctmbstr value, PickListItems* pickList, uint *result );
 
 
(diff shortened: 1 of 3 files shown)


geoffmcl:
This seems related to #780 - looks like a prob. with `wrap => 0`, which seemed broken in `5.6`, fixed in `5.7.9`, and so is in current `next`...

If someone had the time to cherry pick this fix back into the `5.6`, and issue a **5.6.1** release, that would be great...

But given that we have some difficulty in making a **new** release `5.8`, or `6.0`, given #743, that does not seem likely to happen... I don't know how to fix that... see stalled #741...

PR #712, opened 2018-04-04
Make global attribute `dir` accept auto as well.

doronbehar:
https://www.w3schools.com/tags/att_global_dir.asp

### DIFF
--- a/src/attrs.c
+++ b/src/attrs.c
@@ -144,7 +144,7 @@ static const Attribute attribute_defs [] =
   { TidyAttr_DATETIME,                "datetime",                CH_DATE      }, /* INS, DEL */
   { TidyAttr_DECLARE,                 "declare",                 CH_BOOL      }, /* OBJECT */
   { TidyAttr_DEFER,                   "defer",                   CH_BOOL      }, /* SCRIPT */
-  { TidyAttr_DIR,                     "dir",                     CH_TEXTDIR   }, /* ltr or rtl */
+  { TidyAttr_DIR,                     "dir",                     CH_TEXTDIR   }, /* ltr, rtl or auto */
   { TidyAttr_DISABLED,                "disabled",                CH_BOOL      }, /* form fields */
   { TidyAttr_DOWNLOAD,                "download",                CH_PCDATA    }, /* anchor */
   { TidyAttr_ENCODING,                "encoding",                CH_PCDATA    }, /* <?xml?> */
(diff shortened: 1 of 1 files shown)


geoffmcl:
@doronbehar thank you for this PR... we like those...

But maybe this **fix** does **not** go far enough? Or maybe **too** far, depending on your perspective...

As far as I can quickly read in W3C documentation - note we do not normally accept `w3schools`, which as far as I know has no affiliation with the W3C - this `auto` value for the `dir` attribute was only introduced in HTML5... 

See [attribute-auto](https://www.w3.org/International/tests/repo/results/the-dir-attribute-auto), [dir-attribute](https://dev.w3.org/html5/spec-preview/global-attributes.html#the-dir-attribute), [dirauto](https://www.w3.org/International/questions/qa-html-dir#dirauto), and probably others references...

Since `tidy` works in legacy html4 **or** html5 mode depending on the `doctype` found, and none defaults to html5, maybe the `CheckTextDir` service should reflect this?

Maybe we could have something like (code **not** tested) -

```
/* checks dir attribute */
void CheckTextDir( TidyDocImpl* doc, Node *node, AttVal *attval)
{
    ctmbstr const values4[] = {"rtl", "ltr", NULL};
    /* Is #712 - add 'auto' for HTML5 */
    ctmbstr const values5[] = {"rtl", "ltr", "auto", NULL};
    CheckAttrValidity( doc, node, attval, 
       (TY_(IsHTML5Mode)(doc) ? values5 : values4) );
}
```

Or maybe there is a better way to do this **switching**...

What do you, or others, think? Thanks...

geoffmcl:
@doronbehar I have cycled around to shortly merging this PR, but still think the above patch is **required**...

It is only in a **HTML5** document that `auto` is valid...

Yes I could add this **after** the merge, but it seems better if this was part of your **fix**... thanks...

geoffmcl:
@doronbehar  now merged, and then patched as indicated... thanks...

geoffmcl:
@doronbehar no problem...

Have added 2 tests to **my** `tests` repo, **with** `dir="auto"` -

 1. [in_712.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_712.html) - a **HTML5** doc, that now passes `tidy` 5.7.11 plus, with no warnings. And also passes the W3C `nu` validator.
 2. [in_712-1.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_712-1.html) - a **HTML4** doc, which `tidy` 5.7.11, and most other versions, earlier or later, will still flag as a warning, as does the W3C `legacy` validator.
 
As a note to myself, I guess I should also add these to the **official** [regression](https://github.com/htacg/tidy-html5-tests) tests, as a continual reminder of the addition of `dir="auto"` in HTML5, unless someone beats me to it...

Another little step forward... thanks...

Issue #713, opened 2018-04-05
tidy's error messages should include filename somewhere


petdance:
When I run `tidy` from the command line on multiple files, I can't tell which file the errors apply to:

```
$ tidy -q -e *.html
line 343 column 35 - Warning: <img> attribute "height" has invalid value "auto"
line 276 column 11 - Warning: trimming empty <span>                                   
line 275 column 9 - Warning: trimming empty <button>                           
line 409 column 5 - Warning: moved <style> tag to <head>! fix-style-tags: no to avoid.
line 370 column 19 - Warning: <iframe> proprietary attribute "allow"
line 335 column 13 - Warning: <img> attribute "height" has invalid value "auto"
line 269 column 11 - Warning: trimming empty <span>
line 268 column 9 - Warning: trimming empty <button>
line 445 column 5 - Warning: moved <style> tag to <head>! fix-style-tags: no to avoid.
```

Running without the `-q` doesn't show me the filename either.
```
$ tidy -e *.html                    
line 343 column 35 - Warning: <img> attribute "height" has invalid value "auto"       
line 276 column 11 - Warning: trimming empty <span>                              
line 275 column 9 - Warning: trimming empty <button>
line 409 column 5 - Warning: moved <style> tag to <head>! fix-style-tags: no to avoid.
line 370 column 19 - Warning: <iframe> proprietary attribute "allow"
Info: Document content looks like HTML5                 
Tidy found 5 warnings and 0 errors!                                            
                                                        
line 335 column 13 - Warning: <img> attribute "height" has invalid value "auto"
line 269 column 11 - Warning: trimming empty <span>     
line 268 column 9 - Warning: trimming empty <button>    
line 445 column 5 - Warning: moved <style> tag to <head>! fix-style-tags: no to avoid.
Info: Document content looks like HTML5            
Tidy found 4 warnings and 0 errors!  
```

ler762:
yes, it would be nice if the ending
`   Tidy found NNN warnings and NNN errors!`
was changed to
`   Tidy found NNN warnings and NNN errors in file FILENAME`
and the -q option was changed to always spit out the warnings & errors summary line

but until tidy is changed, a work-around is

```
for FNAME in `find . -name "*.html"` ; do
    echo " ---- begin processing $FNAME" >&2
    tidy -q -e $FNAME
done
```

geoffmcl:
[...]

PR #715, opened 2018-04-10
update --gnu-emac option documentation

ler762:
make it a bit clearer what the --gnu-emac option does and give an
example of tidy output with and without the option.


geoffmcl:
@ler762 yup, I thought about mentioning that separation, but this being your **first** PR, which we are **always** thankful for, decided to let it go in this case... will accept it as you like... 

Hope you offer many more PR's, but watch-out **next** time ;=))

And due to my workflow, prefer comments in the issue #713 this started from, but again no problem really... 

Anywhere is better than none ;=))

ler762:
hopefully the last change for this PR

geoffmcl:
@ler762 have now merged #717 which I think contains all the same patches - I hope - so closing this...

PR #754, opened 2018-09-10
change the processing of init files

ler762:
- add a new "-no-config" command line option to not read any initialization file

- refactor init file processing so that only one file is read
The search order is:
 1. command line "-config xxx" or "-no-config" options
 2. environment variable HTML_TIDY=xxx
 3. ~/.tidyrc
 4. /etc/tidy.conf

Issue #752, opened 2018-08-31
mute in ~/.tidyrc runs fine but triggers exit(1)


dse:
If I specify a value for the `mute` option in `~/.tidyrc`, for example:

    mute: MISSING_DOCTYPE

This causes `tidy` to work normally, observing the `mute` option I specified, but it also causes `tidy` to output the following error message and exit with a return code of 1:

    Loading config file "~/.tidyrc" failed, err = 1

Specifying `--mute MISSING_DOCTYPE` on the command line instead of specifying it in `~/.tidyrc` does not trigger this problem.

Reproducing is simple enough.  Run the following command with `mute: MISSING_DOCTYPE` in `~/.tidyrc`:

    echo '<html><head><title>foo</title></head><body></body></html>' | tidy

Then take that line out of `~/.tidyrc` then run the following:

    echo '<html><head><title>foo</title></head><body></body></html>' | tidy --mute MISSING_DOCTYPE

geoffmcl:
@dse had to build a special Windows tidy, add `-DENABLE_CONFIG_FILES:BOOL=YES`, and then enhance the `ExpandTilde(...)` service to see this... 

That problem is that the service used to read the config file, in `TY_(DefineMutedMessage(...)` calls **Report**

```
    /* Must come *after* adding to the list, in case it's muted, too. */
    TY_(Report)( doc, NULL, NULL, STRING_MUTING_TYPE, name );
```

This will bump the `case TidyConfig: doc->optionErrors++; break;` and will output the `Loading config file "~/.tidyrc" failed, err = 1`...

This it the patch to begin the **fix** in Windows -

```diff
diff --git a/src/config.c b/src/config.c
index f5e8379..085f4b6 100644
--- a/src/config.c
+++ b/src/config.c
@@ -877,8 +877,26 @@ static ctmbstr ExpandTilde( TidyDocImpl* doc, ctmbstr filename )
     if (filename[1] == '/')
     {
         home_dir = getenv("HOME");
-        if ( home_dir )
+        if (home_dir) {
             ++filename;
+        }
+#ifdef _MSC_VER 
+        else if (strlen(filename) >= 3) {   /* at least '~/+1' */
+            /* no HOME env in Windows - got for HOMEDRIVE=C: HOMEPATH=\Users\user */
+            char * hd = getenv("HOMEDRIVE");
+            char * hp = getenv("HOMEPATH");
+            if (hd && hp) {
[...]

PR #755, opened 2018-09-16
Aslr

ler762:
having optimization, DEP and ASLR enabled if building with gcc seems like a Good Thing

I thought I'd made two separate branches
processrc for changing how init files were handled
aslr  for enabling DEP and ASLR

but it looks like I fucked it up yet again & both are in this pull request :(

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,11 +197,27 @@ if(WIN32 AND MSVC)
     set( CMAKE_DEBUG_POSTFIX "d" ) # little effect in unix
 else()
     # add any gcc flags
+
+    set ( CFLAGS "${CFLAGS} -O2")
+# -O2: GCC performs nearly all supported optimizations that do not involve
+# a space-speed tradeoff.  As compared to '-O', this option increases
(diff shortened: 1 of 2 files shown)


geoffmcl:
@ler762 yes, seems you created the `aslr` branch from the `processrc` branch, so naturally will contain all... to keep them separate you must start with `next` when creating a new branch...

```
$ git checkout next
$ git checkout -b newbranch
```

Assuming you can separate these two issues... meantime this dual PR should be closed, since the `processrc` #754 is not yet agreed... see comments in #752, specifically [PR](https://github.com/htacg/tidy-html5/issues/752#issuecomment-421813350)

Concerning just commit 66b7ed544 not sure these should be added for every non-win os, and every version of `gcc` out there... 

Seems there should be a way to do this externally, without forcing everyone down the same road... deciding for them...

Maybe if you opened an separate issue to discuss this... and first to see and understand what cmake uses by default... if not already included, why cmake chooses otherwise... what advantages... etc, etc, etc, ...

I would certainly need much more feedback on what is proposed... and prefer an issue to discuss... thanks...

Issue #778, opened 2018-11-21
regression tests fail if /etc/tidy.conf or ~/.tidyrc exists


ler762:
Just doing ``touch ~/.tidyrc`` or ``touch /etc/tidy.conf`` is enough to make some the regression tests fail.

```
$ ls -l /etc/tidy.conf ~/.tidyrc
ls: cannot access '/etc/tidy.conf': No such file or directory
ls: cannot access '/home/Lee/.tidyrc': No such file or directory
 $ tidy -q -export-config | grep -i indent
indent: no
indent-attributes: no
indent-cdata: no
indent-spaces: 2
indent-with-tabs: no
$ touch ~/.tidyrc
$ tidy -q -export-config | grep -i indent
indent: no
indent-attributes: no
indent-cdata: no
indent-spaces: 0
indent-with-tabs: no
```

If tidy is going to have multiple init files it needs to wait until all are read before calling AdjustConfig.
Or not call AdjustConfig at all & make the user responsible for specifying a consistent set of options.
Or fix AdjustConfig to not bork the config.
Or .. what?

Not calling AdjustConfig is enough to have all the regression tests pass if /etc/tidy.conf and/or ~/.tidyrc exist:

```
$ git diff
diff --git a/src/config.c b/src/config.c
index f5e8379..67fc804 100644
--- a/src/config.c
+++ b/src/config.c
@@ -1056,8 +1056,6 @@ int TY_(ParseConfigFileEnc)( TidyDocImpl* doc, ctmbstr file, ctmbstr charenc )
     if ( fname != (tmbstr) file )
         TidyDocFree( doc, fname );

-    AdjustConfig( doc );
-
     /* any new config errors? If so, return warning status. */
     return (doc->optionErrors > opterrs ? 1 : 0);
 }
```

geoffmcl:
@ler762 thank you for adding this issue... this takes over from [Tests Issue 30](https://github.com/htacg/tidy-html5-tests/issues/30)...

And showing an easy way to `see` some of what config has changed, just by calling `AdjustConfig( doc );`... this goes a long way to explain the `test` anomalies, like missing `indent`... thanks...

But it is **not** just an issue of our `regression` tests! As expressed elsewhere, they are code `testers` after all, and should ensure **all** environment and runtime configs are **removed**... They should **never** see the problem... but...

This could yield **strange** results for `any` tidy user, anytime... and needs to be addressed. It is a **BUG**!

[...]

PR #779, opened 2018-11-22
delay calling AdjustConfig

ler762:
Wait until after reading all the init files and parameters before calling AdjustConfig

Calling AdjustConfig right after reading an init file can break things since
tidy might not have the all of the config settings yet.  It needs to wait until
all the init files and parameters have been processed so that it can make it's
adjustments based on the full config.
ref: https://github.com/htacg/tidy-html5/issues/778


geoffmcl:
@ler762 closing this... do not agree with the need to **wait** - Why? nor with **can break things** - What things? 

Please explain if you think this should go further, hopefully in an **issue** that addresses **these things** ;=)) and not in a PR!

But also assume you only used it to show some code, but would always prefer that be attached, or inline patches, in the comment... thanks

PR #782, opened 2018-11-29
Is. #781 - Drop redundant blank lines in -help

### LINKED ISSUE #781: Redundant blank lines when printing -help
Refer the following screenshot:

default

There are two redundant blank lines before the actual -help content (in 5.2.0 there was only one but now there are two), which is not pretty IMO.

I would suggest keep no blank lines at all or only one blank line, if there's any reason to do so.

Lin-Buo-Ren:
Fixes #781.


geoffmcl:
@Lin-Buo-Ren looks good to me... let's see if we get anymore feedback in #781 ...

Issue #761, opened 2018-10-10
AddByte allocAmt overflows for large input files


destroyhimmyrobots:
Given a 2.8 gigabyte input file, the function `AddByte` (https://github.com/htacg/tidy-html5/blob/next/src/lexer.c#L949) will enter an infinite loop when called from `prvTidyAddCharToLexer` on both modern Linux & Darwin systems.

This is likely a result of the allocation strategy (multiplying by 2) and because the `uint` type used to define the `allocAmt` variable is an unsigned 32-bit integer on these systems. For example, the `sys/types.h` header on one system defines that type as `unsigned int`: https://github.com/apple/darwin-xnu/blob/master/bsd/sys/types.h#L92

The initial lexer state when the problem surfaces looks like this:

    lexer->lexsize = 2147483646
    lexer->lexlength = 2147483648
    allocAmt = 0

Eventually, in my debugger it shows the value of `allocAmt` wrapping to 0 after reaching

    allocAmt = 2147483648

at https://github.com/htacg/tidy-html5/blob/next/src/lexer.c#L955 when trying to increase the buffer by one more factor of two. The result overflows uint32 by one. 

One solution may be to make the `allocAmt` a 64-bit integer type.

I searched for some alternative APIs in http://api.html-tidy.org/tidy/tidylib_api_5.6.0/group__IO.html, but it is not clear if these would solve this overflow issue.

geoffmcl:
@destroyhimmyrobots wow, playing with HTML files in the 2.8 gigabyte plus range, certainly points out a `design` **limit** of libtidy... thanks for this...

At present all `text` kept by libtidy is in an allocated for **each** document, `lexer` byte buffer, and thus nodes holding this `text` are maintained just by keeping the `start` and `end` offset into this buffer of the text... a `uint`, all presently 32-bit in most systems...

The present method of just **doubling** the last allocation value, only allows for some 20 reallocation, up to a maximum of 2,147,483,648 byte, 0x80000000, and as you point out the next **double** gives us ZERO, 0, leading to a loop forever lock **bug**...

Yes, one thought would be to use 64-bit offsets, but that would involve changing quite a lot of code... each time and place where the `start` and `end` offsets are stored and used... it is certainly possible... but first to explore other alternatives as well...

[...]

PR #784, opened 2018-12-01
Is. #761 - Double capacity, and remove loop lock - still a 4GB 'uint'

geoffmcl:
Will not exactly **fix** #761, but seems it should improve the situation...


geoffmcl:
Eventually got to fully `test` this, and something is **WRONG**!

In the first place it seems to increase the time to process **large** files, by more than a factor of 10!

It goes without saying this should **not** be merged until **fully** explored...

See issue #761 for further comments...

PR #823, opened 2019-05-08
added OS __ANDROID__ in tidyplatform.h

naveedpash:


geoffmcl:
@naveedpash Sorry for being so slow... @da2x thanks for the push on this...

PR #825, opened 2019-05-10
Small improvements.

MrSorcus:


geoffmcl:
@MrSorcus thank you for your interest in `tidy`... the time... and this PR... thanks...

The `Change "tidyLocalMapItem" to "tidyLocaleMapItem"` cd58c31 looks good... no probs... thank you..

The `Change some functions name to lower case.` 83e8b35 is **NOT** so **small**... sorry...

That is a published 5.6.x, or before, API change, and that is **serious**... 

If I understand correctly, such a change requires a bump in the major shared library version number... presently `5.x.x` to `6.x.x`... etc...

And while I like, and agree with, the `case` correction, I do **not** think it is worth such a major version bump... at this time...

It should however be kept in mind, if/when such a major bump is on hand... for other reasons...

So that second commit needs to be **removed** from the PR... requires great **git** shui...

**OR** this PR closed, and a **new** PR opened, usually from a forked **branch**... containing only the first change... thanks...

OT: github issue: It is **not** good for you, even if this PR was **all** accepted... because it is in your `MrSorcus:next` branch... ugh!

If/when this were to be merged to `htacg:next`, you can have problems, when updating your `fork`... conflicts...

It is **better** you do such changes in a new `branch`... like `$ git checkout -b newbranch`... do changes, commit,... then `$ git push -u origin newbranch`... the convention is to use `issue-XXXX` branch, but not important... and base your PR on that branch...

MrSorcus:
I reopen PR with only first commit.
But what about second commit?

geoffmcl:
@MrSorcus thank you for the new PR #829, addressing the **docs** issue... this will be merged in due course... just as a `docs` correction...

Concerning the API changes, this could be in another branch, like say `api-6.0.0`, or something... that is if my assumption that such API changes must be in a new major version... seek discussion, confirmation of that... but they should somehow be preserved... thanks...

geoffmcl:
@MrSorcus first part - docs fix - done...

As stated, we need to remember the API change, at the appropriate time... thanks

Issue #550, opened 2017-05-09
Compile in Windows XP/Vista, with MSVC8 2005 and MSVC10 2010


geoffmcl:
This is **not** a **bug**, and **not** really a `Feature Request` so am labeling is just `Technical Support`!

While I am **not** suggesting we should **always** try to support the MSVC compilers back to 2005, on the other hand Tidy C code is **not** complicated, and while such support involves only a few changes, and some small code re-arrangement, one has to ask **why not**?!

Like GCC, for a long time now, MSVC14 2015, allows some minimal breaks to what I understand is **pure** K&R C, these older MS compilers did **not**! In essence that means **all** variables, in any given context, **must** be declared **before** any code statements... Not a very difficult thing to do, but is easily forgotten, given that no warning or error is issued by modern compilers...

I have searched for a GCC option to put back this warning... have tried things like `-pedantic` and/or `-std=x99`, and while this generated other warnings, nothing about say the following in tidy.c - 

```
    tmbstr locale = NULL;
    tidySetMessageCallback( tdoc, reportCallback); /* experimental group */

    uint contentErrors = 0;
    uint contentWarnings = 0;
    uint accessWarnings = 0;
```

This will cause an error in MSVC8 and MSVC10, **but** it only needs to be re-arranged to - 

```
    tmbstr locale = NULL;
    uint contentErrors = 0;
    uint contentWarnings = 0;
    uint accessWarnings = 0;
    
    tidySetMessageCallback( tdoc, reportCallback); /* experimental group */
```

and all is well, which to me looks neater code anyway...

A second **stupid** thing is whether a standard library type function should **have** a leading underscore, or **not**. And in this case it seems MS can be optional in some cases, **not** in others... example, again in `tidy.c` -

```
        snprintf(buffer, sizeBuffer, "%s:", tidyOptGetName( topt ));
```

MSVC14 has **no** problem with this, but again MSVC8/10 will first warns that it is undefined, but assumes it is external, but then when we come to the link, will report it as missing external **error**... the fix is sloppy, but works -

#if (defined(_MSC_VER) && (_MSC_VER < 1900))
#define snprintf _snprintf
#endif

Anyway, I am preparing a small set of patches to address each of these small issues, and will present them shortly... 

Some can be overcome by suppressing the MSVC warnings, and we already do that for 4996 4090 4244 4267, and will continue to add more, if need be, so this really **only** addresses those that generate an **error**, and thus stop the compile/link!

[...]

PR #843, opened 2019-10-01
Updated the c code, initalized a var argumentc

duttrohan0302:


geoffmcl:
@duttrohan0302 thanks for the issue, but do **not** understand... please explain...

```diff
@@ -352,8 +353,8 @@ int main(int argc, char **argv)

         --argc;
         ++argv;
-
-        if (argc <= 1)
+        int argumentc = argc
+        if (argumentc <= 1)
            break;
    }
```

As the patch stands at the moment, there seems no reason to create a **new** stack variable, `int argumentc`, copy the current `argc` stack variable to it, and test it for exit/break... over testing `argc` itself... 

What does this gain? What bug is it addressing?

That is aside from the fact that creating this **new** variable in the middle of **code** breaks the general `K&R` C rule, that all stack variable should be created at the beginning of a context... before all code statements (see maybe #550)... but this is minor...

And the given code seems **incomplete** - `int argumentc = argc` - will **not** compile! Did you missing adding a `;`?

Additionally, github shows `@duttrohan0302 wants to merge ... from unknown repository`, which usually indicates that the forked repository has been deleted... or changed... or whatever... so what is the source of this PR?

Am I missing something? 

Am closing this until, at least some, of the above questions are addressed... thanks...

OT: In general, a PR should **only** be created follow a `bug|feature request|...` issue, which explains the reasons, the details, the RFC,... and allows discussion on its merits/testing/need/use/etc/etc ... before presenting code... but there are exceptions...

Issue #310, opened 2015-11-18
Do not mess up Jinja templates


hermanbergwerf:
`tidy` messes up Jinja templates. Would it be possible to add an option to format Jinja templates correctly?

balthisar:
Per our [FAQ](http://www.html-tidy.org/documentation/#part_faq), can you provide the following:

- A description of what you think went wrong.
- The HTML Tidy version (find it out by running tidy -v), and operating system you are running.
- The input that exposes the bug. A small HTML document that reproduces the problem is best.
- The configuration options you’ve used. Command line options like -asxml, configuration files, etc. - - You may use tidy -show-config to get an overview of the active Tidy settings.

Even if we try to figure out what a Jinja template is, there's no guarantee that we'll have anything go wrong, so a short, failing example posted here would be useful.

hermanbergwerf:
Ok, sorry, I like brief descriptions myself and I thought Jinja was kind of common knowledge (Jinja is the backbone of Sphinx, Jekyll, etc.). Also, could you put the contributing guidelines in CONTRIBUTING.md (https://github.com/blog/1184-contributing-guidelines) Jinja is a template engine written in Python. See http://jinja.pocoo.org/.

# What went wrong
Sphinx (and Jekyll) templates are written using Jinja. A template might look something like this:
```html
{% extends "layout.html" %}
{% block body %}
  <ul>
  {% for user in users %}
    <li><a href="{{ user.url }}">{{ user.username }}</a></li>
  {% endfor %}
  </ul>
{% endblock %}
```
I created an auto-formatter for my project to prevent messed up indents etc. I picked `tidy-html5` as HTML formatter. When I formatted a repository that contained Sphinx templates (http://sphinx-doc.org/, this is a documentation generator written in Python that uses Jinja as template backend), it went wrong.

# Tidy version
5.0.0

# Reproduce
```
{% extends "layout.html" %}
{% block body %}
  <ul>
  {% for user in users %}
    <li><a href="{{ user.url }}">{{ user.username }}</a></li>
  {% endfor %}
  </ul>
{% endblock %}
```
Tidy using standard configuration:
```
<!DOCTYPE html>
<html>
<head>
  <meta name="generator" content=
  "HTML Tidy for HTML5 for Linux version 5.0.0">
  <title></title>
</head>
<body>
  {% extends "layout.html" %} {% block body %}
  <ul>
    <li style="list-style: none">{% for user in users %}</li>
    <li>
[...]

Issue #282, opened 2015-10-05
tidy validates incorrectly inside of <template>


shwoodard:
Templates should be validated with different semantics.  IOW, if I am templating an `<li>`, I should not be warned that an `<ul>` will be implicitly inserted.

## Template html src
```html
<template id="tmpl__lane">
  <li class="lane">
    <h2></h2>
    <ol class="stories"></ol>
  </li>
</template>
```

## Error on line one:
`missing </template> before <li>`

## Error on line 2
`inserting implicit <ul>`

## Error on line three
`trimming empty <h2>`

imho, none of these errors should appear due that we are inside of a `<template>`.

geoffmcl:
@shwoodard thanks for opening this issue...

I guess when html5 was added to tidy, the use of the `template` tag was still in **draft** form, so in tidy it is parsed as any other block element... 

Has the W3C spec moved on since then? 

Maybe some with W3C knowledge can comment, and maybe provides some links to read...

To check browser support I scraped a working example from [whatwg.org](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element) - thanks to them -

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>Test 2 - template tag #282 - Cat data</title>
<script>
 // Data is hard-coded here, but could come from the server
 var data = [
   { name: 'Pillar', color: 'Ticked Tabby', sex: 'Female (neutered)', legs: 3 },
   { name: 'Hedral', color: 'Tuxedo', sex: 'Male (neutered)', legs: 4 },
 ];
</script>
<table>
 <thead>
  <tr>
   <th>Name <th>Colour <th>Sex <th>Legs
 <tbody>
  <template id="row">
   <tr><td><td><td><td>
  </template>
</table>
<script>
 var template = document.querySelector('#row');
 for (var i = 0; i < data.length; i += 1) {
   var cat = data[i];
   var clone = template.content.cloneNode(true);
   var cells = clone.querySelectorAll('td');
   cells[0].textContent = cat.name;
   cells[1].textContent = cat.color;
   cells[2].textContent = cat.sex;
   cells[3].textContent = cat.legs;
   template.parentNode.appendChild(clone);
 }
</script>
```

This seems to function fine in chrome and firefox, but **not** in IE, Opera, or Safari, but I may not have all the latest windows 7 64-bit versions installed... especially since this [caniuse.com](http://caniuse.com/#feat=template) report suggests better than I found...

[...]

Issue #199, opened 2015-04-10
HTML5 <template> tag not supported


lacombar:
```
line 11 column 1 - Error: <template> is not recognized!
line 11 column 1 - Warning: discarding unexpected <template>
line 14 column 1 - Warning: discarding unexpected </template>
2 warnings, 1 error were found!
```

geoffmcl:
@lacombar like #198, can you create a PR to keep your name on the commit?

geoffmcl:
@lacombar I just experimented in a fresh tidy-html5 clone and clone of your aerilon repo into tidy-aerilon folder...

I can get all these changes into master by doing -

 1. cd tidy-html5
 2. git remote add aerilon ../tidy-aerilon
 3. git pull aerilon master
 4. git remote remove aerilon

But as suspected this does fetch and merge your #198 and #199 changes in one new commit authored by ME. Using a PR should keep you as the 'Author'

geoffmcl:
@lacombar pull request now merged...

When you get a chance please recheck and close if fixed...

Thanks...

geoffmcl:
@lacombar no further comments in over 3 weeks so assume fixed, and closing this issue...

Feel free to re-open, or open a new issue... thanks...

PR #848, opened 2019-11-28
Support the <slot> tag

lhchavez:
<slot> was added to the WHATWG HTML5 Living Standard on April 20, 2016.

### DIFF
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -1144,6 +1144,7 @@ typedef enum
   TidyAttr_SHOWGRIDX,              /**< SHOWGRIDX= */
   TidyAttr_SHOWGRIDY,              /**< SHOWGRIDY= */
   TidyAttr_SIZE,                   /**< SIZE= */
+  TidyAttr_SLOT,                   /**< SLOT= */
   TidyAttr_SPAN,                   /**< SPAN= */
   TidyAttr_SRC,                    /**< SRC= */
   TidyAttr_SRCSET,                 /**< SRCSET= (HTML5) */
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -988,6 +988,7 @@ typedef enum
   TidyTag_OUTPUT,        /**< OUTPUT */
   TidyTag_PROGRESS,      /**< PROGRESS */
   TidyTag_SECTION,       /**< SECTION */
+  TidyTag_SLOT,          /**< SLOT */
   TidyTag_SOURCE,        /**< SOURCE */
   TidyTag_SUMMARY,       /**< SUMMARY */
   TidyTag_TEMPLATE,      /**< TEMPLATE */
--- a/src/attrdict.c
+++ b/src/attrdict.c
@@ -2880,6 +2880,17 @@ const AttrVersion TY_(W3CAttrsFor_SMALL)[] =
   { TidyAttr_UNKNOWN,               0                                                                          },
 };
 
+const AttrVersion TY_(W3CAttrsFor_SLOT)[] =
+{
+  INCLUDE_ARIA
+  INCLUDE_MICRODATA
+  { TidyAttr_NAME,                  xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 },
+  INCLUDE_CORE_ATTRIBS
+  INCLUDE_CORE_EVENTS
+  INCLUDE_RDFA
+  { TidyAttr_UNKNOWN,               0                                                                          },
+};
+
 const AttrVersion TY_(W3CAttrsFor_SOURCE)[] = 
 {
   INCLUDE_ARIA
--- a/src/attrdict.c
+++ b/src/attrdict.c
@@ -95,6 +95,7 @@
   { TidyAttr_IS,                    xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
   { TidyAttr_LANG,                  xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
   { TidyAttr_ROLE,                  xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
+  { TidyAttr_SLOT,                  xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
   { TidyAttr_SPELLCHECK,            xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
   { TidyAttr_STYLE,                 xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
   { TidyAttr_TABINDEX,              xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 }, \
--- a/src/attrdict.h
+++ b/src/attrdict.h
@@ -142,6 +142,7 @@ extern const AttrVersion TY_(W3CAttrsFor_OUTPUT)[];
 extern const AttrVersion TY_(W3CAttrsFor_MENUITEM)[];
 extern const AttrVersion TY_(W3CAttrsFor_METER)[];
 extern const AttrVersion TY_(W3CAttrsFor_PROGRESS)[];
+extern const AttrVersion TY_(W3CAttrsFor_SLOT)[];
 extern const AttrVersion TY_(W3CAttrsFor_TEMPLATE)[];
 extern const AttrVersion TY_(W3CAttrsFor_TIME)[];
 extern const AttrVersion TY_(W3CAttrsFor_DATA)[];
--- a/src/attrs.c
+++ b/src/attrs.c
@@ -238,6 +238,7 @@ static const Attribute attribute_defs [] =
   { TidyAttr_SHOWGRIDX,               "showgridx",               CH_BOOL      }, /* TABLE Adobe golive*/
   { TidyAttr_SHOWGRIDY,               "showgridy",               CH_BOOL      }, /* TABLE Adobe golive*/
   { TidyAttr_SIZE,                    "size",                    CH_NUMBER    }, /* HR, FONT, BASEFONT, SELECT */
+  { TidyAttr_SLOT,                    "slot",                    CH_PCDATA    },
   { TidyAttr_SPAN,                    "span",                    CH_NUMBER    }, /* COL, COLGROUP */
   { TidyAttr_SRC,                     "src",                     CH_URL       }, /* IMG, FRAME, IFRAME */
   { TidyAttr_SRCSET,                  "srcset",                  CH_PCDATA    }, /* IMG (HTML5) */
--- a/src/attrs.h
+++ b/src/attrs.h
@@ -315,6 +315,7 @@ Bool TY_(AttributeIsMismatched)(Node* node, AttVal* attval, TidyDocImpl* doc);
 #define attrIsSHOWGRIDX(av)         AttrIsId( av, TidyAttr_SHOWGRIDX  )
 #define attrIsSHOWGRIDY(av)         AttrIsId( av, TidyAttr_SHOWGRIDY  )
 #define attrIsSIZE(av)              AttrIsId( av, TidyAttr_SIZE  )
+#define attrIsSLOT(av)              AttrIsId( av, TidyAttr_SLOT  )
 #define attrIsSPAN(av)              AttrIsId( av, TidyAttr_SPAN  )
 #define attrIsSRC(av)               AttrIsId( av, TidyAttr_SRC  )
 #define attrIsSTANDBY(av)           AttrIsId( av, TidyAttr_STANDBY  )
--- a/src/tags.c
+++ b/src/tags.c
@@ -323,7 +323,7 @@ static Dict tag_defs[] =
   { TidyTag_OUTPUT,      "output",       VERS_ELEM_OUTPUT,      &TY_(W3CAttrsFor_OUTPUT)[0],      (CM_INLINE),                   TY_(ParseInline),    NULL           },
   { TidyTag_PROGRESS,    "progress",     VERS_ELEM_PROGRESS,    &TY_(W3CAttrsFor_PROGRESS)[0],    (CM_INLINE),                   TY_(ParseInline),    NULL           },
   { TidyTag_SECTION,     "section",      VERS_ELEM_SECTION,     &TY_(W3CAttrsFor_SECTION)[0],     (CM_BLOCK),                    TY_(ParseBlock),     NULL           },
-  { TidyTag_SLOT,        "slot",         VERS_ELEM_SLOT,        &TY_(W3CAttrsFor_SLOT)[0],        (CM_INLINE),                   TY_(ParseInline),    NULL           },
+  { TidyTag_SLOT,        "slot",         VERS_ELEM_SLOT,        &TY_(W3CAttrsFor_SLOT)[0],        (CM_BLOCK|CM_INLINE|CM_MIXED), TY_(ParseBlock),     NULL           },
   { TidyTag_SOURCE,      "source",       VERS_ELEM_SOURCE,      &TY_(W3CAttrsFor_SOURCE)[0],      (CM_BLOCK|CM_INLINE|CM_EMPTY), TY_(ParseBlock),     NULL           },
   { TidyTag_SUMMARY,     "summary",      VERS_ELEM_SUMMARY,     &TY_(W3CAttrsFor_SUMMARY)[0],     (CM_BLOCK),                    TY_(ParseInline),    NULL           },
   { TidyTag_TEMPLATE,    "template",     VERS_ELEM_TEMPLATE,    &TY_(W3CAttrsFor_TEMPLATE)[0],    (CM_BLOCK),                    TY_(ParseBlock),     NULL           },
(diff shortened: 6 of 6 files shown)


geoffmcl:
@lhchavez thanks for the Feature Request...

I pulled your fork, built a 5.7.28.I848 version... created some sample html... it seems great on some things... but puzzled by a few things...

I had added `<style>` element, in the `template`, and found that I needed `--fix-style-tags no` to the config... 

Then still got a warning - `moved <style> tag to <head>! fix-style-tags: no to avoid.` - hmmm, I said I understood.. why a continued nag...

Then from the code -
```html
  <div>
    <span slot='title-1'>Title A</span> 
    <span slot='title-2'>Title B...
   ...
```

got a warning - `<span> proprietary attribute "slot"` - but maybe this requires an addition to the span tag....

On another sample -

```
<p><slot name="my-text">My default text</slot></p>
```

Tidy outputs -
```
<p></p>
<slot name="my-text">My default text</slot>
<p></p>
```

Just thinking aloud, should `slot` have an `inline` bit, too,... but maybe the code is wrong in other ways... the `nu` html checker reports - `Error: Element slot not allowed as child of element p in this context.` - but maybe this is a different issue... still deciding...

With those niggles discussed, solved, added... looks like a good PR... thanks...

To continue testing, could you supply some, hopefully small sample html to test, explore, etc... thanks

lhchavez:
Done, added one testcase that has some snippets pulled from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot and should now pass cleanly.

lhchavez:
the one thing i could not find an easy way to do is to allow both `<template>` and `<slot>` to include any arbitrary HTML content (since they both have a [transparent content model](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Transparent_content_model)). that's what causing the move of the `<style>` to `<head>`.

I'm using `ParseInline` for `<slot>` in this case, but that's not the correct choice, since the other example in MDN (`<slot name="attributes"><p>None</p></slot>`) does not work. `ParseBlock` does some other weird things since the code does not expect it to be called from within an inline context.

geoffmcl:
@lhchavez thank you for the additional pushes to this PR, but not sure I agree with all, yet...

But pulled and built `tidy-rc848-1` for continued testing...

And thanks for the `case-848.html` test sample, from MDN docs...

There, in [PR-32](https://github.com/htacg/tidy-html5-tests/pull/32), suggest a small enhancement, to remove some `nu` nags -

```diff
diff --git a/cases/testbase/case-848.html b/cases/testbase/case-848.html
index d74c3ca..2ff2bf9 100644
--- a/cases/testbase/case-848.html
+++ b/cases/testbase/case-848.html
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
-  <title>test</title>
+<meta charset="utf-8">
+<title>add 'slot' tag</title>
 </head>
 <body>
   <p><slot name="my-text">My default text</slot></p>
```

But these are not particularly important... 

Question on `<p><slot name="my-text">My default text</slot></p>`?

I do want to understand why `nu` issues the ERROR: `Element slot not allowed as child of element p in this context.` ... for this `case-848.html` source... while `libtidy` thinks it is ok... who is right, wrong, maybe,.. here...

I have a test file [in_848.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_848.html), which satisfies `nu`, and `tidy-rc848-1`, which contains code like `<th><slot name='title-1'></slot></th>`... just trying to understand...

Yes, **strongly** note you have switched `<slot>` from `CM_BLOCK/ParseBlock` to `CM_INLINE/ParseInline`... but still have some doubts about that... others have **both** bits... and use `ParseBlock`... and other combinations... what is most correct for `<slot>`... do not know yet...

And that raises the question about `<slot name="my-text"><p>My default text</p></slot>`, which I think would be valid in a `template`... need to try more examples/samples... help on those appreciated...

Using definitions by MDN, or even by W3C, for that matter, does little to understand the problem in the `libtidy` context... which started before such definitions came into style... but maybe they help describe the tag/attr... and thus help where to look...

I started to look back in the issues related to `template`, and found some - #199, #282, #310, ... and I am sure there are others... presently trying to wade through those...

Feel the PR has advanced, but still some `definition` questions, trial, tests, ...

Look forward to further feedback, comment, patches, PR, etc on this... thanks...

lhchavez:
> I do want to understand why `nu` issues the ERROR: `Element slot not allowed as child of element p in this context.`

This _might_ be a bug in `nu`, since it says that

> Contexts in which element slot may be used:
>  > Where phrasing content is expected.
>
> Content model for element p:
>  > Phrasing content.

Which sounds like it's contradicting itself.

> And that raises the question about `<slot name="my-text"><p>My default text</p></slot>`, which I think would be valid in a template...

yes, that should be valid in a template, but there is no ready-to-use solution to parse arbitrary HTML that should be allowed in `<template>`. Both `ParseBlock` and `ParseInline` are wrong in this case, but `ParseBlock` triggers fewer edge cases after manual inspection. I can definitely create a `ParseTransparent` that just calls `ParseBlock` and has a `TODO` comment explaining that it's a subpar solution that needs more work, linking to an issue about it.

Re: `CM_INLINE` vs `CM_BLOCK`/`CM_INLINE|CM_BLOCK|CM_MIXED`, I think that for `<slot>`, `CM_INLINE` is the better choice, since `<slot>` is only allowed in places where phrasing content is expected (per the error from `nu` above and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot)). Tidy's model does not match _exactly_, but the the vast majority of the tags in the [Phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) category are marked as only `CM_INLINE` in Tidy.

geoffmcl:
@lhchavez thank you for the feedback...

It seems we are 90% there, only differing on the `slot` def_tags definition...

You suggest `CM_INLINE` in 'libtidy` can be **equated** to `Phrasing content` in MDN...no, no, no...

I can **not** find `slot` in their list... lost... but that is not the point... that is MDN... 

Presently `libtidy` has **NO** concept of `Phrasing content`... maybe it could be added... somehow... if it can be defined...

> Both `ParseBlock` and `ParseInline` are wrong in this case

You further suggest a `ParseTransparent`, based on `ParseBlock`, with some `TODO` notes, but at the moment the PR uses `ParseInline`... an **important** choice...

Is what you suggest based on performance on `edge cases`... that smacks of trial and error... all for that... yeah ;=))

#### What happens 

What happens if the flag is say, `CM_INLINE|CM_BLOCK[|CM_MIXED]`, and the parser `ParseBlock` is used... 

Or some other combinations... 

How do they perform in the `main`, **AND** `edge cases`... not sure what that is...

Started an analysis of the current `tag_defs` table... to see if that would yield some `categorization` of the entries...  especially. `inline&|block` with `parser`... but that is SLOW going...

> RE: template... no ready-to-use solution to parse arbitrary HTML ...

Here, I tried to point to some earlier - **turn off tidy** - discussion... in general, it seems the consensis is - **This HTML Tidy** - what do you want it to do? 

There are many other text manipulation tools, to avoid `libTidy` seeing this funny html code, or `mess`... to tidy...

Parsing by `libTidy` is OFF, well, mostly OFF, when, parsing tags like `<script>`... 

Is it suggested that the `<template>` tag be treated as such... refs for that...

But that seems a **different**, and **NEW** issue... must refer to the previous discussions... thanks...

> This _might_ be a bug in `nu`, since it says that

Maybe, but maybe we do not understand, fully, those cute definitions of `Phrasing content`, et al... this time in `nu`... in a `libTidy` context... 

More importantly, how that can that be `simulated` in `libTidy`... if it is required...

#### But back to this great PR -

I would like to see some more **testing** of other table options... before deciding which is the best... will continue that... options/example/results testing appreciated...

Look forward to further feedback, comment, patches, PR, etc on this... thanks...

geoffmcl:
@lhchavez I have now conducted lots of further test, and concluded `slot` should use `ParseBlock`, in `libTidy` terms, with the bit flags of `inline|block|mixed`, and thus suggest the following additional patch to this PR -

```diff
diff --git a/src/tags.c b/src/tags.c
index 9f72a70..760c37b 100644
--- a/src/tags.c
+++ b/src/tags.c
@@ -323,7 +323,7 @@ static Dict tag_defs[] =
   { TidyTag_OUTPUT,      "output",       VERS_ELEM_OUTPUT,      &TY_(W3CAttrsFor_OUTPUT)[0],      (CM_INLINE),                   TY_(ParseInline),    NULL           },
   { TidyTag_PROGRESS,    "progress",     VERS_ELEM_PROGRESS,    &TY_(W3CAttrsFor_PROGRESS)[0],    (CM_INLINE),                   TY_(ParseInline),    NULL           },
   { TidyTag_SECTION,     "section",      VERS_ELEM_SECTION,     &TY_(W3CAttrsFor_SECTION)[0],     (CM_BLOCK),                    TY_(ParseBlock),     NULL           },
-  { TidyTag_SLOT,        "slot",         VERS_ELEM_SLOT,        &TY_(W3CAttrsFor_SLOT)[0],        (CM_INLINE),                   TY_(ParseInline),    NULL           },
+  { TidyTag_SLOT,        "slot",         VERS_ELEM_SLOT,        &TY_(W3CAttrsFor_SLOT)[0],        (CM_INLINE|CM_BLOCK|CM_MIXED), TY_(ParseBlock),     NULL           },
   { TidyTag_SOURCE,      "source",       VERS_ELEM_SOURCE,      &TY_(W3CAttrsFor_SOURCE)[0],      (CM_BLOCK|CM_INLINE|CM_EMPTY), TY_(ParseBlock),     NULL           },
   { TidyTag_SUMMARY,     "summary",      VERS_ELEM_SUMMARY,     &TY_(W3CAttrsFor_SUMMARY)[0],     (CM_BLOCK),                    TY_(ParseInline),    NULL           },
   { TidyTag_TEMPLATE,    "template",     VERS_ELEM_TEMPLATE,    &TY_(W3CAttrsFor_TEMPLATE)[0],    (CM_BLOCK),                    TY_(ParseBlock),     NULL           },
```

At first I tried to reason it out, analyzing the current `def_tags` definition table, using a `tidy-tags.pl` script, which outputs a html table version... 

See the [temptable.htm](http://geoffair.org/tmp/temptable.htm) using your `slot` branch, and [temptable2.htm](http://geoffair.org/tmp/temptable2.htm) using my `slot2` branch... **WARNING** These are only in my web `tmp` folder, and can be deleted/overwritten at any time... sorry...

But they do not seem to answer the question...

I then constructed more samples for testing... in addition to in_848.html, already published to my repo, added 848-1, 848-11, 848-12, 848-2, 848-3, and eventually 848-4... will try to get around to publishing these... but...

I built more **versions** of tidy... if we consider the table is made using 3 bit flags, inline (I), block (B), mixed (M), and 2 parsers, inline (I), and block, (B)... so built versions -

  - tidy-I-I.exe - your `slot` branch version - 1 inline flag, inline parser
  - tidy-IBM-B.exe - my `slot2` branch - 3 flags, block parser
  - then others - B-B, block/block, IB-B, add inline, IBM-I... etc
  
Then used each version of `tidy`, on each of my samples... the best run is using my IBM-B version... your I-I fails, creates bad, incompatible, html output, on several samples... 

Perhaps the most important is 848-4, which is based on the working, online sample, at  [MDN/web-components](https://github.com/mdn/web-components-examples/tree/master/element-details) ... with the addition of a config -

```
// issue 848-4
// new-blocklevel-tags: element-details
custom-tags: blocklevel
fix-style-tags: no
```

It also re-open the addition of the `fix-style-tags` option, but that can be a separate issue...

So I am now convinced the above patch needs to be applied to this PR... then feel it is good to go... at least after maybe a few tweaks, time, more tests...

Look forward to further feedback, comment, patches, PR, etc on this... thanks...

lhchavez:
> You suggest `CM_INLINE` in `libtidy` can be **equated** to`Phrasing content` in MDN...no, no, no...
> 
> I can **not** find `slot` in their list... lost... but that is not the point... that is MDN...

welp, that was my bad. i should have linked the canonical W3C definition of [Phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content), which _does_ contain the `<slot>` tag. MDN is a great resource, but it can go out of sync every now and then.

And I'm sorry if I sounded like I was suggesting that `CM_INLINE` was **equal** to the Phrasing content concept. I realize they are _slightly_ different, but it's the closest that is available at the moment. At least, it's very close _conceptually_ IIUC.

> Presently `libtidy` has **NO** concept of `Phrasing content`... maybe it could be added... somehow... if it can be defined...

IIUC the definifion is "if it is contained in the [canonical Phrasing content tag list](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content)".

> > Both `ParseBlock` and `ParseInline` are wrong in this case
> 
> You further suggest a `ParseTransparent`, based on `ParseBlock`, with some `TODO` notes, but at the moment the PR uses `ParseInline`... an **important** choice...

Errr, sorry, I got a bit confused with the `ParseBlock` bit, since I was thinking about `<template>`. For `<slot>`, the W3C spec says that it has a [Transparent content model](https://html.spec.whatwg.org/multipage/dom.html#transparent), which means that it inherits the content model of the container tag.

Here's where it gets tricky. `<slot>` can only be found in places where Phrasing content is allowed, but since Phrasing content is itself Flow content, this is allowed by the spec:

```html
<section>  <!-- this is Sectioning content, that has a Flow content model -->
  <slot>...</slot> <!-- this now has a Flow content model, so ParseBlock is more correct -->
  <p>  <!-- this is Flow content, that has a Phrasing content model -->
    <slot>...</slot> <!-- this now has a Phrasing content model, so ParseInline is more correct -->
  </p>
</section>
```

> Is what you suggest based on performance on `edge cases`... that smacks of trial and error... all for that... yeah ;=))

yeah :/ as I mentioned, both options are suboptimal and break *some* cases, as outlined above.

> 
> #### What happens
> What happens if the flag is say, `CM_INLINE|CM_BLOCK[|CM_MIXED]`, and the parser `ParseBlock` is used...

Orginally I made the decision of using `ParseInline` on a less exhaustive list of examples, but it looks like `ParseBlock` should let all the cases that the spec allows at the expense of also allowing things that the spec does not allow. But I guess it's an okay tradeoff.

> > RE: template... no ready-to-use solution to parse arbitrary HTML ...
> 
> Here, I tried to point to some earlier - **turn off tidy** - discussion... in general, it seems the consensis is - **This HTML Tidy** - what do you want it to do?

What I'm most interested in is the automatic rewriting feature of tidy, which we would lose by turning it off :(

geoffmcl:
@lhchavez thank you for the additional commit 99b6c660 ... this looks good...

Have not yet had a chance to update, rebuild, and test... etc... will try to get to that soonest... but the PR looks good...

I can see you are still trying to **tie** together the **two** sets of logic, W3C/MDM - Flow/Phrasing **vs** libTidy - ParseBlock/ParseInline... and do not quite **agree** with all the comments, in your sample... but, no problem...

It is important to have a reasonable sample set... and more welcome... actual use cases...

So along with the other 7, in my repo, in_848.html in_848-1.html in_848-11.html in_848-12.html in_848-2.html in_848-3.html in_848-4.html, added yours, as  [in_848-5.html](https://github.com/geoffmcl/tidy-test/blob/master/test/input5/in_848-5.html)... tks...

But am not sure what extra it adds... it is **passed** by both tidy-I-I.exe and the tidy-IBM-B.exe... so does not sway the argument either way... ParseBlock vs ParseInline, in `tidy` terms...

So, perhaps it is agreed, as it stands at the moment, if merged, `libTidy` can be **too** liberal with its acceptance of the tag `slot`... but maybe that could be left to a **new** issue, being raised later...

There is also the pending problem with the `style` tag, the `fix-style-tags` option, in in_848-4.html, but as stated, that can also be a **separate** issue... 

It seems downright **wrong** to move `style` to the `head`, in this instance... aside from also causing bad rendering...

I don't exactly understand to `turn off` the `automatic rewriting feature of tidy`... you will have to explain this more... if you want to persue this...

So after I get around to a bit more testing, I think this is a **good** addition, to `libTidy`... thanks...

Look forward to further feedback, comment, patches, PR, etc on this... thanks...

geoffmcl:
@lhchavez sorry for the long delay on this... I have subsequently changed PCs, and am in the process of setting up the testing of this again...

At least one quick change in the PR, is where we add the `TidyAttr_SLOT` enum, in `tidyenum.h`! The list `looks` alphabetic, but read #851... The SO version of the library would need to be bumped almost everytime such a new option is added... 

As further suggested there, I think the wording in OPTIONS.md, ATTRIBUTES.md, and probably others, needs to be `reverted` to state something like `a new option must be added just before the last...` ... maybe @balthisar could help with this?

Meantime, I will try to get back to testing `slot`... it is an **important** addition to libtidy...

Look forward to further feedback, comments, other patches, PR, etc, on this... thanks...

geoffmcl:
@lhchavez have build your forks `slot` branch, with the following patch, in line with #851...

```diff
diff --git a/include/tidyenum.h b/include/tidyenum.h
index 543a82e..8fe5f52 100644
--- a/include/tidyenum.h
+++ b/include/tidyenum.h
@@ -988,13 +988,13 @@ typedef enum
   TidyTag_OUTPUT,        /**< OUTPUT */
   TidyTag_PROGRESS,      /**< PROGRESS */
   TidyTag_SECTION,       /**< SECTION */
-  TidyTag_SLOT,          /**< SLOT */
   TidyTag_SOURCE,        /**< SOURCE */
   TidyTag_SUMMARY,       /**< SUMMARY */
   TidyTag_TEMPLATE,      /**< TEMPLATE */
   TidyTag_TIME,          /**< TIME */
   TidyTag_TRACK,         /**< TRACK */
   TidyTag_VIDEO,         /**< VIDEO */
+  TidyTag_SLOT,          /**< SLOT */
 
   N_TIDY_TAGS            /**< Must be last */
 } TidyTagId;
@@ -1144,7 +1144,6 @@ typedef enum
   TidyAttr_SHOWGRIDX,              /**< SHOWGRIDX= */
   TidyAttr_SHOWGRIDY,              /**< SHOWGRIDY= */
   TidyAttr_SIZE,                   /**< SIZE= */
-  TidyAttr_SLOT,                   /**< SLOT= */
   TidyAttr_SPAN,                   /**< SPAN= */
   TidyAttr_SRC,                    /**< SRC= */
   TidyAttr_SRCSET,                 /**< SRCSET= (HTML5) */
@@ -1344,7 +1343,8 @@ typedef enum
   TidyAttr_AS,                     /**< AS= */
    
   TidyAttr_XMLNSXLINK,             /**< svg xmls:xlink="url" */
-   
+  TidyAttr_SLOT,                   /**< SLOT= */
+
   N_TIDY_ATTRIBS                   /**< Must be last */
 } TidyAttrId;
```

If you agree, appreciate it being added to this PR... thanks...

I built your `slot` branch, and used it to run your forked regression tests, `pr-848` branch, and it all ran fine, passing the new `case-848.html`...

And of course that case fails using the current 5.7.29, as expected... thanks...

Are there any other tests I need to run? I will look around for some `slot` usage examples...

So, if the above `enum` change, in the PR, is agreed, then it seems we are ready to go... 

Look forward to further feedback, comments, other patches, PR, etc, on this... thanks...

geoffmcl:
@lhchavez eventually got to it... now merged... thanks...

PR #853, opened 2019-12-18
Protect against NULL in PruneSection.

esclim:
A Null node happens when pruning unbalanced <![if ...]> <![endif]>.

### DIFF
--- a/src/clean.c
+++ b/src/clean.c
@@ -1638,6 +1638,9 @@ static Node* PruneSection( TidyDocImpl* doc, Node *node )
 
     for (;;)
     {
+        if (node == NULL)
+            return NULL;
+        
         ctmbstr lexbuf = lexer->lexbuf + node->start;
         if ( TY_(tmbstrncmp)(lexbuf, "if !supportEmptyParas", 21) == 0 )
         {


esclim:
Sample HTML file and config file.
[test.zip](https://github.com/htacg/tidy-html5/files/3980205/test.zip)

geoffmcl:
@esclim after a quick test, with your supplied sample... thanks for that... 

Looks like a good catch, and fix... of a bug just waiting to happen... thanks...

geoffmcl:
@esclim, sorry for the delay...

Have added your sample to my [repo](https://github.com/geoffmcl/tidy-test), as `in_853.html` and `cfg_853.txt` ... so I could pull, and test in linux...

```
$ tidy -v
HTML Tidy for Linux version 5.7.28
$ tidy -config test/input5/cfg_853.txt test/input5/in_853.html
line 3 column 6 - Warning: inserting missing 'title' element
Segmentation fault (core dump)
```

**WOW**, there, this **BUG** causes a `segfault`!!!

**Must** get to merging this PR soonest... thanks...

PR #865, opened 2020-02-21
Add vcpkg installation instructions

PhoebeHui:
Tidy-html5 is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build tidy , ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for tidy-html5 and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README/BUILD.md
+++ b/README/BUILD.md
@@ -40,6 +40,17 @@ If you do **not** need the tidy library built as a 'shared' (DLL) library, then 
 
 See the `CMakeLists.txt` file for other CMake **options** offered.
 
+Alternatively, you can build and install tidy using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+  1. `git clone https://github.com/Microsoft/vcpkg.git`
+  2. `cd vcpkg`
+  3. `./bootstrap-vcpkg.sh`
+  4. `./vcpkg integrate install`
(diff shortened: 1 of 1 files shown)


geoffmcl:
@PhoebeHui, thank you for your PR, and nothing wrong with mentioning `vcpkg` in our docs, somewhere, but...

But I do not think it necessary to offer VCPKG as an `"Alternatively, you can build and install tidy using ..."`... in our BUILD.md... where we want to promote the simple `git clone/cmake` ... 

Perhaps VCPKG is best mentioned in our [Get Tidy](https://www.html-tidy.org/#homepage19700601get_tidy), where other `Package Managers` are added...

I have experience with other projects, which use VCPKG to good effect, especially when 3rdparty libraries, multiple sources, are involved... or elsewhere...

But HTML Tidy, has no such dependencies, so ... no!

Anyway, the site pages, offers you an alternative... so am closing this PR...

PhoebeHui:
Thanks for your response!

PR #881, opened 2020-05-18
Complete pt_br translation

hugotiburtino:
Complete some missing fields at the localize/translations/language_pt_br.po


geoffmcl:
@hugotiburtino, thanks for these...

I will try to get around to merging this, when I start considering the outstanding PRs... soon I hope... thanks...

PR #887, opened 2020-06-30
Added support for Swift Package Manager

Taxxodium:
This will allow tidy-html5 to be used with the Swift Package Manager.

Once installed, the developer just has to write `import tidy-html5` in their Swift code.


geoffmcl:
@Taxxodium, thank you for the PR... I think...

Assuming we want `support for Swift Package Manager`, I guess it depends on what we must add to our repo...

Commit ba561e5 seems to add a whole host of things, which is **NOT** agreed...

Could you present a simple proposal to `Tidy`, get that discussed, decided, **before** we consider a PR... thanks...

Taxxodium:
@geoffmcl This PR just makes it easy for Swift developers to add the library to their projects using the Swift Package Manager. Before this, developers would need to go either through various configuration settings to use the C library in their Swift code or make an Objective C wrapper.
By adding support for the Swift Package Manager, developers just include it and can use it off the bat.

Unless you don't want the library to be used in such languages as Swift, then this PR is not needed and I guess developers could use my fork instead of the official repository.

PR #899, opened 2020-10-09
Free attributes before return NULL

ltx2018:
fix memleak in GetTokenFromStream()

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -3457,6 +3457,10 @@ static Node* GetTokenFromStream( TidyDocImpl* doc, GetTokenMode mode )
         return node;  /* the COMMENT token */
     }
 
+    /* check attributes before return NULL */
+    if (attributes)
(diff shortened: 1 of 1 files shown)


ltx2018:
And here is the reproduce testcode(based from oss-fuzz):
[poc.zip](https://github.com/htacg/tidy-html5/files/5354105/poc.zip)

geoffmcl:
@ltx2018 thank you for this next PR... and the poc.c sample code, to reproduce...

But actually you can use `tidy` itself,  like just running -

```
$ echo "<?xml a" | valgrind tidy -q
```

And applying your 9e05e16 patch clears the heap summery - `in use at exit: 0 bytes in 0 blocks` - great...

Look forward to others testing this, and reporting... thanks...

Baring any negative reports, but do not expect any, will get to merging this soonest... thanks...

geoffmcl:
@ltx2018 PR now merged... thanks...

Issue #903, opened 2020-10-31
SVG attributes flagged as proprietary


cqcallaw:
The following HTML input to tidy (build from 188988022da4a64d80bc4a2eba21c33d57eb5152 on the *next* branch) validates with zero errors and zero warnings on https://validator.w3.org/nu/#file, but generates the following warnings in tidy. The "proprietary" attributes seem to be well-defined in [section A.3.7 of the SVG 1.1 DTD](https://www.w3.org/TR/SVG11/svgdtd.html#DTD.1.7), so this looks like a simple implementation gap that can be filled by adding the attributes to `attrs.h`, `attrs.c`, and `attrsdict.c` If so, I can contribute a pull request.

## Tidy Output
```
$ ./build/cmake/tidy -i -w 120 svg.html 
line 8 column 9 - Warning: <svg> proprietary attribute "stroke-width"
line 8 column 9 - Warning: <svg> proprietary attribute "stroke-linecap"
line 8 column 9 - Warning: <svg> proprietary attribute "stroke-linejoin"
line 8 column 9 - Warning: <svg> proprietary attribute "stroke-miterlimit"
line 8 column 9 - Warning: <svg> proprietary attribute "stroke-opacity"
line 8 column 9 - Warning: <svg> proprietary attribute "stroke-dasharray"
line 8 column 9 - Warning: <svg> proprietary attribute "fill"
line 8 column 9 - Warning: <svg> proprietary attribute "stroke"
Info: Document content looks like HTML5
Tidy found 8 warnings and 0 errors!
...
```

## HTML
```
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="UTF-8">
        <title>SVG sample</title>
    </head>
    <body>
        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" stroke-width="2" stroke-linecap="butt" stroke-linejoin="round" stroke-miterlimit="4" stroke-opacity="1" stroke-dasharray="none" fill="currentColor" stroke="currentColor">
            <path d="m 0.75,12.0 11.25,-9.0 4.5,3.375 0.0,-1.5 2.25,0.0 0.0,3.375 4.5,3.75 -2.25,0.0 0.0,9.0 -5.625,0.0 0.0,-7.5 -6.75,0.0 0.0,7.5 -5.625,0.0 0.0,-9.0 z" />
        </svg>
    </body>
</html>
```

geoffmcl:
[...]

PR #911, opened 2020-11-20
Trim trailing whitespace

cqcallaw:
No code changes here, just a quality-of-life edit per #903. I simply ran the following command:

```
find * -type f -exec sed -i 's/[ \t]*$//' "{}" \;
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,8 +56,8 @@ if (LIBTIDY_VERSION)
     list(GET VERSION_LIST 2 TIDY_POINT_VERSION)
 else ()
     message(FATAL_ERROR "*** FAILED to get a VERSION from version.txt!")
-endif ()   
- 
+endif ()
+
 # Establish version date
 if (LIBTIDY_DATE)
     string(REPLACE "." ";" DATE_LIST ${LIBTIDY_DATE})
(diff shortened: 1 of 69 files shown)


geoffmcl:
@cqcallaw thank you for this PR... and the recent update to clear merge conflicts...

I am afraid I really can't fully explain why I **hesitate** on this... 

It is something to do with my workflow... when chasing down a bug, I often do code diffs to previous, old, code, to explore, understand, get clues, etc, etc... Yes, I know `diff` has a `-bB` options to ignore spaces and blank line, but that does not feel the same... somehow I feel I might lose something, since as pointed out many of the trailing space do come from quite old code... perhaps not very rational...

Maybe we could split it up into two parts - one for the code, `*.c|h`, and one for the other docs, which I am not so concerned about... or something...

But I am 49:51, so not **all** against... it does nearly seem a **good thing**... I think!

Maybe, with some more time, I will come around... please bear with me... feedback welcome... thanks...

cqcallaw:
Philosophically, I like keeping the source code tidy, since tidiness is the project's _raison d'être_, but I've seen enough code to respect a maintainer's intuition.

The whitespace does present a challenge to me as a contributor, as I have to toggle a setting in my editing environment (Visual Studio Code) when I switch projects; the per-project setting seems to be broken. Breaking your workflows probably does more harm to the project than breaking mine, though.

geoffmcl:
@cqcallaw, sorry, but this PR gives me the `heebie geebies` ;=() It has some positive attributes, but...

But, in fact, for some reason, hinders me working on other parts of `Tidy`, maybe more **important** than the normally, invisible, do nothing, trailing space...

Asside from my workflow concerns mentioned, maybe there are cases where a trailing 2 spaces is significant as in `*.md` docs. Need to check for those... 

But also really do **not** want to send the message that `tidy` main repo is 100% **trailing space intolerant**... And, not that it really matters, find quite a number of other FOSS repos likewise inclined...

Like historically, `tidy` is, with **tabs in code**... even has a tool `tab2space`... which I also never quite understood...

Of course, we **also** do not want to imply that anything goes... but that should be dealt with on a case by case basis...

Also reject the _raison d'être_ argument. The `tidy` in tidy is about `pretty printing` HTML code, **not** 'C' code, and other text documents...

On the whole, still remain 49:51, so closing this all-in-one PR for now... sorry... and sorry if it breaks your editor workflow...

As suggested, if there are any particularly **bad** cases, then these shoud be dealt with separately...

Further comments and feedback welcome... maybe I am missing something... thanks...

Issue #681, opened 2018-02-19
Tidy 5.6.0 on Mac says Not a file when file is not writeable


thvv:
I check files before install with 
tidy --markup no --quiet yes --show-warnings no index.html
I get the error 
Document:  "index.html" is not a file!
well, it certainly is. But it is mode 444. Which is fine because I am not updating it, just checking it.
Changing the file mode to 666 makes it work.

I installed this version from macports tidy-5.6.0_0.darwin_16.x86_64.tbz2

geoffmcl:
@thvv **wow** what an interesting discovery...

It seems it is **not** only on the MAC, but is the **same** in `unix/linux`, and `Windows`, using current tidy 5.7.3...

I just set a file to 444 - `$ chmod 0444 temp.html`, and then `$ tidy temp.html` and I get `Document: "temp.html" is not a file!` - Zute, or as `Snagglepuss` would say `Heavens to Murgatroyd`!!!

It certainly seems at least one `write` bit must be set, like `$ chmod 0644`, and `tidy` will be ok...

On the face of it this does **not** seem right. Provided you are **not** using the `-modify` option, which **will** try to overwite the file, it seems only **read** permission should be required...

**AH HA**! The answer is **simple**. The `tidyDocParseFile` service has the following code -

```
    int status = -ENOENT;
    FILE* fin = fopen( filnam, "r+" );

    if ( !fin )
    {
        TY_(ReportFileError)( doc, filnam, FILE_NOT_FILE );
        return status;
    }

    fclose( fin );
```

Now reading some `fopen` references that access string `"r+"` means `**read/update**: Open a file for update (both for input and output). The file must exist.`

And **note** that if I set the file **read only** in Windows, `> attrib +R temp.html`, tidy will **fail** at the **same** place...

Now after that code block, WIN32 and UNIX separate... 

`UNIX` continues with `fin = fopen( filnam, "rb" );`... you will **note** the `+` has been removed...

Windows goes another code path, using in essence -

```
    HANDLE fp = CreateFileA( filnam, GENERIC_READ, FILE_SHARE_READ, NULL,
                              OPEN_EXISTING, 0, NULL );
    ...
    fin->map = CreateFileMapping( fp, NULL, PAGE_READONLY, 0, 0, NULL );
    ...
    data->view = MapViewOfFile( data->map, FILE_MAP_READ,
                                (DWORD)( data->pos >> 32 ),
                                (DWORD)data->pos, numb );
```

[...]

PR #926, opened 2021-03-27
Is. #681 - read-only files, and dirs

geoffmcl:
Please test, and report/comment in issue #681... thanks...

### DIFF
--- a/src/tidylib.c
+++ b/src/tidylib.c
@@ -1125,19 +1125,26 @@ int TIDY_CALL  tidyParseSource( TidyDoc tdoc, TidyInputSource* source )
     return tidyDocParseSource( doc, source );
 }
 
-
+#ifdef WIN32
+#define M_IS_DIR _S_IFDIR
+#else // !WIN32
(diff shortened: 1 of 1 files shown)


jmadiot:
Hello, thank you for the fix!  This fixes the problem for me (linux). I detail here what I did for @geoffmcl to check that I correctly tested and maybe to make it easier for others to test:

```bash
mkdir /tmp/testtidy
cd /tmp/testtidy
git clone https://github.com/htacg/tidy-html5.git
cd tidy-html5/
git switch issue-681
cd build/cmake/
cmake ../.. -DCMAKE_BUILD_TYPE=Release
make
echo "foo" > foo.html
./tidy foo.html
# correct output
chmod -w foo.html
./tidy foo.html
# also correct output!
# I also check that the branch next had the problem:
git switch next
make
./tidy foo.html
# expectedly got the error of issue 681: Document: "foo.html" is not a file!
chmod +w foo.html
./tidy foo.html
# normal output again
```

Issue #738, opened 2018-05-25
rendering of textarea tag broken


geoffmcl:
From Karl, of edbrowse...
 
One item that has been on the to-do list for a long time -
you muck with the contents of `<textarea>` but somewhat like `<script>`, you should really leave it alone.
That's not just me, it's standard, and done by every other browser.

```html
<body>
<textarea>hello<br>world</textarea>
</body>
```

Lots of websites put sample html into a textarea, html that you might want to cut & paste into your own web page or whatever,
you shouldn't mess with the tags at all.
https://www.prchecker.info/check_page_rank.php
Look at the stuff in the textarea.
I get around it by a preprocessing routine html-tidy.c line 17, and it works so well I almost forgot it was happening,
but at some point you may want to address this.
The contents of textarea should be `hello<br>world` as written, not `helloworld`.

Thus there are 3 modes.
1. Process and clean up the tags.
2. Don't touch or change a single byte, as in `<script>` or `<style>`.
If you change anything at all here, things don't work.
3. don't interpret any html tags but still expand `&characters;` Only two I know here are `<textarea>` and `<title>`

Hope this helps, and thank you again.

My @geoffmcl reply:

I am reviewing the `<textarea>` tidy code, and understand what you want changed... ok, yes, the contents of textarea should be `hello<br>world` as written, not **helloworld**, and a warning that a `<br>` was deleted...

I can get that with a small change in the textarea text gathering... it is done in `ParseText` service... 

But we want the text like as in `<script>var txt = 'hello<br>world';</script>`, where the text is treated as CDATA, and is not changed in any way... is the **raw** text stream until the `</textarea>` is encountered, if ever... so I change it to `ParseScript`...

Now we will get `<textarea>hello&lt;br&gt;world</textarea>`, which is not bad, but is there an option to skip escaping...

I think this change only happens in the output, pretty print... In there, if I change the output of textarea to use `PPrintScriptStyle`, then I get -

```html
<textarea>
hello<br>world
</textarea> 
```

Do not know if I can get rid of the newlines, but you do not consume html, just the node list, maybe you will not have this problem...

Added to `tidy` issues, so I can assign an issue number to the **fixes**, probably in a branch initially...

The current `tidy` diff is - 

```diff
diff --git a/src/pprint.c b/src/pprint.c
index 321045e..25f72bc 100644
--- a/src/pprint.c
+++ b/src/pprint.c
[...]

PR #929, opened 2021-03-27
Is. #738 - textarea tag broken

geoffmcl:
Please test this, and report, comment, ... in issue #738 ... thanks...


balthisar:
As mentioned in the conversation in #738, this change currently breaks the HTML display output in at least one of the test case regressions. I'll close this for now, although the branch will remain for further development. Please feel free to re-open when this is addressed.

PR #953, opened 2021-06-13
simple fix for the range of the condition.

ihsinme:
simple fix for the range of the condition.

### DIFF
--- a/src/streamio.c
+++ b/src/streamio.c
@@ -803,7 +803,7 @@ static void EncodeIbm858( uint c, StreamOut* out )
 /* Convert from Latin0 (aka Latin9, ISO-8859-15) to Unicode */
 static uint DecodeLatin0(uint c)
 {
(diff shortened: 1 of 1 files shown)


balthisar:
@ihsinme, is there a bug associated with this? Can you provide a failing testcase for current behavior, and a passing testcase with your changes?

I'm not a character encoding expert, and your change is to some very old, presumed stable code, so I'd really like you to coach me on what this is supposed to do.

Thanks.

ihsinme:
good day.
in this code, everything is just `if (159 <c && c <191)` does not restrict rigidly. in fact, for `c` equal to `160,161,162,163`, the `switch` has no handlers.

and to optimize the code to correct the condition correctly.

balthisar:
@ihsinme, thanks, I see. I should have looked a bit more closely! I'll merge this then bump the version.

Issue #519, opened 2017-03-28
Tidy should word wrap its message output


balthisar:
Tidy should word wrap its message output when used in a console of arbitrary sizes. Currently all Tidy output is limited to 80 characters regardless of console dimensions.

- Sometimes when working with multiple consoles, in a crowded window, the size is fewer than 80 columns.
  - While many consoles wrap, some do not.
  - Consoles that wrap do no break on whitespace; they break at the end of the line.
- Often, though, the console far exceeds 80 columns.
  - Tidy wraps at 80 columns despite having additional columns available.
  - Users want to use the full width of their console.
- All well-behaved console applications use the entire given console width.
  - Can you imagine reading a man page forced to 80 columns?
  - What if vi or nano didn’t use the full console width?
  - Tidy is currently not “well behaved” in this respect.
- When outputting to a file, we often prefer soft-wrap rather than hard wraps, because editor size is like terminal size, above.
- When fetching messages from the callback API, the current forced wrapping makes adaption to GUI rather ugly, in that there’s no fixed number of columns, and fonts are more often proportional in size than not.

* * *

### Proposal

- Remove line breaks (other than those present for formatting or as paragraph breaks) from Tidy’s strings.
- Modify the current output functions to wrap text at an arbitrary number of columns.
- Have the console application use the width of the terminal by default as the word wrap point.
- If the `file` option is used, or any output is piped to a file, or Tidy is run in a non-interactive shell, it should default to 80 characters.

### Optional

- We should add a new configuration option (for example, `console-width`) to allow users to specify their own preference.
  - For example, some people are perfectly happy with 80 characters of output all the time.
  - Other uses may want to specify no wrapping at all for redirected output (so that text editors works with whole paragraphs rather than bits and fragments interrupted by carriage returns).

### Alternative proposal

- Do not include the new option.
- Do not detect the tty width.
- _Do_ eliminate newlines in the source text.
- Fix output to 80 columns as it is today, but let the word-wrap algorithm perform the wrapping rather than code it by hand.
  - This still addresses the callback API issues by providing sentences and paragraphs instead of broken fragments.

* * *

Summary:

- Users that don’t want modern console behavior can use `console-width` to keep things statically frozen in time for console use
- Others get to enjoy standard behavior in the console.
- No change at all when used with `file` option, or pipes/redirection, or run in a login-only shell.
- API delivers real sentences instead of fragments.

geoffmcl:
[...]

PR #963, opened 2021-07-09
Dripping

gbebgdee:


balthisar:
I like it (see #519), but can you clean up the commit a little bit? We shouldn't leave the console application with the callback code activated, for example, as well as the other commented out code.

More importantly, are you certain that the generic approach works for EVERY existing string? A quick scan doesn't look like there would be any issues. One concern, though, is if you take a look at e.g., `TC_TXT_HELP_3`,  by swallowing the newlines, you're also swallowing the vertical spacing. Maybe it's not a concern.

I'm looking at this quickly, but if a string ends with two newlines, are you going to overrun the string that you're copying? 

I wonder if it's not time to finally just remove the newlines from the source, and perform wrapping in the console application. This has been done and rejected before.

Let me know your thoughts.

Whatever the case, I may defer this a couple of weeks in order to issue a stable 5.8 release with strings as they are, and make this the first commit on 5.9.

PR #976, opened 2021-07-21
Spelling fixes, thanks to @jschleus.

balthisar:


balthisar:
As there is no public-facing implication, I will not bump the version.

PR #1008, opened 2021-10-27
OSS-Fuzz Fixes

### LINKED ISSUE #1001: bug in new function strrep in tmbstr.c
```patch
 src/tmbstr.c | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/tmbstr.c b/src/tmbstr.c
index e352698..447c6dd 100644
--- a/src/tmbstr.c
+++ b/src/tmbstr.c
@@ -255,7 +255,7 @@ void TY_(strrep)(tmbstr buffer, ctmbstr str, ctmbstr rep)
         if(p)
         {
             char buf[1024];
-            memset(buf,'\0',strlen(buf));
+            memset(buf,'\0',sizeof(buf));
 
             if(buffer == p)
             {

```
if use vc to build the solution with debug config, and call tidyRunDiagnostics with the html in [the attachment zip file](https://github.com/htacg/tidy-html5/files/7241668/test.zip), will trigger the problem.

alpire:
This MR fixes a number of issues reported by OSS-Fuzz. 

# Infinite loop in `ParsePre`

The loop was introduced 91f29ea7 when switching to a non-recursive algorithm. `InlineDup` should not be called when `ParsePre` restarts. This issue has been preventing OSS-Fuzz builds for a few months, which means recent code changes haven't been fuzzed yet.

[OSS-Fuzz Issue](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36721)

# Fix initialization of stack buffer in `strrep`

`strlen(buf)` is not guaranteed to return the size of the buffer, as it might contain some null bytes in the middle, so this PR switches to `sizeof` instead. Fixes #1001. This wasn't reported by OSS-Fuzz due to the build failure above, but it triggers quickly when running the fuzzer locally.

#  Fix out-of-bounds write in `NormalizeSpaces`

`NormalizeSpaces` decodes and re-encodes UTF-8 characters while looking to replace non-breaking spaces with regular spaces. When the UTF-8 decoding hits an error, a replacement character (0xFFFD) is returned and re-encoded as a 3-byte UTF-8 character. In some cases, this increases the size of strings, leading to writing past the end of the allocated buffer.

Note that this *changes* the behavior of the function in the case of invalid UTF-8 strings. This is my first time contributing to the project, so it was difficult to determine if the behavior change made sense even though the tests are passing. I'd be happy to try another fix if you had another solution in mind. 

[OSS-Fuzz Issue](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13191)


balthisar:
Closing, then re-opening. Hopefully this triggers the CI, which should have run.

balthisar:
@alpire, thanks for the PR, and sorry for taking so long to get to it. Things are a bit behind since @geoffmcl has been absent.

I'm especially sorry that the integration testing didn't run automatically when you posted the PR; you wouldn't have wasted time waiting for me to force it to run manually. As you can see, though, there seem to be regression test failures on two out of the three Windows builds. Given that Cygwin passed and the other two environments are Unix-like, too, it's probably a Windows-specific bug.

I'd be happy to merge this PR if you can correct the issue. At some point, I may have time to work on it myself, but as you can see, there's a HUGE backlog.

Thanks!

alpire:
@balthisar Thanks for taking a look at the PR. The failing test was the one I just added, but the failure was (surprisingly) unrelated to the PR changes. The bug was in the windows mapped IO EOF handling. A fix is available in 5f3eb6605ad343049d96d38ec4f2ccb4688bcfec, which hopefully won't break other tests on windows.

ddkilzer:
Note:  There is another approach to fixing the out-of-bounds write crash (in `src/clean.c`) in PR #1138.

PR #1013, opened 2021-12-10
Adds missing {} to else and fixes a cosmetic issue

hoijui:
found through compiler-warnings

### DIFF
--- a/src/access.c
+++ b/src/access.c
@@ -889,7 +889,7 @@ static Bool CompareColors( const int rgbBG[3], const int rgbFG[3] )
 * blue = 5.
 *********************************************************************/
 
-static Bool GetRgb( ctmbstr color, int rgb[] )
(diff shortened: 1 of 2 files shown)


balthisar:
Closing, then re-opening in order to trigger CI.

balthisar:
@hoijui, thanks for the contribution. The change looks straightforward and the regression tests pass on all platforms. However there's an internal API change causing the API test to fail.

I'll try to dig into it eventually, but if you have any insight, that would be appreciated.

hoijui:
uff.. had a look at it, but could not see why it would fail. :/
the whole output does not contain "error", just a lot of warnings ...
I did not really use C since more then 10 years.

... ouh, should I maybe just rebase on current HEAD?

balthisar:
I'll have a look at it sometime this weekend. It's possible that the API test isn't robust, and is at fault.

PR #1016, opened 2021-12-22
makes sure that when building a target, its dependencies' PUBLIC and INTERFACE include directories will be used appropriately

SuperCodeHero:
makes sure that when building a target, its dependencies' PUBLIC and INTERFACE include directories will be used appropriately.

See [https://stackoverflow.com/questions/40227333/cmake-setup-multiple-projects-and-dependiencies-between-them/40242257](https://stackoverflow.com/questions/40227333/cmake-setup-multiple-projects-and-dependiencies-between-them/40242257)


balthisar:
Closing and re-opening to force CI testing.

balthisar:
@SuperCodeHero, thanks for the PR. Merging now.

PR #1019, opened 2022-01-05
Attribute 'type' in tag 'ol' allowed for (X)HTML5

### LINKED ISSUE #1012: The type attribute of the ol element is allowed in the actual HTML, but tidy-html5 does not allow it
If the ol element has a type attribute, tidy-html5 will issue a warning.

~~~html
<!DOCTYPE html>
<html lang="ja">
  <head>
    <meta charset="utf-8" />
    <title>ol</title>
  </head>
  <body>
    <ol type="i">
      <li>hoge</li>
    </ol>
  </body>
</html>
~~~

~~~console
$ tidy ol-type.html
line 8 column 5 - Warning: <ol> attribute "type" not allowed for HTML5
Info: Document content looks like HTML5
Tidy found 1 warning and 0 errors!

<!DOCTYPE html>
<html lang="ja">
<head>
<meta name="generator" content=
"HTML Tidy for HTML5 for Linux version 5.8.0">
<meta charset="utf-8">
<title>ol</title>
</head>
<body>
<ol type="i">
<li>hoge</li>
</ol>
</body>
</html>

About HTML Tidy: https://github.com/htacg/tidy-html5
Bug reports and comments: https://github.com/htacg/tidy-html5/issues
Official mailing list: https://lists.w3.org/Archives/Public/public-htacg/
Latest HTML specification: http://dev.w3.org/html5/spec-author-view/
Validate your HTML documents: http://validator.w3.org/nu/
Lobby your company to join the W3C: http://www.w3.org/Consortium

Do you speak a language other than English, or a different variant of
English? Consider helping us to localize HTML Tidy. For details please see
https://github.com/htacg/tidy-html5/blob/master/README/LOCALIZE.md
~~~

~~~console
$ tidy --version
HTML Tidy for Linux version 5.8.0
~~~

I checked and found that the source code defines that the type attribute does not exist in HTML5.

<https://github.com/htacg/tidy-html5/blob/2d15699ccf1564d11e764fb955e83c64d963d5ad/src/attrdict.c#L2300>

However, I actually consulted the HTML standards document and found that the ol element does have a type attribute.

<https://html.spec.whatwg.org/multipage/grouping-content.html#the-ol-element>

The type attribute of the ol element seems to have been restored as follows.

* [9798 – The type attribute on `<ol>` and `<ul>` is semantic and should be permitted](https://www.w3.org/Bugs/Public/show_bug.cgi?id=9798)
* [[ac] (2) Reintroduce `<ol type>` as conforming. · whatwg/html@4e3753d](https://github.com/whatwg/html/commit/4e3753d88a09243914e190e81036e12e6a1495b4)

Is this a bug?

Even if it is not a bug, I would like to optionally disable this warning only because Pandoc outputs HTML of this format.

MarkusPister:
Mark the attribute 'type' as allowed for tag 'ol' in (X)HTML5.

The attribute type was deprecated in HTML4. This decision was reverted
for HTML5, see

    https://www.w3.org/TR/2014/NOTE-html5-diff-20141209/#changed-attributes

This fixes htacg#1012.

### DIFF
--- a/src/attrdict.c
+++ b/src/attrdict.c
@@ -2321,7 +2321,7 @@ const AttrVersion TY_(W3CAttrsFor_OL)[] =
   { TidyAttr_START,                 xxxx|HT32|H40T|H41T|X10T|H40F|H41F|X10F|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 },
   { TidyAttr_STYLE,                 xxxx|xxxx|H40T|H41T|X10T|H40F|H41F|X10F|H40S|H41S|X10S|XH11|xxxx|HT50|XH50 }, /* CORE override */
   { TidyAttr_TITLE,                 xxxx|xxxx|H40T|H41T|X10T|H40F|H41F|X10F|H40S|H41S|X10S|XH11|XB10|HT50|XH50 }, /* CORE override */
-  { TidyAttr_TYPE,                  xxxx|HT32|H40T|H41T|X10T|H40F|H41F|X10F|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx|xxxx },
+  { TidyAttr_TYPE,                  xxxx|HT32|H40T|H41T|X10T|H40F|H41F|X10F|xxxx|xxxx|xxxx|xxxx|xxxx|HT50|XH50 },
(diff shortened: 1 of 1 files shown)


balthisar:
@MarkusPister, thanks. Although a case file would have been appreciated, all of the current regression tests pass, so I'll merge this and bump the version. Thanks!

PR #1028, opened 2022-03-09
fixing a possible write beyond the array boundary.

ihsinme:
If you don't initialize array `buf` and use `strlen` to determine the zero length, there is a possibility that the value of function `memset` will go beyond array `buf`. This will happen if there are no zeros in the declared array.

### DIFF
--- a/src/tmbstr.c
+++ b/src/tmbstr.c
@@ -255,7 +255,7 @@ void TY_(strrep)(tmbstr buffer, ctmbstr str, ctmbstr rep)
         if(p)
         {
             char buf[1024];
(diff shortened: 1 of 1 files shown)


ihsinme:
good afternoon.
any news on this PR?

zopsicle:
On line 262 it also calls `strcpy(buf,rep);` with seemingly no bounds check? If `rep` is larger than 1024 then it’ll overflow the buffer. Not an immediate problem as `strrep` is currently only called with constants and localized messages, but if it ever ends up being called with user input this is catastrophic.

PR #1066, opened 2023-01-11
Use the found xsltproc

### LINKED ISSUE #1065: CMakefile should use xsltproc discovered by find_program()
Although `find_program()` is used to find `xsltproc`, the build does not use the discovered executable and assumes that `xsltproc` can be reach via the PATH environment variable:

```
COMMAND xsltproc ARGS ${TIDY1XSL} ...
```

earlchew:
Do not assume xsltproc is in PATH, and use the executable discovered by find_program().

Fixes #1065


dg0yt:
#1044 does the same, plus it adjust the variable name.
But there is no rewiew/merge activity here.

PR #1085, opened 2023-05-02
fix: add final freeing of Stack to freeStack

SeanDougherty:
We noticed a memory leak when evaluating tidy w/ Valgrind. Together w/ @sameluch we determined the source of the leak to be a missing free of the Stack variable.

(Will add soon: valgrind of before and after the fix)

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -4538,11 +4538,12 @@ FUNC_UNUSED Node* TY_(peek)(Stack *stack)
 /**
  *  Frees the stack when done.
  */
-void TY_(freeStack)(Stack *stack)
+void TY_(freeStack)(Stack *stack, TidyDocImpl *doc)
 {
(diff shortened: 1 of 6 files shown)


src/lexer.c
```diff
@@ -4538,11 +4538,12 @@ FUNC_UNUSED Node* TY_(peek)(Stack *stack)
...
  */
-void TY_(freeStack)(Stack *stack)
+void TY_(freeStack)(Stack *stack, TidyDocImpl *doc)
 {
     TidyFree( stack->allocator, stack->firstNode );
     stack->top = -1;
     stack->capacity = 0;
     stack->firstNode = NULL;
     stack->allocator = NULL;
+    TidyFree( doc->allocator, stack)
```
Kijewski, line 4548: There is a semicolon missing at the end of the line.

PR #1088, opened 2023-06-12
Add Serbian translation

eevan78:
Here is the PO file of translation to Serbian language.


eevan78:
While testing the Serbian localization, I have found out that the display of options in console tool is not taking care of multibyte characters. I have opened the issue #1112

PR #1140, opened 2025-04-03
CMakeLists: Adjust for CMake 4.0's removal of deprecated syntax.

### LINKED ISSUE #1139: tidy-html5 fails to build with CMake 4.0
CMake 4.0 was released on March 28th, and is a lot more strict than previous versions. Because of the changes, it has broken the build process of tidy. A lot of things that were previously warnings are now no longer allowed.

When trying to build this package with CMake 4.0, the first error that you'll see is:

```
CMake Error at CMakeLists.txt:20 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

To fix this, I changed Line 20 of CMakeLists.txt to read "cmake_minimum_required (VERSION 2.8.12...4.0)"

After this though, it still failed to build:

```
-- Detecting CXX compile features - done
-> TIDY_SO_VERSION = 58. This is an EVEN (stable) release.
-- *** Debug Logging is NOT enabled.
-- *** Building support for runtime configuration files.
-- *** Also building DLL library SHARED, version 5.8.0, date 2021.07.10
-- *** Generating man tidy.1 custom commands...
CMake Error at CMakeLists.txt:530 (add_custom_command):
  Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given.


CMake Error at CMakeLists.txt:538 (add_custom_command):
  Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given.


CMake Error at CMakeLists.txt:546 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): DEPENDS


-- Configuring incomplete, errors occurred!
```

I've fixed this locally and will send in a PR shortly. I can confirm that my patch works with CMake 4.0 and CMake 3.30.x.

renodr:
CMake 4.0 is more strict than previous versions, and has enforced removals of deprecated syntax. Compatibility with versions of CMake prior to 3.5 is no longer permitted either, so we'll change it to say 4.0 to fix that problem. After that, the custom commands for generating the manual page need to be adjusted as CMake now needs to know when to run commands, so we will generate the manual page after the package is built by passing POST_BUILD to the custom commands.

Tested with CMake 4.0 and 3.31.6.

Fixes #1139

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@
 # @date    Consult git log.
 ##############################################################################
 
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 2.8.12...4.0)
 
 set(LIB_NAME tidy)
 set(LIBTIDY_DESCRIPTION "${LIB_NAME} - HTML syntax checker")
@@ -528,6 +528,7 @@ if (UNIX AND SUPPORT_CONSOLE_APP)
  
         # Run the built EXE to generate xml output .
         add_custom_command(
+            POST_BUILD
             TARGET man
             COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME} -xml-help > ${TIDYHELP}
             COMMENT "Generate ${TIDYHELP}"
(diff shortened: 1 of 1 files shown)


a17r:
Why don't you simply raise the minimum to something >=3.5, or >=3.10 (to get rid of yet another deprecation warning)?

PR #1153, opened 2025-06-09
Tamil translation file added

TamilNeram:
created the PO file for tamil language


TamilNeram:
@balthisar, Please add the Tamil translation file

